PR #40, opened 2020-11-27
12 add voltage level editor

### LINKED ISSUE #12: add voltage level editor
Add a container that displays and manipulates the `voltage level` element in the substation section of the SCL file.
The following user interaction shall be included

- add bay element
- add power transformer element
- edit name, desc, nomFreq, numPhases attributes as well as voltage element  of the voltage level element
- voltage element attributes shall be multiplier="k" and unit="V"
- delete the voltage level element
- add LNode to the voltage level element - opens UI to connect and save the bay element in the destination filter
- copy bay structure (excluding the LNode)

c-dinkel:
Closes #12


JakobVogelsang (approved):
LGTM

PR #48, opened 2020-12-03
fix(wizard-textfield): reset custom validity

### LINKED ISSUE #45: Custom checkValidity() in wizard-textfield does not reset customValidity
In the `mwc-textfield` doc it says that:
```If this message is not the empty string, then the element is suffering from a custom validity error and does not validate.```
When not reset to an empty string the outcome of `super.checkValidity()` is always false. 

Possible solution would be to reset custom validity message before checking `super.checkValidity()`.

JakobVogelsang:
Resolves #45


c-dinkel (approved):
LGTM

PR #54, opened 2020-12-09
feat(open-scd): add save functionality

### LINKED ISSUE #4: Add "save `SCD` file" button
Add a button saving the serialized `XML` string of the currently edited `XMLDocument` as a `*.scd` file.

Blocked by #2

c-dinkel:
Closes #4


c-dinkel:
This cannot be tested because downloads are built in browser functionality.

src/open-scd.ts
```diff
@@ -126,6 +126,24 @@ export class OpenSCD extends Setting(
     }
   }
 
+  private save(): void {
+    const blob = new Blob([new XMLSerializer().serializeToString(this.doc)], {
+      type: 'application/xml',
+    });
+
+    const a = document.createElement('a');
```
JakobVogelsang, line 134: Why this?

c-dinkel: Because https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download does not work in Safari yet and we need a way to trigger the actual download, so we add an anchor to the document body.

JakobVogelsang (commented):
A Save As... function would be nice to have. This was one of the most asked features in the java version.

c-dinkel:
I've added a "Save as" button.

PR #64, opened 2020-12-10
fix(substation/editors): desc textfield default empty string

### LINKED ISSUE #56: Misleading disabled desc textfield
For someone not familiar with the standard it might be misleading that a textfield is disabled. Especially when creating new elements the predefined attributes should never be `null`. 

Suggestion: predefined value for `desc` textfield should be empty string instead of null.

JakobVogelsang:
Closes #56


c-dinkel (approved):
LGTM 👍

PR #71, opened 2020-12-14
feat(schemas): add schemas for editor1 and edition 2.1

JakobVogelsang:


c-dinkel (approved):
LGTM

src/schemas.ts
```diff
@@ -3356,4 +5599,3744 @@ export const schemas: Partial<Record<string, string>> = {
     </xs:key>
   </xs:element>
 </xs:schema>`,
+  '2007B4': `<?xml version="1.0" encoding="UTF-8"?>
```
c-dinkel, line 5602: Don't we have Versions `2007B2` and `2007B3`?

PR #72, opened 2020-12-14
feat(open-scd): enable opening of large files

c-dinkel:
Enables editing of files larger that the browser's `string` size limit.


c-dinkel:
Will not be integrated into the mainline editor.

PR #85, opened 2020-12-22
feat(logging): add tooltip to make long string readable

JakobVogelsang:
I needed the log from the schema validator and found it very disturbing that long messaged were cut off. The tooltip is a quick solution. We might discuss how we want to solve this on a long run.


c-dinkel (approved):
LGTM

PR #99, opened 2021-01-07
Add complex actions

### LINKED ISSUE #98: Add complex actions
The log easily gets flooded with many individually un- and redoable entries when a single user action causes many simple editor actions (`Create`, `Update`, `Move`, or `Delete`) to happen at once.

A complex editor action type which consists in an array of simple editor actions that are handled and committed to the log as a single un- and redoable entry is intended to help alleviate this issue.

c-dinkel:
Resolves #98 .

### DIFF
--- a/__snapshots__/open-scd.md
+++ b/__snapshots__/open-scd.md
@@ -1,0 +1,397 @@
+# `open-scd`
+
+#### `looks like its snapshot`
+
+```html
+<mwc-drawer
+  class="mdc-theme--surface"
+  hasheader=""
+  id="menu"
+  type="modal"
+>
+  <span slot="title">
+    Menu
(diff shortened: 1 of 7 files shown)


c-dinkel:
The code in this draft is currently not exercised at all and needs unit tests before merging. I will refactor and write tests as soon as @JakobVogelsang agrees to the proposed API for `ComplexAction`s.

JakobVogelsang (commented):
I like the API. I think it fits our need. The only thing we might consider is a flexible title for complex actions. Ideally you can decide for each service individually.

JakobVogelsang (approved):
LGTM

JakobVogelsang (approved):
LGTM

PR #100, opened 2021-01-09
Guess substation cbsw assumtion

### LINKED ISSUE #16: Add guess function to substation equipment
When no or an empty substation section is present OpenSCD shall guess the substation structure and its connection to the IED elements. Two algorithms shall be installed:

1. Guess on `CSWI` assumption
This assumption is based on the `CSWI` being a bay controller and therefore every IED that has at least one such logical node shall be interpreted as a bay controller. 
- For every bay controller (IED with at least one `CSWI` logical node) a bay shall be created. 
- If the bay controller has a joint logical node `XCRB` or`XSWI` a `ConductingEquipment` shall be created with the specific `type` attribute `CBR`, `DIS`, respectively.  Joint means that the logical device `CSWI` resides in as well as the `prefix` attribute and the `inst` attribute of `CSWI` and `XCBR`, `XSWI`, respectively, match.
- All IEDs that did not have a `CSWI` in their data model shall be sorted to the `Bay` elements based on the number of GOOSE and SMV connections. In case there are two `Bay` elements with the same amount of connection compared to another `Bay` the alphabetically first shall be chosen. 
- Naming convention for the `Bay` element shall be `Qxx` (e.g Q01)
- Before creating `Bay` elements the `IED` elements shall be sorted alphabetically

2. Guess based on the `name` attribute of the `IED` element
Very often the `name` attribute of `IED` is constructed from the location of the `IED` within the substation structure. A typical example might be `AA1E1Q01` standing for substation `AA1` voltage level `E1` and bay `Q01`. However the identifier a bay (in the example 'Q') the voltage level ( in the example 'E') might differ from region to region. 
  - Allow to define identifier(s) for both voltage level's and bay's being the first inputs to the guess algorithm
  - First sort the IEDs to the voltage levels based on the defined voltage level identifiers 
  - Second sort the IEDs within voltage levels to the bays based on the identifier(s) for the bay
  - Sort the `Bay` elements alphabetically

3. Handling switch gear
Adding switch gear to `Bay` element as `ConductingEquipment` independent of the `ctrModel` might lead to a lot unnecessary `ConductingEquipment`. Reason is that logical nodes `CSWI`, `XCBR` and `XSWI` are not only uses for switch gear in the bay but for other automation functions.  That is why `ctlModel` shall be used to differentiate between the switch gear and other controlled objects in the bay.
- Allow user to choose for the `ctlModel` of the switch gear

 

Restrictions:
- `name` attribute of the `VoltageLevel` element shall be unique within the `Substation` element
- `name` attribute of the `Bay` element shall be unique within the `VoltageLevel` element
- `name` attribute of the `ConductingEuipment` element shall be unique within the `Bay` element
- `LNode` element shall be unique with the substation meaning the combinations of its attributes shall be unique

JakobVogelsang:
Closes #16


c-dinkel:
@JakobVogelsang what do you think of my proposed change separating `Substation` creation and guessing of the `VoltageLevel` in 3364e9b544c4be48c0f0dc8fd528936c7a96521c ?

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+  WizardAction,
+  WizardInput,
+  CloseableElement,
+  EditorAction,
+  sortElementByNameAttribute,
+} from '../../foundation.js';
+
+let bayNum = 1;
+let cbNum = 1;
+let dsNum = 1;
```
c-dinkel, line 21: Could we get rid of this global state by moving from module-global variables to function parameters and closures?

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+    cswi.parentElement?.querySelector(
+      `LN[lnClass="XSWI"]${
+        cswi.getAttribute('prefix')
+          ? `[prefix="${cswi.getAttribute('prefix')}"]`
+          : ``
+      }${
+        cswi.getAttribute('inst') ? `[inst="${cswi.getAttribute('inst')}"]` : ``
+      }`
+    )
+  )
```
c-dinkel, line 94: Is this complex `if` statement really needed if we're going to `return 'DIS'` either way?

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+    return ln.getAttribute('prefix')! + ln.getAttribute('inst');
+
+  if (ln.getAttribute('inst') && getSwitchGearType(ln) === 'CBR')
+    return 'QA' + cbNum++;
+
+  return 'QB' + dsNum++;
+}
+
+function isSwitchGear(ln: Element, selectedCtlModel: string[]): boolean {
+  if (!ln || !ln.ownerDocument) return false;
```
c-dinkel, line 111: `ln` is never falsey since it has to be of type `Element`, and `Element.ownerDocument` is never `null`, so when would this ever `return false`?

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+        Array.from(
+          doc.querySelectorAll(
+            `DOType[id="${doType}"] > DA[name="ctlModel"] > Val`
+          )
+        )
+      )
+      .filter(val => selectedCtlModel.includes((<Element>val).innerHTML.trim()))
+      .length
+  )
+    return true;
```
c-dinkel, line 142: Instead of `if (x) return true; return false;` you can simply write `return x` or `return x !== null` or the like.

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+function getCSWI(ied: Element): Element[] {
+  return Array.from(
+    ied.querySelectorAll('AccessPoint > Server > LDevice > LN[lnClass="CSWI"]')
+  );
+}
+
+function getValidCSWI(ied: Element, selectedCtlModel: string[]): Element[] {
+  if (!ied || !ied.parentElement) return [];
```
c-dinkel, line 154: `ied` is never falsey since it has to have type `Element`.

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+  if (switchGear.length) {
+    const bayElement = new DOMParser().parseFromString(
+      `<Bay name="Q${bayNum++}" desc="Bay for controller ${
+        ied.getAttribute('name') ?? ''
+      }"></Bay>`,
+      'application/xml'
+    ).documentElement;
+
+    const condEq = switchGear.map(cswi => {
+      const condEq = new DOMParser().parseFromString(
```
c-dinkel, line 176: `condEq` shadowing `condEq` in the initialization of `condEq` is hard to read. Can we find more meaningful names here?

There are several singular array names in this file that might be pluralized, for example.

src/editors/substation/guess-wizard.ts
```diff
@@ -0,0 +1,266 @@
...
+        icon: 'play_arrow',
+        label: get('guess.wizard.primary'),
+        action: guessBasedOnCSWI(doc),
+      },
+      content: [
+        html`<h2>Can we supply an explanation?</h2>
+          <p>
+            This wizard page may benefit from some translated text explaining
+            what has to be chosen and why.
+          </p>`,
```
c-dinkel, line 249: From a user's point of view, this wizard dialog does not seem self explanatory to me yet. Do you think explanatory text could be appropriate here?

src/foundation.ts
```diff
@@ -317,3 +318,8 @@ export const restrictions = {
   decimal: '((-|\\+)?([0-9]+(\\.[0-9]*)?|\\.[0-9]+))',
   unsigned: '\\+?([0-9]+(\\.[0-9]*)?|\\.[0-9]+)',
 };
+
+/** Sorts selected `ListItem`s to the top and disabled ones to the bottom. */
+export function sortElementByNameAttribute(a: Element, b: Element): number {
+  return a.getAttribute('name')!.localeCompare(b.getAttribute('name')!);
+}
```
c-dinkel, line 325: Since this function compares and does not sort, I suggest renaming it to `compareNames` or something similar.

test/integration/editors/guess-wizarding-editing.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  describe('renders one wizard page', () => {
+    it('asking which ctlModel the is used for switchgear', async () => {
+      expect(element.wizardUI.dialogs.length).to.equal(1);
+      expect(
+        element.wizardUI.dialog!.querySelectorAll(
+          '#ctlModelList > mwc-check-list-item'
+        ).length
+      ).to.equal(5);
+    });
+    it('the first one beeing status-only', async () => {
```
c-dinkel, line 34: The spelling "beeing" is considered a bit out of date, "being" being the preferred spelling now, but here I'm really niitpicking 😜

test/integration/editors/guess-wizarding-editing.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  it('creates only unique conducting equipment names', () => {
+    const nameArray: string[] = Array.from(
+      validSCL.querySelectorAll(
+        ':root > Substation > VoltageLevel > Bay > ConductingEquipment'
+      )
+    ).map(item => item.getAttribute('name')!);
+    const nameSet = new Set(nameArray);
+    expect(nameArray.length).to.equal(nameSet.size);
+  });
+  it('creates unique conducting equipment name, if not prefix is there', () => {
```
c-dinkel, line 134: "if not prefix is there" is a bit hard to read, but I can guess the meaning from the code.

c-dinkel (changes requested):

c-dinkel (approved):
LGTM

PR #104, opened 2021-01-13
Pretty print

### LINKED ISSUE #93: Create an SCD block the tabs in OpenSCD 2.8.1 until save and re-open
Creating a SCD from scratch in OpenSCD 2.8.1 block all the tabs and buttons for create the substation.
 
When the project is saved and re-open then the tabs and buttons are still available

c-dinkel:
Closes #93 .


src/open-scd.ts
```diff
@@ -130,11 +130,27 @@ export class OpenSCD extends Setting(
     this.saveUI.close();
   }
 
+  private formatXml(xml: string, tab?: string) {
+    let formatted = '',
+      indent = '';
+
+    if (!tab) tab = '\t';
+    xml.split(/>\s*</).forEach(function (node) {
```
c-dinkel, line 138: Could we use a lambda here for stylistic consistency?

src/open-scd.ts
```diff
@@ -130,11 +130,27 @@ export class OpenSCD extends Setting(
     this.saveUI.close();
   }
 
+  private formatXml(xml: string, tab?: string) {
```
c-dinkel, line 133: Do we use the configurable indentation anywhere?

src/open-scd.ts
JakobVogelsang: For sure! Good idea

src/open-scd.ts
JakobVogelsang: Not know at least. You are right. If we might need it we can add later.

JakobVogelsang (approved):
LGTM

PR #109, opened 2021-01-25
Add clone feature

### LINKED ISSUE #81: Add copy/paste functionality to the substation editor
Substation elements as `Bay` and `VoltageLevel` do not vary a lot in their structure in a substation. This is especially true for the `Bay` element. 
Copying a `Bay` element structure and pasting to the same or other parent elements would be a speed booster. 

Restrictions:
-Make sure that then `name` attribute is unique
-Make sure that all `LNode` elements stay unique

c-dinkel:
Use numbers instead of "- copy" to distinguish cloned elements.

Closes #81 .


c-dinkel:
@JakobVogelsang do you think you could find the time to adapt the tests to the new numbering scheme?

PR #111, opened 2021-02-03
Create network configuration plug in

JakobVogelsang:
The code is ready for review


src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,418 @@
...
+          class="menu-item right"
+          icon="delete"
+          @click="${() => this.remove()}}"
+        ></mwc-icon-button>
+      </div>
+      <h4>${this.apName}</h4>
+    `;
+  }
+
+  static creatAction(parent: Element): WizardAction {
```
c-dinkel, line 104: Don't be `UNIX` Ken Thompson, be `Plan 9` Ken Thompson!

> Ken Thompson was once asked what he would do differently if he were redesigning the UNIX system. His reply: "I'd spell creat with an e."
> Kernighan, Brian W.; Pike, Rob (1984). The UNIX programming environment. Prentice-Hall. ISBN 0139376992, p. 204.
>
> [in Plan 9, he got his wish ;-)](https://github.com/akavel/plan9/blob/master/sys/include/9p.h#L185)

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,418 @@
...
+          icon: 'save',
+          label: get('save'),
+          action: ConnectedAPEditor.creatAction(element),
+        },
+        content: [ConnectedAPEditor.renderWizardPage(element)],
+      },
+    ];
+  }
+
+  static isEqualAddress(oldAddr: Element, newAdddr: Element): boolean {
```
c-dinkel, line 213: Why are these `static` methods? Could some of them be hoisted outside of the class as functions or is this for testing?

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,418 @@
...
+      Array.from(oldAddr.querySelectorAll(selectors.Address + ' > P')).filter(
+        pType =>
+          !newAdddr
+            .querySelector(`Address > P[type="${pType.getAttribute('type')}"]`)
+            ?.isEqualNode(pType)
+      ).length === 0
+    );
+  }
+
+  static editAction(parent: Element): WizardAction {
```
c-dinkel, line 224: This seems pretty complex, do you think you could refactor out a subroutine or two?

src/editors/communication/foundation.ts
```diff
@@ -0,0 +1,106 @@
...
+  h2 > nav,
+  h3 > nav,
+  h1 > abbr > mwc-icon-button,
+  h2 > abbr > mwc-icon-button,
+  h3 > abbr > mwc-icon-button {
+    float: right;
+  }
+
+  abbr[title] {
+    border-bottom: none !important;
```
c-dinkel, line 102: Try to find a more specific `css` selector to use instead of `!important`, it's [considered bad style for pretty good reasons](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception)

src/editors/communication/subnetwork-editor.ts
```diff
@@ -0,0 +1,417 @@
...
+  oldBitRate: Element | null,
+  BitRate: string | null,
+  multiplier: string | null,
+  SubNetwork: Element
+): EditorAction {
+  if (oldBitRate === null)
+    return {
+      new: {
+        parent: SubNetwork,
+        element: new DOMParser().parseFromString(
```
c-dinkel, line 53: Please use `createElement` over `DOMParser` always in actions to get elements with the correct `ownerDocument`

src/foundation.ts
```diff
@@ -337,3 +347,10 @@ declare global {
     ['log']: LogEvent;
   }
 }
+export const unique = (
```
c-dinkel, line 350: This does not report whether `value` is unique within `self` (could be named `array`, `within`, or something like that?).

`indexOf(value) === `index` only tells you if `index` is the first place where `value` appers, not if it's the only one. If that's the intended behaviour, maybe rename the function to `first`?

Also, is this actually used anywhere? If so, are they few enough places that this could be inlined?

c-dinkel (changes requested):

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,418 @@
...
+  WizardAction,
+  WizardInput,
+  newActionEvent,
+  compareNames,
+  getValue,
+  createElement,
+  ComplexAction,
+} from '../../foundation.js';
+
+import { Checkbox } from '@material/mwc-checkbox';
```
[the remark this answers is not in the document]
JakobVogelsang: Good hint thank you. I totally agree. With web-componets we should make certain the components we need are importet within the web-component. I also rearanged the import statements to be better readable.

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,418 @@
...
+}
+
+/** Description of a `ListItem` representing an `IED` and `AccessPoint` */
+interface itemDescription {
+  value: apAttributes;
+  connected?: boolean;
+}
+
+/** Sorts disabled `ListItem`s to the bottom. */
+function compare(a: itemDescription, b: itemDescription): number {
```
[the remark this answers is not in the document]
JakobVogelsang: What about `compareListItemConnection`? Too long?

src/editors/communication/connectedap-editor.ts
JakobVogelsang: :) Thank you Ken Thompson for appreciating the smaller things in live.

src/editors/communication/connectedap-editor.ts
JakobVogelsang: The pattern is used in all editors now. Initialy we used functions and switched to static method to better unit test.

src/editors/communication/connectedap-editor.ts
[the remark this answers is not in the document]
c-dinkel: Since the thing we're comparing are `itemDescription`s and not `Connection`s, `compareItemDescriptionByConnected` would be more accurate, but that absolutely seems too long for a non-exported local function. So if you want to repeat the types of the operands in the name rather than say what the function does, `compareItemDescriptions` might work e.g.  In general I think this is not that important and `compare` can stand.

However: Why is the type name `itemDescription` lowercase?

src/editors/communication/connectedap-editor.ts
c-dinkel: If it's necessary for testing we can absolutely keep doing this, but I wouldn't go so far as to call it a pattern, and I will continue to try to export regular functions instead of embedding them as static methods wherever possible, see e.g. [the first answer to this question about C++](https://stackoverflow.com/questions/1434937/namespace-functions-versus-static-methods-on-a-class).

src/editors/communication/foundation.ts
JakobVogelsang: We don't need that anyway. `themes` covers it!

src/editors/communication/connectedap-editor.ts
```diff
@@ -254,10 +275,9 @@ export class ConnectedAPEditor extends LitElement {
 
       const oldAddress = parent.querySelector(selectors.Address);
 
-      if (
-        oldAddress !== null &&
-        !ConnectedAPEditor.isEqualAddress(oldAddress, element)
-      ) {
+      if (oldAddress !== null && isEqualAddress(oldAddress, newAddress)) {
+        // INFO: We cannot use updateAction on address as bot address
```
c-dinkel, line 279: That comment is really helpful, thanks! The all-caps word `INFO: ` in the beginning draws a lot of attention though. Could we leave it out?

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,437 @@
...
+            parent,
+            element: oldAddress,
+            reference: oldAddress.nextElementSibling,
+          },
+        });
+        complexAction.actions.push({
+          new: {
+            parent,
+            element: newAddress,
+            reference: parent.firstElementChild?.nextElementSibling ?? null,
```
c-dinkel, line 291: Why do we always want to use the second child element as reference? Is it correct not to use `oldAddress.nextElementSibling` here?

src/editors/communication/connectedap-editor.ts
```diff
@@ -0,0 +1,437 @@
...
+    ): EditorAction[] => {
+      const instType: boolean =
+        (<Checkbox>wizard.shadowRoot?.querySelector('#instType'))?.checked ??
+        false;
+
+      const newAddress = createAddressElement(inputs, parent, instType);
+
+      const complexAction: ComplexAction = {
+        actions: [],
+        title: `Edit Address (${parent.getAttribute('iedName')})`,
```
c-dinkel, line 273: This needs to be translated. Are there any other user visible English language strings in this commit that might still require translations?

c-dinkel (changes requested):

src/editors/communication/connectedap-editor.ts
```diff
@@ -270,13 +270,17 @@ export class ConnectedAPEditor extends LitElement {
...
+          apName: parent.getAttribute('apName') ?? '',
+        }),
       };
 
       const oldAddress = parent.querySelector(selectors.Address);
 
-      if (oldAddress !== null && isEqualAddress(oldAddress, newAddress)) {
-        // We cannot use updateAction on address as bot address child elements P are changed
+      if (oldAddress !== null && !isEqualAddress(oldAddress, newAddress)) {
+        // INFO: We cannot use updateAction on address as both address
```
c-dinkel, line 282: 👎

c-dinkel (approved):
LOTM

PR #122, opened 2021-02-12
fix(themes): add list-item text colors

### LINKED ISSUE #80: Dark theme does not adapt color in drawer
Brough in with #79 . Unfortunately there is a bug in the library we are using. We are waiting for the bug fix here and are confident that the bug will be resolved in near future.

c-dinkel:
Resolves #80 .


JakobVogelsang (approved):
LGTM

PR #131, opened 2021-02-17
fix(lnodewizard): client lns saved with empty string instead of Clien…

JakobVogelsang:
…t LN


JakobVogelsang:
The bug was to write `Client LN` into `LNode`'s `ldInst` attribute for logical nodes allocated in the access point directly. `Client LN` is a pure display name now and an empty string is set for logical nodes allocates in the access point directly.

c-dinkel (approved):
LGTM

PR #136, opened 2021-02-27
feat(logging): add buttons to filter log messages by kind

### LINKED ISSUE #79: Log Information
Hey,
I am not sure If there is a purpose to keep the System behavior in the Log section as well.
But since I save the project, or change the language in the left toolbar, It didn't appear in the logs.

Additionally 
2020-12-18 11_25_54-OpenSCD and 1 more page - Work - Microsoft​ Edge
 It would be better to have white text on the toolbar when you turn on the dark theme. Its nearly not visible

JakobVogelsang:
closes #79


src/open-scd.ts
```diff
@@ -567,6 +568,39 @@ export class OpenSCD extends Setting(
...
+      min-width: 55vw;
+      float: left;
+    }
+
+    #filterContainer > mwc-icon-button-toggle {
+      --mdc-icon-button-size: 36px;
+    }
+
+    #filterContainer > mwc-icon-button-toggle:nth-child(1) {
+      position: absolute;
```
c-dinkel, line 585: If we position absolutely, let's render the icons in the dialog header on the right hand side.

c-dinkel (changes requested):

src/Logging.ts
```diff
@@ -52,9 +55,18 @@ export function Logging<TBase extends LitElementConstructor>(Base: TBase) {
     /** Index of the last [[`EditorAction`]] applied. */
     @property({ type: Number })
     currentAction = -1;
+    @property({ type: Number })
```
c-dinkel, line 58: This is really `type: Boolean`, maybe make these fields `private` `internalProperty`s?

c-dinkel (changes requested):

c-dinkel (approved):
LGTM

PR #137, opened 2021-02-27
fix(logging): do not log results of long running processes twice

### LINKED ISSUE #113: Validation result is logged twice
## How to reproduce

Open a project, thereby triggering validation.

## Expected result

The validation result should be logged only once.

## Actual result

The validation result gets logged twice.

JakobVogelsang:
close #113

### DIFF
--- a/src/Importing.ts
+++ b/src/Importing.ts
@@ -234,7 +234,7 @@ export function Importing<TBase extends LitElementConstructor>(Base: TBase) {
     }
 
     /** Loads and parses an `XMLDocument` after [[`srcIED`]] has changed. */
-    async importIED(src: string, doc: Document): Promise<string> {
+    async importIED(src: string, doc: Document): Promise<void> {
       let iedDoc: Document | null = null;
 
       const response = await fetch(src);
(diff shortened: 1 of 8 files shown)


src/Importing.ts
```diff
@@ -248,7 +248,13 @@ export function Importing<TBase extends LitElementConstructor>(Base: TBase) {
...
           })
         );
-        throw new Error(get('import.log.loaderror'));
+        this.dispatchEvent(
+          newLogEvent({
+            kind: 'error',
+            title: get('import.log.loaderror'),
+          })
+        );
+        return;
```
c-dinkel, line 257: Reject by `throw new Error();` or `throw new Error(get('import.log.loaderror'));`

src/Waiting.ts
```diff
@@ -29,8 +29,8 @@ export function Waiting<TBase extends LitElementConstructor>(Base: TBase) {
       this.work.add(e.detail.promise);
       this.workDone = Promise.allSettled(this.work);
       await e.detail.promise.then(
```
c-dinkel, line 31: Instead use `promise.catch(reason => console.warn(reason))`

c-dinkel (changes requested):

c-dinkel (approved):
LGTM

PR #160, opened 2021-03-18
fix: allow re-opening the same file again

### LINKED ISSUE #7: Allow re-opening of the same file
`<open-scd>` is currently only listening to `#file-input`'s `change` event, meaning a file is not loaded if it's the same file as before.

We could either implement a "reload file" button parsing the same `File` object again or listen to something like `#file-input`'s `click` event, checking whether the selected file has changed, and asking whether to reload if not.

c-dinkel:
This finally closes #7 , one of our most long-standing issues 🥳

### DIFF
--- a/src/open-scd.ts
+++ b/src/open-scd.ts
@@ -505,12 +505,16 @@ export class OpenSCD extends Setting(
         </div>`
       }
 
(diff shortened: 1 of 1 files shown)


JakobVogelsang (approved):
LGTM

PR #170, opened 2021-04-12
feat(communicationmapping): create control block connections with clientln

### LINKED ISSUE #134: Manage Report service connections
As a user of OpenSCD I want to connect static reports to a client receiver. I want to see the if a report has a `ClientLN` and want to edit or create new `ClientLN` connections

**Requirement**:
- Connection of several logical nodes to a report control block instead of connecting each logical node one after another shall be possible to speed up this engineering step.
- Highlight already connected logical nodes for a report control block.

**Type Definition**:
`ClientLN` definitions are set and interpreted by OpenSCD directly and not by the user via input. The validity therefore is done indirectly by validating other attributes as the IED name etc.

**Restriction** :
- `iedName` may not be empty
- `ldInst` is required and shall be set to `LD0` for logical nodes reside outside a server [Tissue1395](https://iec61850.tissue-db.com/tissue/1395)
- `prefix` can be missing or empty string when prefix in logical node is missing or empty. OpenSCD `shall` set an empty string in this case
- `lnInst` is required. Missing string for `LLN0`

**Edition 1 vs Edition 2**
- `ClientLN`'s attribute `prefix` was always required but could be empty string for logical node outside of server (pure client IED's) in Edition 1. This was changes to a recommendation to set `LD0` in this case in Edition 2. In Edition 2.1 `LD0` must be set for this case.

JakobVogelsang:
closes #134


c-dinkel (approved):
"Disconnect" needs to be translated, otherwise LGTM

Issue #172, opened 2021-04-13
Update Substation does not work with missing substation in the existing project


JakobVogelsang:
It is not possible to deal with the following scenario:
1.) You have a customer file where you have added a substation section.
2.) The customer gave you an updated file - of course without substation section
3.) You want to import the substation section to the new updated customer file

No wizard is opening.

PR #173, opened 2021-04-14
fix(communicationmapping): ClientLN handling for LNs reside in AccessPoint

### LINKED ISSUE #171: Duplicate ClientLN
When creating a report control block connection the check for existing `ClientLN` is not correct. This is only the case for logical nodes residing in in AccesPoint. For older versions the standard clearly says that `then the value of ldInst can be arbitrary `. At the moment ´arbitrary´ is restricted to empty string or LD0. This collides with older SCL file editions.

JakobVogelsang:
Closes #171 .


c-dinkel (approved):
This fix LGTM, but CommunicationMapping looks like it could use some more refactoring in general.

c-dinkel:
> Closes #172

I don't think this really closes #172 . Which ticket does this close instead?

PR #180, opened 2021-04-22
Implement getreference function

### LINKED ISSUE #166: LNode's are not properly added to the SCL
When adding `LNode`'s to the SCL the order is of importance. Please take into account the sequence definition in the schema.

c-dinkel:
Closes #166 .


c-dinkel (commented):
This feature looks great in general, but the `tags` structure needs more review for completeness and correctness. I'll try to help with that asap.

PR #186, opened 2021-04-27
feat(logging): show snackbars on info and warning

### LINKED ISSUE #184: Add info and warning snackbars
As a user I want to be notified of successful actions with a quick popup message, not just of error messages.

### Example

I want to be notified of a successful `Import IED` action with an "IED imported successfully." message.

c-dinkel:
Closes #184 .

I did not manage to regenerate test snapshots and would appreciate help with this.


JakobVogelsang (commented):
LGTM

PR #187, opened 2021-04-27
feat(filtered-list): add check all to filtered-list

### LINKED ISSUE #165: Select all - feature wish
Hello, it would be very helpful with a select all function when selecting logical nodes in OpenSCD. 
Then you can just de-select whatever nodes you don't need and save some time  

image

JakobVogelsang:
Closes #165


c-dinkel (approved):
LGTM

PR #189, opened 2021-04-29
feat(communication-editor): show all connected ap within ied

### LINKED ISSUE #129: Highlight IEDs with multiple Access Points
For the IEDs which have multiple Access points, it would be nice if clicked on the IED in AP1, the other AP of IED in another subnetwork is "highlighted" to let the user recognize that : "it is actually the same IED with more than one APs".

APs

Hope it is clear. Thx!

JakobVogelsang:
Closes #129


c-dinkel (approved):
LGTM

PR #192, opened 2021-05-01
feat(subscriberinfo): add subscriber info to sampledvaluecontrol

### LINKED ISSUE #182: Update subscriber incorrect for Edition 1 files
The element `IEDName` that is added to file with `Update Subscriber Info` does not take into account the different structures of `IEDName` in Edition 1 and Edition 2 and above. Edition 1 does define this as such:

```
<xs:element form="qualified" maxOccurs="unbounded" minOccurs="0" name="IEDName" type="tName"/>
```

From Edition 2 this element can have additional attributes specifying the subscriber location in more detail. 

At the moment this functionality in Edition 1 files produce invalid files.

JakobVogelsang:
Closes #182

### DIFF
--- a/src/foundation.ts
+++ b/src/foundation.ts
@@ -2437,6 +2437,15 @@ export function findControlBlocks(extRef: Element): Set<Element> {
   return controlBlocks;
 }
 
+/** @returns the version of the SCL project */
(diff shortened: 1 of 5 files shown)


c-dinkel (approved):
LGTM

PR #196, opened 2021-05-12
feat(plugin): reload official plugins without reset

### LINKED ISSUE #176: Load new plugins without reset

JakobVogelsang:
Closes #176


public/js/plugins.js
```diff
@@ -0,0 +1,50 @@
+export const officialPlugins = [
```
c-dinkel, line 1: Is there a reason this file is `js` and not `ts`? For customisability without recompilation?

c-dinkel (approved):
LGTM

PR #206, opened 2021-06-02
refactor(triggered): make "Open project" a plugin

c-dinkel:


public/js/plugins.js
```diff
@@ -1,4 +1,11 @@
 export const officialPlugins = [
+  {
+    name: 'Open file',
+    src: '/src/triggered/OpenFile.js',
+    icon: 'coronavirus',
```
dlabordus, line 5: Is this correct, "coronavirus"? Or some typing in the wrong editor?

c-dinkel: Sorry, that was a placeholder icon that we never updated.

c-dinkel:
This is ready to be merged without any unit tests, since the new plugin consists entirely in completely untestable Browser API code.

https://github.com/com-pas/compas-open-scd/tree/open-compas-plugin is waiting for this.

src/open-scd.ts
```diff
@@ -193,7 +171,6 @@ export class OpenSCD extends Setting(
     if (ctrlAnd('z')) this.undo();
     if (ctrlAnd('l')) this.logUI.open ? this.logUI.close() : this.logUI.show();
     if (ctrlAnd('m')) this.menuUI.open = !this.menuUI.open;
-    if (ctrlAnd('o')) this.fileUI.click();
```
dlabordus, line 196: Christian, maybe discus if this is a problem. Users are now not able to open a project anymore using ctrl+o.

dlabordus (commented):
See my comment on the ctrl+o.

src/open-scd.ts
c-dinkel: I added a shortcut that opens the first `loader` plugin on `ctrl+o`

JakobVogelsang (approved):
LGTM

public/js/plugins.js
```diff
@@ -21,11 +21,18 @@ export const officialPlugins = [
...
+    icon: 'folder_open',
+    default: true,
+    kind: 'loader'
+  },
+  {
+    name: "Import IEDs",
+    src: "/src/triggered/ImportIEDs.js",
+    icon: "snippet_folder",
+    default: true,
+    kind: "triggered"
```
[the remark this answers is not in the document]
c-dinkel: Loader is currently the type of plugin displayed at the very top of the menu and clickable without a `doc` loaded. Import IEDs should only be available when a doc is already loaded and could just stay at the top of the `triggered` plugin section.

PR #209, opened 2021-06-05
Add basic dotype manipulation capability

JakobVogelsang:


src/editors/templates/dotype-wizards.ts
```diff
@@ -0,0 +1,724 @@
...
+        label: get('add'),
+        action: addPredefinedDOType(parent, templates),
+      },
+      content: [
+        html`<mwc-select
+          fixedMenuPosition
+          outlined
+          icon="playlist_add_check"
+          label="values"
+          helper="Default enumerations"
```
c-dinkel, line 535: This may need a translation if it needs to stay in.

c-dinkel (changes requested):
Found one missing translation, looks very good to me otherwise, clean code and an easy read.

src/editors/templates/dotype-wizards.ts
c-dinkel: Helper has been translated.

c-dinkel (approved):
LGTM

PR #229, opened 2021-06-20
feat(validators/validatetemplates): add validation based on NSD files

### LINKED ISSUE #222: DataTypetemplate validator
As a user of OpenSCD I want to validate the templates defined in the `DataTypeTemplates` section against `IEC 61850-7-2`, `IEC 61850-7-3`, `IEC 61850-7-4` and `IEC 61850-8-1`.

**Features**
- Export error to log with following information: error description, filename, line, part of IEC (e.g. 61850-7-4) that defines the rule

**Error description**
- Logical node type (LNodeType) does have all mandatory data objects: ... is missing
- Data object type (DOType) does not have all mandatory data attributes: ... is missing
- Common data class does not follow the definition in IEC 61850-7-3
- Missing control type data attribute structures ... for ctlModel ...
- Invalid control data structure ... : data attribute ... is missing
- Invalid enumeration definition: EnumVal ... is incorrect
- Invalid constructed data attribute definition ... : ... is missing
- Invalid constructed data attribute definition ... : ... is invalid

**Restrictions**
- The validator shall only be created for the newest version of SCL files in the beginning `2007B4`. 
- Use the following files: `IEC_61850-7-4_2007B3`, `IEC_61850-7-3_2007B3`, `IEC_61850-7-2_2007B3` and `IEC_61850-8-1_2003A`

JakobVogelsang:
Closes #222


public/js/plugins.js
```diff
@@ -48,6 +48,13 @@ export const officialPlugins = [
     default: true,
     kind: 'validator'
   },
+  {
+    name: 'Validate data',
```
c-dinkel, line 52: As a novice user, a more descriptive name would help me identify the DataTypeTemplates as the target of this validation.

Maybe change to `'Validate DataTypeTemplates'` or similar?

src/validators/ValidateTemplates.ts
```diff
@@ -0,0 +1,418 @@
...
+
+const iec6185072 = fetch('public/xml/IEC_61850-7-2_2007B3.nsd')
+  .then(response => response.text())
+  .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+
+const iec6185081 = fetch('public/xml/IEC_61850-8-1_2003A2.nsd')
+  .then(response => response.text())
+  .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+
+const serviceSCD = ['SPC', 'DPC', 'INC', 'ENC', 'BSC', 'ISC', 'APC', 'BAC'];
```
c-dinkel, line 20: Why the `SCD` in the name? Could we rename to `const services`?

src/validators/ValidateTemplates.ts
```diff
@@ -0,0 +1,418 @@
...
+
+async function validateCoOperStructure(oper: Element): Promise<LogDetail[]> {
```
c-dinkel, line 22: Why the "`Co`" in all these function names?

c-dinkel (approved):
LGTM

Very clean, this was a pleasure to read, and above all: great functionality! 🎉

src/validators/ValidateTemplates.ts
JakobVogelsang: `CO` is one of the functional constraints defined in the standard and stands for control.

src/validators/ValidateTemplates.ts
JakobVogelsang: This is wrong, my apologies. The correct term is `serviceCDCs`. The term is used with the *.nsd files and stands for Common Data Classes that have the ability to be controlled.

public/js/plugins.js
JakobVogelsang: `Validate DataTypeTemplates` would exceed the width of the menu. But `Validate data` is not precise enough, I would agree. I instead would change is to `Validate Templates`. At least in with in OpenSCD we use `Templates` for `DataTypeTemplates` anyway. What do you think?

PR #231, opened 2021-06-23
fix(wizard-dialog): remove button actions only after action was successful

dlabordus:
Christian, 

The fix for the bug we found, when we don't want to close the wizard, for instance because a field was incorrect, and no WizardActions where returned, the action on the button should not be removed.


JakobVogelsang (changes requested):
The fix is very fine, thank you for the contribution. I suggested two test so that in the future we will not do the same mistake in a [PR](https://github.com/com-pas/compas-open-scd/pull/20)

JakobVogelsang (commented):
LGTM

JakobVogelsang:
Thanks you!!

PR #237, opened 2021-07-01
Refactor plugins

JakobVogelsang:
We wanted to add the about-box as another plugin and realized that the plugin API might need another refactor.

@dlabordus you might want to review as well.


ca-d (approved):
LGTM

PR #242, opened 2021-07-08
refactor(wizards): move wizards in substation editor to wizard library

JakobVogelsang:


ca-d (approved):
Looks great, thanks!

Maybe sometime in the future we could unify the naming for all tag related functions and data.

PR #246, opened 2021-07-11
Add git plugin

### LINKED ISSUE #243: Add web component to navigate in tree stuctures
As a user of OpenSCD I want to effectively browse through the data model of an IED, in case I want to add data to a control block or just for the sack of it. 

For better user experience, the component shall have a search bar, that in the first shot shall filter for LN classes.
Based on the experience with other tree like implementations, we need to lazy load below LN level. Otherwise, loading might take very long.

ca-d:
Resolves #243 .

Also adds the git version control system as a plugin and a local IndexedDB filetype as a usage example for the hierarchical navigation widget `finder-pane` .


ca-d:
There will be another pull request for this later, after the filesystem integration has been turned into a plugin.

PR #248, opened 2021-07-12
refactor: add finder-pane for tree nagivation

### LINKED ISSUE #243: Add web component to navigate in tree stuctures

ca-d:
Resolves #243 .

### DIFF
--- a/__snapshots__/finder-pane.md
+++ b/__snapshots__/finder-pane.md
@@ -1,0 +1,189 @@
+# `finder-pane`
+
+#### `displays nothing with default properties`
+
+```html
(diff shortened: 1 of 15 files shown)


JakobVogelsang (approved):
LVeryGTM 👍

PR #264, opened 2021-07-20
fix(plugging): fix plugin auto-update

### LINKED ISSUE #253: fix(plugging): repair broken plugin auto-update mechanism
Hi OpenSCD team,
Is it only for me that,
in Firefox at the moment there is no "open project" and "new project" main screen appears.
image

and also in Chrome;
Even new project is created as Ed.2 then there is no "Add substation" button visible. And the loading bar below ribbon continuously waving.
image

same happens in Edge.

Could you quickly fix this issues or show me a workaround at least for one browser?
Best Regards
Burak

ca-d:
Closes #253 .

### DIFF
--- a/src/Plugging.ts
+++ b/src/Plugging.ts
@@ -41,7 +41,7 @@ export const pluginIcons: Record<PluginKind | MenuPosition, string> = {
   bottom: 'play_circle',
 };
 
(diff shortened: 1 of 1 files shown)


JakobVogelsang (approved):
LGTM

PR #265, opened 2021-07-20
Test errors in firefox but not in chrome

### LINKED ISSUE #117: Test errors in Firefox but not in Chrome
The [FastCheck assisted test](http://localhost:9876/debug.html?grep=lnodewizard%20defines%20a%20existLNode%20function%20that)

```
lnodewizard
    defines a existLNode function that
        returns true on existing LNode references in the substation element
```

fails with counterexamples like `["A0","A",null,"ANNW",""]` in Firefox but not in Chrome.

@JakobVogelsang do you want to include Firefox in our `karma.conf.cjs` so Travis CI can catch bugs like this one in the future? I think we should keep Firefox support alive if possible.

JakobVogelsang:
Closes #117

### DIFF
--- a/src/editors/templates/lnodetype-wizard.ts
+++ b/src/editors/templates/lnodetype-wizard.ts
@@ -33,7 +33,7 @@ import { ListItem } from '@material/mwc-list/mwc-list-item';
 import { Select } from '@material/mwc-select';
 import { SingleSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
 import { Switch } from '@material/mwc-switch';
-import { Formfield } from '@material/mwc-formfield';
+import { isAbsolute } from 'path/posix';
 
 function updateDoAction(element: Element): WizardActor {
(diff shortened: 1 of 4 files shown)


src/editors/templates/lnodetype-wizard.ts
```diff
@@ -33,7 +33,7 @@ import { ListItem } from '@material/mwc-list/mwc-list-item';
 import { Select } from '@material/mwc-select';
 import { SingleSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
 import { Switch } from '@material/mwc-switch';
-import { Formfield } from '@material/mwc-formfield';
+import { isAbsolute } from 'path/posix';
```
ca-d, line 36: Where does this dependency come from?

ca-d (changes requested):
Seems like some node module (`path/posix`) made it into our browser code.

Looks good otherwise.

PR #266, opened 2021-07-21
fix(logging): remove superfluous invisible reset filter

ca-d:
The "reset" LogEvent kind accidentally introduced a new filter `mwc-icon-button` to the log dialog, which was invisible (no icon), but prevented users from filtering for only "info" kind log entries.


JakobVogelsang (approved):
LGTM

PR #269, opened 2021-07-26
fix(validatetemplates): don't require "Oper" if ctlModel=status-only

### LINKED ISSUE #268: fix(validatetemplates): don't require "Oper" if ctlModel=status-only
There should already be some untested non-working code that's supposed to ensure this in `ValidateTemplates.ts` .

ca-d:
Closes #268 .

### DIFF
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -71,9 +71,9 @@ export const de: Translations = {
       title: 'Templates validieren',
       mandatoryChild:
         '{{ tag }} {{ id }} fehlt ein obligatorisches {{ childTag }}-Kind {{ childId }}',
-      missingCoDO: 'Control Datenobjekt vom Typ {{ type }} fehlt',
-      cannotValidate:
-        'Kann {{ childTag }} {{ childId }} in {{ tag }} {{ id }} nicht validieren',
+      missingCoDO: 'Control-Dienst Datenattribut vom Typ {{ type }} fehlt',
+      missingCtlModelDef:
+        'CtlModel Definition fehlt. Kann {{ childTag }} {{ childId }} in {{ tag }} {{ id }} nicht validieren.',
     },
   },
   textfield: {
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -59,12 +59,22 @@ export const de: Translations = {
         '{{ parent }} enthält bereits ein {{ child }} Kind namens "{{ name }}"',
     },
   },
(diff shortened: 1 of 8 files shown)


src/translations/de.ts
```diff
@@ -59,12 +59,22 @@ export const de: Translations = {
...
+      valid: '{{ name }} erfolgreich validiert',
+      invalid: '{{ name }} Validierung fehlgeschlagen',
+      fatal: 'Fataler Validierungsfehler',
+      loadError: 'Konnte Schema {{ name }} nicht laden',
+    },
+    templates: {
+      title: 'Templates validieren',
+      mandatoryChild:
+        '{{ tag }} {{ id }} fehlt ein obligatorisches {{ childTag }}-Kind {{ childId }}',
+      missingCoDO: 'Control Datenobjekt vom Typ {{ type }} fehlt',
```
JakobVogelsang, line 74: I had it already incorrect and as you put so much effort into it: Cancel, SBO, SBOw and Oper are data attributes. I would suggest the following:
`Control-Service Datenattribut vom Typ {{type}} fehlt`

src/translations/en.ts
```diff
@@ -90,12 +90,22 @@ export const en = {
...
+      valid: '{{ name }} validation successful',
+      invalid: '{{ name }} validation failed',
+      fatal: 'Fatal validation error',
+      loadError: 'Could not load schema {{ name }}',
+    },
+    templates: {
+      title: 'Validate templates',
+      mandatoryChild:
+        '{{ tag }} {{ id }} is missing mandatory child {{ childTag }} {{ childId }}',
+      missingCoDO: 'Control DO of type {{ type }} is missing',
```
JakobVogelsang, line 105: Would suggest here:
`Control service related data attribute {{type}} is missing`

src/translations/en.ts
```diff
@@ -90,12 +90,22 @@ export const en = {
...
+      cannotValidate:
+        'Cannot validate {{ childTag }} {{ childId }} within {{ tag }} {{ id }}',
```
JakobVogelsang, line 107: How about:
`Cannot validate ... {{id}}:{{reasonCode}}`
This would allow adding something like ctlModel is not defined.

JakobVogelsang (commented):
Thanks for the effort with this most likely annoying bug fix! 
Some strings can be more precise, but other than that, LGTM.

Issue #203, opened 2021-06-01
Max. Number of Clients check for ClientLN assignment of Reports


BurTah00:
Hi,
In a training with colleagues from US & LA, we discussed this feature.

From SCL it is clear to get the info for maximum number of clients for Client LN assignment to a report control block instance.
OpenSCD should only allow up to this number for client LN assignment of an RCB, not more.

Thanks!

JakobVogelsang:
This is very true and will come with other functional validation in the IED section. How urgent is this issue? We plan to add this Aug/Sep.

BurTah00:
no so high priority. Just to verify that it'll be implemented.

ca-d:
> no so high priority. Just to verify that it'll be implemented.

@JakobVogelsang added this last week, the feature is part of the current version 🎉 .

PR #270, opened 2021-07-27
feat(communicationmapping): move to ied-editor

### LINKED ISSUE #225: Move schema validator to plugin
Move schema validator to plugin of a `validator` kind.

### LINKED ISSUE #260: Move communication mapping to ied-editor

JakobVogelsang:
Closes #260 #203


src/zeroline/substation-editor.ts
```diff
@@ -125,7 +125,7 @@ export class SubstationEditor extends LitElement {
         ${this.renderHeader()}
```
ca-d, line 125: Since we support multiple `Substation` sections now, the header could probably use a `move` button.

ca-d (changes requested):
A move Substation button is the only thing that's obviously missing, UX feedback will be forthcoming as soon as I've done manual testing.

src/zeroline/bay-editor.ts
```diff
@@ -129,7 +129,7 @@ export class BayEditor extends LitElement {
       <div>
         ${until(
```
ca-d, line 130: Do you think it would be worth trying this synchronously again now that we have a faster algorithm for placing IEDs? If the performance is acceptable, this would remove the flickering when manually requesting Updates, e.g. in wizards.

ca-d:
The `ied-editor` doesn't appear clickable because of the following typo:

```css
 #container:hover > .icom {
 ```
 
 (should be `.icon` instead of `.icom`)

**Comment by ca-d** [comment]
I suggest we use two separate `mwc-icon-button`s in the `zeroline-pane` for connecting and disconnecting reports, in order to save a click and remove the "connect report" primary action from the "Communication Mapping" wizard.

Then we would have:
* One button in `zeroline-pane` for disconnecting any report (like it is now, but without the "connect" action)
* One button per `ied-editor` for disconnecting reports involving it (like it is now, but without the "connect" action)
* One button in `zeroline-pane` for connecting any two IEDs with reports (to be refactored into per-`ied-editor` buttons later)

**Comment by ca-d** [comment]
The `ied-editor` currently still displays the default text caret [cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) and allows [user selection](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) of the icon text (developer_board).

File: test/integration/editors/triggered/CommunicationMappingPlugin.test.ts [hunk 3]
```diff
@@ -1,39 +1,32 @@
...
-      .then(str => new DOMParser().parseFromString(str, 'application/xml'));
-    element.doc = doc;
+    element = <ZerolinePane>parent.querySelector('zeroline-pane')!;
     await element.updateComplete;
   });
 
   describe('communication mapping wizard', () => {
     beforeEach(async () => {
-      await element.run();
+      await element.commmap.click();
```
ca-d, line 29: We don't actually need to `await` a `click()`, since it doesn't return a `Promise`.

test/unit/wizards/clientln.test.ts
```diff
@@ -0,0 +1,319 @@
...
+        expect(
+          doc.querySelector(
+            'IED[name="IED2"] ReportControl[name="ReportEmpty"] ClientLN'
+          )
+        )?.to.not.exist;
+        (<List>(
+          parent.wizardUI.dialog?.querySelector('#sourcelist')
+        )).items[1].click();
+        await parent.updateComplete;
+        await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
ca-d, line 186: Could we try to await something like `parent.wizardUI.updateComplete` and then `parent.wizardUI.dialog.updateComplete` instead of these timeouts? The more of these setTimeout promises we use, the longer the test suite takes, and I have some hope that the actual animation may be slightly (~20ms) quicker than we guess here. It's not a huge difference in the individual case, but here we have many timeouts, so it might add up to a significant fraction of a second.

test/unit/wizards/controlwithiedname.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    )).click();
+    await parent.updateComplete;
+    expect(
+      doc.querySelector(
+        ':root > IED[name="IED2"] ExtREf[iedName="IED1"][ldInst="Disconnectors"][lnClass="XSWI"][doName="Pos"][daName="stVal"]'
+      )
+    ).to.not.exist;
+  });
+
+  it('updates selected ExtRefs in case intAddr is present', async () => {
```
ca-d, line 71: Again, great tests, but a few extra empty lines would help me in reading.

src/translations/en.ts
```diff
@@ -317,9 +318,9 @@ export const en = {
     nonewitems: 'no new IEDName elements to add',
     message: '{{updatenumber}} IEDName elements added to the project',
   },
-  commMap: {
+  commmap: {
```
ca-d, line 321: Was there a particular reason to change this back to `commap`, or just taste?

src/translations/en.ts
```diff
@@ -317,9 +318,9 @@ export const en = {
     nonewitems: 'no new IEDName elements to add',
     message: '{{updatenumber}} IEDName elements added to the project',
   },
-  commMap: {
```
ca-d, line 320: Was there a particular reason to change this back to `commap`, or just taste?

src/wizards/commmap-wizards.ts
```diff
@@ -0,0 +1,141 @@
...
+              selector(cbTag, cbId)
+            );
+            const [_, sourceIED, controlBlock] = cbId.match(/^(.+)>>(.*)$/)!;
+
+            return html`<mwc-list-item
+              twoline
+              graphic="icon"
+              hasMeta
+              @click="${(evt: SingleSelectedEvent) => {
+                evt.target!.dispatchEvent(
```
ca-d, line 115: 😍  This is much more readable than the original from the plugin, great job!

test/unit/wizards/clientln.test.ts
JakobVogelsang: This was not really necessary after rewriting the tests. :)

src/translations/en.ts
JakobVogelsang: Just a taste. I found it strange to use an upper case letter in such a short word.

src/wizards/commmap-wizards.ts
JakobVogelsang: :)

JakobVogelsang:
> 
> 
> I suggest we use two separate `mwc-icon-button`s in the `zeroline-pane` for connecting and disconnecting reports, in order to save a click and remove the "connect report" primary action from the "Communication Mapping" wizard.
> 
> Then we would have:
> 
>     * One button in `zeroline-pane` for disconnecting any report (like it is now, but without the "connect" action)
> 
>     * One button per `ied-editor` for disconnecting reports involving it (like it is now, but without the "connect" action)
> 
>     * One button in `zeroline-pane` for connecting any two IEDs with reports (to be refactored into per-`ied-editor` buttons later)

Your suggestion sounds already very good. The only thing that I have saved is the disconnecting on the IED level. Access to it is the communication anyway, and the extra effort and complexity on IED level seemed to be too much for the benefits.

__snapshots__/clientln connection wizards.md
```diff
@@ -0,0 +1,98 @@
...
+          style="height:56px"
+          tabindex="0"
+          value="IED1"
+        >
+          IED1
+        </mwc-list-item>
+        <mwc-list-item
+          aria-disabled="false"
+          mwc-list-item=""
+          style="height:56px"
```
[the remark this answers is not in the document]
JakobVogelsang: This wizard is deleted now!

ca-d (approved):
LGTM 🥳

PR #275, opened 2021-07-31
feat(abstractda): add val manipulation capability

### LINKED ISSUE #274: Add `Val` manipulation capanbility to BDA and DA wizard

### LINKED ISSUE #279: feat(da): add dchg, qchg and dupd to da wizard

### LINKED ISSUE #280: refactor(DAType and DOType wizards): move da and bda wizards to library

JakobVogelsang:
Closes #274 
Closes #280 
Closes #279 

This PR does not only add basic `Val` manipulation capability but also `dchg`, `qchg` and `dupd` - last three attributes of `DA`. I also found the pattern for the name of both `BDA` and `DA` was not correct.
I also take the chance and pushed the touched wizards to the wizard library.


src/wizard-select.ts
```diff
@@ -72,14 +72,7 @@ export class WizardSelect extends Select {
   }
 
   checkValidity(): boolean {
-    if (
-      this.reservedValues &&
-      this.reservedValues.some(array => array === this.value)
-    ) {
-      this.setCustomValidity(get('textfield.unique'));
```
ca-d, line 79: Great change 🎉 ! Does this mean we could also get rid of the `textfield.unique` translation, or do we still need it in the `wizard-textfield`?

src/wizards/abstractda.ts
```diff
@@ -0,0 +1,243 @@
...
+    html`<wizard-textfield
+      label="Val"
+      .maybeValue=${Val}
+      helper="${translate('scl.Val')}"
+      nullable
+    ></wizard-textfield>`,
+  ];
+}
+
+export function getValAction(
```
ca-d, line 211: Do you generally prefix functions like this with `get` or would

```ts
export function valAction(
```

also work for you?

src/wizards/bda.ts
```diff
@@ -0,0 +1,241 @@
...
+      else newElement.setAttribute('desc', desc);
+      newElement.setAttribute('bType', bType);
+      if (type === null) newElement.removeAttribute('type');
+      else newElement.setAttribute('type', type);
+      if (sAddr === null) newElement.removeAttribute('sAddr');
+      else newElement.setAttribute('sAddr', sAddr);
+      if (valKind === null) newElement.removeAttribute('valKind');
+      else newElement.setAttribute('valKind', valKind);
+      if (valImport === null) newElement.removeAttribute('valImport');
+      else newElement.setAttribute('valImport', valImport);
```
ca-d, line 63: Maybe we could think about refactoring blocks like this into a helper function in our `foundation` in the future. How common do you think these if/else blocks are going to be?

src/wizards/foundation.ts/enums.ts
```diff
@@ -0,0 +1,55 @@
+export const functionalConstraintEnum = [
```
ca-d, line 1: Why does the `wizards/foundation` directory carry a `.ts` extension?

ca-d (changes requested):
This looks and reads extremely well. The only thing that sticks out to me is the `foundation.ts` directory name which looks like a TypeScript file.

src/wizard-select.ts
JakobVogelsang: At the moment, we still use it to check for name attribute duplication.

src/wizards/abstractda.ts
JakobVogelsang: We do this at the moment for `Voltage` for example and will do so for `MinTime`, `MaxTime` ....
Those are child elements  that in our wizards appear as attributes. 

Working on the wizards, I found that we should talk about a naming convention here. I would therefore postpone this to [Issue](https://github.com/openscd/open-scd/issues/281)

src/wizards/bda.ts
JakobVogelsang: This is very true. This pattern is very common and can be avoided with a simple function [Issue](https://github.com/openscd/open-scd/issues/282)

src/wizards/foundation.ts/enums.ts
JakobVogelsang: This is a typo, sorry :)

ca-d (approved):
LGTM

PR #278, opened 2021-08-02
feat(gsecontrol): edit GSEControl and its referenced elements

### LINKED ISSUE #145: GOOSE control block manipulation capability
As a user of OpenSCD I want to manage `IED` configuration. I want to delete/update the `GSEControl` element. (GOOSE control block)

**Requirement**:
- `GSEControl` of the type `GSSE` is deprecated and therefore for update is not allowed. But those shall be displayed.
- On delete make sure that the referenced DataSet is removed too.
- On delete make sure that DataSet is not removed if referenced to another control block.
- On delete make sure to remove GSE element too if present.

**Type Definition**:
- TBD

**Restriction** :
- TBD

**Edition 1 vs Edition 2**
- TBD

JakobVogelsang:
Closes #145


src/wizards/foundation.ts/p-types.ts
```diff
@@ -0,0 +1,20 @@
+export const pTypesGSESMV: string[] = [
```
ca-d, line 1: This should move from `foundation.ts` to `foundation`.

ca-d (changes requested):
LGTM, apart from the `foundation.ts` directory name.

src/wizards/foundation.ts/p-types.ts
JakobVogelsang: Done!

PR #288, opened 2021-08-09
feat(filtered-list): select only visible items on checkAll

### LINKED ISSUE #285: Using filters during LN references
During the LN References to Substation elements, the Select all button beside the filter textbox creates the idea of selecting only the LN showed after the filter, but actually it selects all the LN within the DM and ignores the filter

JakobVogelsang:
Closes #285


ca-d (approved):
LGTM

PR #291, opened 2021-08-10
fix(ied-editor): add tooltip with full IED name

### LINKED ISSUE #289: Usability issues of new symbols (tooltips)
image
The names of the IEDs are not readable. 
Long names start with ... (three dots) and only the ends are shown. Expected to show the beginning of the word.
No tooltip for the names.
Please don't forget to add tooltips for any enhancement that requires reading something.
These two newly add signals also doesn't have tooltips.
Thanks!

JakobVogelsang:
Closes #289

### DIFF
--- a/src/zeroline/ied-editor.ts
+++ b/src/zeroline/ied-editor.ts
@@ -43,7 +43,9 @@ export class IedEditor extends LitElement {
   render(): TemplateResult {
     return html`
       <div id="container" tabindex="0">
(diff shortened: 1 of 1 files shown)


ca-d (changes requested):
This change should have made some tests red, at least some snapshot test.

Does this mean some tests went missing?

JakobVogelsang:
I think the ied-editor does not have a snapshot test, yet. As this is a small thing I would suggest to add with report or smv capability.

PR #296, opened 2021-08-27
first draft of a working drag and drop implementation

b-ono:
This is a working drag and drop implementation for the substation editor. Still missing things like unit tests and a review. Currently follows the same semantic as the current move implementation. 

I'm not too happy about the current drop implementation as it relies on the fact that the current editor tags have a direct relation to the SCL tags. An explicit way to specify that e.g. the voltage-level-editor edits the "VoltageLevel" SCL element would be better.

### DIFF
--- a/__snapshots__/bay-editor.md
+++ b/__snapshots__/bay-editor.md
@@ -4,7 +4,7 @@
 
 ```html
 <section tabindex="0">
-  <h3>
+  <h3 draggable="true">
     COUPLING_BAY — Bay
     <abbr title="[add]">
       <mwc-icon-button icon="playlist_add">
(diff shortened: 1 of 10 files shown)


ca-d:
> I'm not too happy about the current drop implementation as it relies on the fact that the current editor tags have a direct relation to the SCL tags. An explicit way to specify that e.g. the voltage-level-editor edits the "VoltageLevel" SCL element would be better.

There should be some way to find out which SCL element a particular editor is responsible for editing by checking something like `editor.element.tagName`.

JakobVogelsang (commented):
LvvvvvGTM. Thanks you for this nice function!!
Some tests are missing though ;) do you want to do it or is it ok when I/we add tests?

PR #312, opened 2021-09-16
Show History (HItem) from SCL File in log window with a separate toggle to show/hide them

dlabordus:
@JakobVogelsang,
The only problem still in there is the icons that are always shown disabled, but working. I hope you can find something that's causing this.
I create the branch in our forked repository, hopefully that doesn't make it hard to help with the icon problem.


JakobVogelsang:
Hi @dlabordus ,

I did have a closer look. I think there is a bug with the `mwc-icon-toggle-button`. see [issue](https://github.com/material-components/material-web/issues/2712). This is a strange bug that prevents the `svg` to jump from `onIcon` to `offIcon`. 

I will have a closer look tomorrow.



My guess is that the pathSVG for sclHistory is not properly defined. I found `<path .../>` to not work, but `<path ...></path>`. So don't know why there is a `rect` and why the namespace is part of the `svg` tag and `path`. But

JakobVogelsang:
Good morning @dlabordus,
I tried, in a step-by-step approach, to find out what went wrong with the `mwc-icon-button-toggle` and so have created a working branch. Would be curious if this works on your machine as well. The strange thing is that it is exactly your code...
Somehow I have the impression that this might a build issue with snowpack. Once the `mwc-icon-button-toggle` does go bananas, it does not come back. It even went crazy with a fresh  cloned repo...

dlabordus:
Good morning @JakobVogelsang 
I checkout your copy branch and build/run it. It still doesn't work on my machine. So it does look like some kind of build issue on my machine. I'm using the following versions of NodeJS and NPM.
- npm -version: 6.14.15
- node -v:  v14.17.6
I will try removing all NPM Cache and see if that helps....

JakobVogelsang:
Can you please look for the version of mwc-icon-button-toggle.

dlabordus:
Both in the logging and in the node_modules directory it says
- @material/mwc-icon-button-toggle@0.22.1
Removing all kind of caches did fix it. (node_modules directory, cache directory in .npm)

JakobVogelsang:
Thank you. I will report that to the mwc team.

JakobVogelsang:
Do you mind if we take my branch. Here a added some more tests and a clarification? Maybe you can have another look as a reviewer?

PR #328, opened 2021-10-03
refactor(zeroline/...-editor): add common editor-container component

JakobVogelsang:


JakobVogelsang:
> 
> 
> I kind of liked the old button grouping behavior where all the buttons that directly manipulate the editor's element will always stay next to each other in one single row (see right hand side in screenshot), it makes it easier to find the right control at a glance.
> 
> icon-button-nav-group
> 
> @JakobVogelsang do you agree or do you prefer the new behavior on the left hand side in above screenshot?

Yes it is, we might discuss if with the more vert option we could move some buttons there. The header becomes crowded :)

PR #331, opened 2021-10-06
Add finder list

### LINKED ISSUE #256: Multi finder-pane

ca-d:
This adds a new version of the `finder-pane` component with a simplified API and a new boolean "multi" attribute which may be set to allow the user to select many different paths at the same time, adding or removing selections at any level of the hierarchy. All chosen paths are conveniently reflected to the programmer through the `.selection` property.

Closes #256

### DIFF
--- a/src/finder-list.ts
+++ b/src/finder-list.ts
@@ -84,6 +84,14 @@ export class FinderList extends LitElement {
   @property({ attribute: false })
   loaded: Promise<void> = Promise.resolve();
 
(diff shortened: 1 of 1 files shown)


ca-d:
Selecting an entry does not always mark the selected entry as "activated".

JakobVogelsang:
Uff sorry Chris. I push the last commit directly into the PR other than discessed! :|

PR #338, opened 2021-10-26
Add finder list

### LINKED ISSUE #256: Multi finder-pane

ca-d:
Closes #256 .

This adds `finder-list`, a simplified version of the old `finder-pane` with an extra `multi` option to enable selecting multiple paths.


JakobVogelsang (approved):
Thanks for this great job!!!

If any I have some remarks that are a matter of taste:
- the method `getPaths` is a bit tough to understand, is very compact. Maybe there is a way to make it more readable
- I am not sure if we need a filtered list for children < 5. At the moment the search field of the filtered list is nice as it tells the beginning of the list. Maybe this could be done with a line or a none interactive list item.

PR #347, opened 2021-11-02
ci: add github actions for test, build and deploy

### LINKED ISSUE #348: Move from travis to github actions
Travis has changed its business model, and we cannot run test, build scripts as well as deploy to openscd.github.io at the moment. The OpenSCD team has therefore decided to move to GitHub actions.

JakobVogelsang:
Closes #348 

Some remarks to the scripts:
- **test**: runs only on push, except pushes on main.
- **test-and-build**: runs for new pull_requests and to any new commit within a pull_request. `pull_request` as trigger mains that the test and build runs after merge
- **build-and-deploy**: the ssh-key is required to have write access to openscd.github.io. We deploy on the mater branch


.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and Deploy
+on:
+  push:
+    branches:
+      - main
```
dlabordus, line 5: For publishing artifacts we using the release mechanisme from Github.
on:
  release:
    types: [released]

See https://github.com/com-pas/compas-scl-data-service/blob/develop/.github/workflows/release-project.yml.
Github will create a tag and we extract that and set the version. For this project through Maven, but I think NodeJS/NPM also has something like that.

.github/workflows/test.yml
```diff
@@ -0,0 +1,22 @@
+name: Test
```
dlabordus, line 1: Is there a big difference between only test and doing the build?
In our fork we run the Test and Build on every push. This is also enough for the pull requests.

dlabordus (approved):

.github/workflows/build-and-deploy.yml
JakobVogelsang: Thanks for hint. We will have a closer look at that. We have a half automated release process at the moment and we for sure can automate more.

.github/workflows/test.yml
JakobVogelsang: Not really. Was just an idea to save some time.

PR #354, opened 2021-11-04
refactor: use custom element constructors instead of tag names

### LINKED ISSUE #350: Scope elements tag names in open-scd
Open-wc recommends to scope element tag names to avoid naming collision, in addition to may other advantages coming with this approach. 

We want to introduce this concept in open-scd as well. However, we have decided to use `careHtml` to do this. There are several steps we need to do one after another:
1. Export `html = wrapHtml(litHtml)` from foundation.js 
2. Remove import of `html` from `lit-element`, `lit-html` and `@open-wc/testing` and replace with `html` from foundation
3. Replace custom element usage in our render function that look like so: `html`<wizard-selected...` to something that look liek so `html`<§{WizardSelect}...`. This has to be done with your components `WizardTextField`, `WizardSelect`, `WizardDialog`, but not `open-scd`
4. Remove `@customElement` decorator 
5. Remove all import statement like so `import ".../wizard-textfield.js"`
6. Update tests and snapshots

mandurisaimounika:
Closes #350


ca-d (approved):
We usually try to separate internal (end in .js) from external (no .js at the end) dependencies' import statements by an empty line and try to put foundation.js first among internal dependencies.

LGTM otherwise

JakobVogelsang (approved):
LGTM

JakobVogelsang:
@ca-d I would suggest doing the rearrangement of all import statements at the end of this refactor. There are just too many inconsistencies at the moment and makes it harder to review if done right now.

ca-d:
> @ca-d I would suggest doing the rearrangement of all import statements at the end of this refactor.

Sounds good.

ca-d:
Sadly, we only noticed with the second to last component migrated (tab) that there are hard internal dependencies between mwc-tab and mwc-tab-bar which forbid us from registering our own tag names for mwc elements at the moment.

We may think about raising an issue with the material-web team about future support for [Scoped Custom Element Registry](https://github.com/WICG/webcomponents/issues/895) .

Thank you for your painstaking work @mandurisaimounika , your changelog was invaluable for us in preparing https://github.com/openscd/open-scd/pull/395 .

PR #357, opened 2021-11-05
fix(editor-container): some UI related issues

### LINKED ISSUE #355: Substation editor shows some inconsitencies
With the newest version of OpenSCD the Substation editor does have some smaller UI issues.

1. Header of elements is highlighted only on clicking on the header and not on the container itself. This is only the case in Chrome, Edge, Opera not in Firefox and must have something to do with the slot mechanism we are using as this was not an issue before. 

2. There is a strange  black outline with the substation-editor. Also, just in Chrome ... and also on clicking within a child container not on a header. Besides, only visible on debugged versions, not on openscd.github.io.

3. The add element menu bar opens in the right upper corner for substations but on the left upper corner for other elements in the substation

4. There are tiny spaced between outline and header with the voltage level element. This is not the case for older versions of OpenSCD

JakobVogelsang:
Closes #355

### DIFF
--- a/__snapshots__/bay-editor.md
+++ b/__snapshots__/bay-editor.md
@@ -39,10 +39,7 @@
     <mwc-icon-button icon="delete">
     </mwc-icon-button>
   </abbr>
-  <div
-    id="ceContainer"
-    slot="container"
-  >
+  <div id="ceContainer">
     <conducting-equipment-editor>
     </conducting-equipment-editor>
     <conducting-equipment-editor>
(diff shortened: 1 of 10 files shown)


JakobVogelsang:
One small remark to this PR. The first commit does solve the collapsed margins issue, as we did it before the use of the editor-container. However, action icons mwc-fab might not be completely visible.

ca-d (approved):
LGTM, thanks 🥳

PR #366, opened 2021-11-09
First piece of SLD Editor

Flurb:


src/editors/SLD/bay-sld.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Get all the Connectivity Nodes of this particular Bay.
+   */
+  get connectivityNodeElements(): SldElement[] {
+    const sldelements: SldElement[] = [];
+
+    Array.from(getChildElementsByTagName(this.element, 'ConnectivityNode'))
+      .forEach(child => {
```
JakobVogelsang, line 103: this need  another filter like so:
```
.filter(cNode => cNode.getAttribute('name')!=='grounded')
```
this is a predefined ground terminal that we don't need to draw is it is indicated with the icon of the equipment already

JakobVogelsang:
There is a newer version #376

PR #368, opened 2021-11-09
refactor: use class as tag for wizard-textfield custom element

mandurisaimounika:


JakobVogelsang:
@ca-d : we query in tests and in other files with the tag name that is defined by carehtml. This is good enough? Or must we make sure to never query for tag's that are created by care html?

ca-d:
> @ca-d : must we make sure to never query for tag's that are created by care html?

I don't think we need to do that as long as carehtml's name generation stays stable. Once that starts to change we should look at other ways.

JakobVogelsang (approved):
LGTM

PR #374, opened 2021-11-12
feat(menu/UpdateDescritionABB): update ExtRef with internal signal description

JakobVogelsang:


ca-d:
Does this pull request have an associated issue? If not, should we add Assignee (JakVog), Labels, and Project directly here? Do we even have an appropriate project or don't we want to track this work on our boards?

src/foundation.ts
```diff
@@ -682,14 +682,14 @@ function extRefIdentity(e: Element): string | number {
   ].map(name => e.getAttribute(name));
 
   const cbPath = srcCBName
-    ? `${serviceType}:${srcCBName} ${srcLDInst ?? ''}/${
-        srcPrefix ?? ''
-      } ${srcLNClass} ${srcLNInst ?? ''}`
+    ? `${serviceType}:${srcCBName} ${srcLDInst ?? ''}/${srcPrefix ?? ''} ${
+        srcLNClass ?? ''
```
ca-d, line 686: Should these two fixes (this and line 692) maybe go into their own pull request with their own associated issue and regression test?

src/translations/de.ts
```diff
@@ -48,6 +48,7 @@ export const de: Translations = {
     title: 'Menü',
     viewLog: 'Protokoll anzeigen',
     viewDiag: 'Daignose anzeigen',
+    updatedesc: 'Signalbeschreibungen zu {{manufacturer}} IEDs hinzugefügt',
```
ca-d, line 51: I think `menu` should probably only contain built-ins, plugin translations should go into their own namespaces. (Also, this translation is for a log entry, not a menu entry.)

In my opinion it would be worth it to replicate this translation effort vendor dependently for each plugin.

src/translations/en.ts
```diff
@@ -46,6 +46,7 @@ export const en = {
     title: 'Menu',
     viewLog: 'View log',
     viewDiag: 'View diagnostics',
+    updatedesc: 'Add signal descriptions to {{manufacturer}} IEDs',
```
ca-d, line 49: This says "add" (like a menu entry), the German translation says "added" (like a log entry, which is how this is actually being used as far as I can see).

ca-d (changes requested):
I request a change only to the new translation introduced, the pull request refactor is more of a suggestion.

src/menu/UpdateDescriptionABB.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/** Plug-in that enriched ExtRefs desc attribute based on intAddr attribute (ABB)*/
+export default class UpdateDescriptionAbb extends LitElement {
+  /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
+  doc!: XMLDocument;
+
+  /** Entry point for this plug-in */
+  async run(): Promise<void> {
+    const items = Array.from(this.doc.querySelectorAll('IED')).flatMap(ied =>
```
ca-d, line 112: Should this say `':root > IED'` just to be safe?

src/translations/de.ts
JakobVogelsang: Agreed!

src/translations/en.ts
JakobVogelsang: Agreed

src/menu/UpdateDescriptionABB.ts
JakobVogelsang: Agreed!

src/foundation.ts
JakobVogelsang: Yes we should. I would love to do that when refactoring identity and selector out of the foundation and with it write more and more specific tests especially for the more complex element like `ExtRef`. What do you think about this?

ca-d: Sounds good.

ca-d (approved):
LGTM

PR #376, opened 2021-11-14
First version Single Line Diagram editor

Flurb:


public/js/plugins.js
```diff
@@ -6,6 +6,13 @@ export const officialPlugins = [
     default: true,
     kind: 'editor',
   },
+  {
+    name: 'Single Line Diagram',
+    src: '/src/editors/SingleLineDiagram.js',
+    icon: 'edit',
+    default: true,
```
dlabordus, line 13: Shouldn't this be 'false' to only show it when some explicitly selects it.

JakobVogelsang: This is true. So far it is an alpha version we should not make it visible by default

danyill (commented):
hiya! I've made a few brief comments and I realize this PR is WIP, I hope that's OK. If they are not of assistance please don't let me slow down this great work :wink:

src/editors/SingleLineDiagram.ts
```diff
@@ -0,0 +1,251 @@
...
+                        .forEach(eq => eq.appendChild(terminal))
+                });
+        });
+    }
+
+    /**
+     * Draw all the Substation elements.
+     */
+    drawSubstationElements() {
+        this.drawVoltageLevels();
```
danyill, line 200: Some substation elements are being drawn as entirely separate `svg` elements. Perfectly fine but could I suggest that there could be some benefit in using a single svg canvas/drawing and using `<g>` elements inside it. I think this is what is intended but possibly there's a namespace or parent/child appending issue in the `draw` routines.

In case it's of use, here's [a gist](https://gist.github.com/danyill/af80f8e053c26302e184df958293a6c3) and an scd file: [IEC station 1.scd.zip](https://github.com/openscd/open-scd/files/7548762/IEC.station.1.scd.zip)

public/js/plugins.js
Flurb: Resolved! :D

src/editors/SingleLineDiagram.ts
Flurb: Thanks for the feedback @danyill .

The idea is indeed to have everything drawn on 1 svg, and each element is a <g> element containing elements representing that element.

As far as I can see, I do see that Conducting Equipments and Connectivity Nodes are indeed svg's, which is not how we want it and which I implemented wrong. I will fix this :)

Thanks!

src/editors/SingleLineDiagram.ts
```diff
@@ -0,0 +1,251 @@
...
+        this.voltageLevels.forEach(voltageLevel => {
+            const voltageLevelElement = createVoltageLevelElement(voltageLevel);
+            this.svg.appendChild(voltageLevelElement);
+        })
+    }
+
+    drawBays(): void {
+        this.bays.forEach(bay => {
+            const bayElement = createBayElement(bay);
+            this.svg.querySelectorAll(`g[id=${getParentElementName(bay)}]`)
```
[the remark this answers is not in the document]
Flurb: Will fix this, good one!

src/editors/SingleLineDiagram.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * Draw all available Bus Bars of this SCL document.
+     */
+    drawBusBars(): void {
+        this.busBars.forEach(busBar => {
+            const busBarElement = createBusBarElement(busBar, this.biggestVoltageLevelXCoordinate);
+            
+            this.svg.querySelectorAll(`g[id=${getParentElementName(busBar)}]`)
```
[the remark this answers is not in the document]
Flurb: Again, will fix this, good one!

Flurb:
Hi @JakobVogelsang @ca-d , I think this is a good start of the SLD editor.
When do you guys think it's ready to (squash) merge?

JakobVogelsang:
@Flurb I am at the review. I, too, think this is a good start. We are close to finishing it.

danyill:
This is great. Thanks for this wonderful addition.

We're currently able to pan zoom into the top menu bar which seems undesirable and might be worth doing something about.

image

image

Flurb:
Yes that's a known bug and should be fixed! :)

Thanks for the feedback.
I'm sure some SCL files will render a bit weird, and it might not be perfect yet.

I think it's a matter of trying out a lot of SCL files, see what comes out and adjust it :) @danyill

danyill:
> Yes that's a known bug and should be fixed! :)

I was curious whether my modest CSS skills would allow me to do something. Perhaps this at the end of `SingleLineDiagram.ts`? I don't understand the overall architecture sufficiently to provide a very informed view so I won't do a PR...  :wink: 

```html
    render(): TemplateResult {
        // TODO: Width and Height should be a percentage, not fixed height/width.
        return html`<div id="panzoomContainer">
            <div id="panzoom">
                <svg xmlns="http://www.w3.org/2000/svg"
                    id="svg"
                    width="5000"
                    height="5000">
                </svg>
            </div>
        </div>`;
    }

    static styles = css`
    #panzoomContainer {
        overflow: hidden;
    }
    `;
}
```

> I think it's a matter of trying out a lot of SCL files, see how it renders and adjust (if needed) it if we see some weird renders :) @danyill

Yes, once this is merged I'll do some tests and perhaps some small PRs to deal with little things which I can manage as a learning exercise.

Flurb:
Looks good!
And I learned something about the transform update, looks a lot more clean! :D

Flurb:
By the way, I can't review it because I'm the PR author.

JakobVogelsang:
Sorry @Flurb I totally forgot to comment. Sorry for being so rude and directory pushing into your PR. I hope this is ok?

I liked the structure of the code and also the comments it was very clear to read. 

One thing still needs to be done and I didn't find a way to do that without bigger changes and without discussing with you first: the element PowerTransformer is not drawn, yet. They too can be connected with Terminals however as a grandchild  of PowerTransformer and child of TapChg.

JakobVogelsang:
> Looks good!
> And I learned something about the transform attribute, looks a lot more clean! :D
> 
> 

Btw I am not sure if you wanted to draw ConductingEquipment larger than 25px. If so you can use scale in transform in addition to translate that could look like:

setAttribute('transform', 'translate(x,y) scale(64/25)')

Flurb:
> > Looks good!
> > And I learned something about the transform attribute, looks a lot more clean! :D
> 
> Btw I am not sure if you wanted to draw ConductingEquipment larger than 25px. If so you can use scale in transform in addition to translate that could look like:
> 
> setAttribute('transform', 'translate(x,y) scale(64/25)')

I don't think we have specific requirements for how large the ConductingEquipments should be.
If larger looks better, it's fine for me 👍

ca-d (approved):
This looks really good, thanks for the code, comments and tests @Flurb, @danyill and @JakobVogelsang 🥳 !

Test coverage could be a bit more extensive, but we're not yet going to activate this plugin by default anyway, so we can add more tests as we go.

Flurb:
> This looks really good, thanks for the code and tests @Flurb and @JakobVogelsang 🥳 !
> 
> Test coverage could be a bit more extensive, but we're not yet going to activate this plugin by default anyway, so we can add more tests as we go.

#403 adds more tests

PR #380, opened 2021-11-19
refactor: scope mwc-select, mwc-checkbox

JakobVogelsang:


src/foundation.ts
```diff
@@ -3,16 +3,22 @@ import { directive, Part } from 'lit-html';
 import wrapHtml from 'carehtml';
 
 import { List } from '@material/mwc-list';
-import { Select } from '@material/mwc-select';
 import { TextField } from '@material/mwc-textfield';
 import AceEditor from 'ace-custom-element';
 
 import { WizardTextField } from './wizard-textfield.js';
 import { WizardSelect } from './wizard-select.js';
+import { Select } from './foundation/components.js';
```
JakobVogelsang, line 11: There are two imports of Select from ./foundation/components.js

src/foundation.ts
```diff
@@ -3,16 +3,22 @@ import { directive, Part } from 'lit-html';
...
 
 export const html = wrapHtml(litHtml);
 
-export { Dialog, Button } from './foundation/components.js';
+export {
+  Dialog,
+  Button,
+  Checkbox,
+  CircularProgressFourColor,
+  Select,
```
JakobVogelsang, line 20: Here another import of Select

src/foundation.ts
JakobVogelsang: Oh I got it the one is an import and the other one a export statement

PR #391, opened 2021-11-22
Bug fix for not connected routes in SLD Editor

Flurb:


src/editors/SingleLineDiagram.ts
```diff
@@ -15,14 +15,15 @@ import {
   getParentElementName,
   getAbsolutePositionWithCustomCoordinates,
   SVG_GRID_SIZE,
-  drawRoute,
+  drawRouteBetweenElements as drawRoute,
```
JakobVogelsang, line 18: Why not change both the names. Why keeping drawRoute as function name if you think the other one is better?

JakobVogelsang (commented):
Can be merged. The question a raised as a matter of taste

src/editors/SingleLineDiagram.ts
Flurb: Hi @JakobVogelsang , that's a weird change indeed.
I think it is something Visual Studio did, I will change it!

JakobVogelsang (approved):
LGTM thanks!

JakobVogelsang:
I will rebase and merge first thing tomorrow @Flurb. Thanks again for this nice editor

PR #396, opened 2021-11-24
feat(SingleLineDiagram): add PowerTransformer to the SLD

JakobVogelsang:
The PR contains some steps:

First, I exchanged the name in the `<g>` element with the identity string. The first is a global identifier, the second not. @Flurb this was a great idea from you to write them into the `<g>` as we can make use of the selector functions that - with the `tagName` and `identifier` - can give the element in the doc. This allows us to very, very simple add user interaction. E.g. we even could reuse all the wizards that we use in the Substation plug-in editor to edit create manipulate add LNode and all the beautiful things we do :)

Also I 

1.) I decoupled the connectivity node icon from the conducting equipment icons
2.) added a two winding power transformer icon
3.) realized that substation must be valid parent as PowerTransformer can be direct child of Substation: triggered by this did a refactor of the get...Coordinates/get...Position handling. In this PR I propose to call x and y in the SCL as coordinates and x and y in the SCL as position. Also I tried to use absolute - with parent position - and relative -without.
4.) Did another refactor to get bus bar length: Triggered as relative position is required here. With the getAbsolutePosition the function became very slim
5.) Finally added the PowerTransformer related function.

I think we can test most sld-drawing related function very clean with snapshot tests. I did this as a beginning to the createPowerTransformer...

My allegories for this big PR!

### DIFF
--- a/__snapshots__/Single Line Diagram drawing.md
+++ b/__snapshots__/Single Line Diagram drawing.md
@@ -1,0 +1,68 @@
+# `Single Line Diagram drawing`
+
+## `creates a group element for every given PowerTransformer element that`
+
+####   `looks like its latest snapshot`
+
+```html
+<g
+  id="AA1>TA1"
+  sxy:x="1"
+  sxy:y="9"
+  type="PowerTransformer"
+>
+  <line
+    stroke="currentColor"
+    stroke-linecap="round"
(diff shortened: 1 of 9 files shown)


Flurb (approved):
Looks good, some good refactoring skills! 👯

PR #400, opened 2021-11-29
refactor(editor-container): replace with more generic action-pane

ca-d:
The new `action-pane` component is the extraction of the pure user interface part of the old `editor-container` component, meant to replace the latter while being usable in a more generic way for other purposes than editing a single SCL element.


JakobVogelsang (approved):
LGTM

PR #401, opened 2021-11-29
test: replace karma with web-test-runner

### LINKED ISSUE #351: Updating to `web-test-runner` from `karma-test-runner`
We had two choice to prepare for the move to test runner. We either have to fix import statements in the test or scope elements tag name as proposed #350.

We have decided to do the latter as this has other advantages for us in the future and is the recommended approach of open-wc. Thas means moving can only be done after #350 

However, after trying with mwc components we realized this cannot be done at the moment. So we directly update import statements.

ca-d:
Closes #351 .


src/editors/communication/p-types.ts
```diff
@@ -129,7 +129,7 @@ export const typePattern: Partial<Record<string, string>> = {
   'OSI-NSAP': typeBase.OSI,
   'SNTP-Port': typeBase.port,
   'MMS-Port': typeBase.port,
-  DNSName: '\\S*',
+  DNSName: '[^ ]*',
```
ca-d, line 132: This should be fine due to restricted text entry in `mwc-textbox`. We might need to adapt this if we change input methods.

ca-d (commented):
LGTM

JakobVogelsang (approved):
LGTM

src/editors/communication/p-types.ts
JakobVogelsang: I add a FIXME as a remider.

PR #402, opened 2021-11-29
Added terminal and Connectivity Node information

Flurb:
Information is read-only.

Saving is disabled, because you can't just change the pathName of a Connectivity Node for example.
Because then, you have to change the terminals as well of connected Conducting Equipments for example.

So editing is disabled, just viewing.
However, with Pro mode you can edit of course.
But I assume people using Pro mode know what they are doing.

### DIFF
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -196,6 +196,29 @@ export const de: Translations = {
     },
     unknownType: 'Unbekannter Typ',
   },
+  connectivitynode: {
+    name: '????',
+    wizard: {
+      nameHelper: '????',
+      pathNameHelper: '????',
+      title: {
+        add: '????',
+        edit: '????',
+      },
+    },
+  },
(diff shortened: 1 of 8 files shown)


src/translations/de.ts
```diff
@@ -196,6 +196,29 @@ export const de: Translations = {
     },
     unknownType: 'Unbekannter Typ',
   },
+  connectivitynode: {
```
ca-d, line 199: I'll need to take a look at these German translations.

@Flurb Do you guys want to add a Dutch translation to the user interface at some point?

ca-d (changes requested):
LGTM apart from missing translations.

src/translations/de.ts
ca-d: I'll try to resolve this myself.

ca-d (approved):

PR #403, opened 2021-11-29
feat(SingleLineDiagram): add other ortho-connector

JakobVogelsang:


JakobVogelsang:
@Flurb : there would be all my suggestions. There is one test failing that I could not fix. I will ask @ca-d to have a look and convert to PR again.

Flurb:
The error is weird, because Panzoom does have an export default.
And the error says it doesn't have it.

dlabordus (approved):

PR #410, opened 2021-12-02
Added Edit Wizard to ConductingEquipment in SLD editor

Flurb:


JakobVogelsang:
@Flurb: it would be interesting to show and edit x and y coordinates here, as well as dir. I would suggest copying the conductingequipment.ts into a wizards folder within singlelinediagram folder and to extend the existing edit wizards by three input fields: x, y and dir.

Flurb:
> @Flurb: it would be interesting to show and edit x and y coordinates here, as well as dir. I would suggest copying the conductingequipment.ts into a wizards folder within singlelinediagram folder and to extend the existing edit wizards by three input fields: x, y and dir.

That could be a nice feature for the future, this PR is pre read only and that's what we wanted right now :)

JakobVogelsang (approved):
LGTM

PR #417, opened 2021-12-08
refactor(conducting-equipment-editor): use action-icon

JakobVogelsang:


src/zeroline/conducting-equipment-editor.ts
```diff
@@ -7,34 +7,32 @@ import {
...
+
+import { BayEditor } from './bay-editor.js';
 import { wizards } from '../wizards/wizard-library.js';
 
 /** [[`SubstationEditor`]] subeditor for a `ConductingEquipment` element. */
 @customElement('conducting-equipment-editor')
 export class ConductingEquipmentEditor extends LitElement {
-  @property({ type: Element })
+  /** SCL element ConductingEquipment */
+  @property({ attribute: false })
```
ca-d, line 22: What's the difference between this and `@state()`?

ca-d (approved):
LGTM, gets rid of a lot of unneeded code and adds useful tests, a great combination!

src/zeroline/conducting-equipment-editor.ts
JakobVogelsang: @state() is meant to be used with private and protected properties. It indeed sets the attribute to false per default, but I think is not meant to be referenced from outside which does not fit for `element`

```
Properties declared with @state shouldn't be referenced from outside the component. [reference](https://lit.dev/docs/v1/components/properties/)
```

PR #421, opened 2021-12-10
feat(wizards/reportcontrol): add edit wizard for ReportControl element

JakobVogelsang:


JakobVogelsang:
@danyill: I was wondering if you have time and are interested to do some reviews. In particular, I have the issues related with [control blocks](https://github.com/openscd/open-scd/projects/6) in mind.  @Flurb is helping out here already, but I think it would be beneficial for the project to have more reviewers.

src/wizards/reportcontrol.ts
```diff
@@ -0,0 +1,222 @@
...
+    const rptID = getValue(inputs.find(i => i.label === 'rptID')!)!;
+    const indexed = getValue(inputs.find(i => i.label === 'indexed')!);
+    const max = getValue(inputs.find(i => i.label === 'max Clients')!);
+    const bufTime = getValue(inputs.find(i => i.label === 'bufTime')!);
+    const intgPd = getValue(inputs.find(i => i.label === 'intgPd')!);
+
+    let reportControlAction: EditorAction | null;
+    let rptEnabledAction: EditorAction | null;
+    if (
+      name === element.getAttribute('name') &&
```
Flurb, line 62: I'm not a big fan of huge if's like this, are there fancy ways in Typescript to reduce the number of if's in this case? (Maybe working with an interface?)

src/wizards/reportcontrol.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const actions: EditorAction[] = [];
+    if (reportControlAction) actions.push(reportControlAction);
+    if (rptEnabledAction) actions.push(rptEnabledAction);
+    return actions;
+  };
+}
+
+function renderReportControlAttributes(
+  name: string | null,
```
Flurb, line 103: Same for this, I think it's good to have a maximum number of parameters for functions. More parameters means more complexity (in this case it's not that complex, but it's good to have a broad max number of parameters).

Flurb: The parameter value can be discussed, but max 4 is a fair starting point imo.

danyill:
> @danyill: I was wondering if you have time and are interested to do some reviews. In particular, I have the issues related with [control blocks](https://github.com/openscd/open-scd/projects/6) in mind like this one. @Flurb is helping out here already, but I think it would be beneficial for the project to have more reviewers.

I'd be happy to try and help although my skills are modest. I'll keep my 👀 open, feel free to ping me. Next year I hope to have more time available to put in this direction as well.

src/wizards/reportcontrol.ts
JakobVogelsang: Thanks for the hint. This is an important point. I have defined an object that allows to define the render options. What do you think about this solution?

Flurb (commented):
In this situation you keep the same number of parameters and instead you define an interface. I think it's the same 'problem', you keep 1 function that does a lot, which is more complicated, functions should do (in best case scenario of course) 1 thing.

Another option would be defining multiple render functions (splitting it up), what do you think of that?

Flurb (approved):

PR #436, opened 2021-12-15
refactor: add action-icon

ca-d:
This adds an `action-icon` user interface component to unify the way we display icons with associated user actions.


JakobVogelsang (approved):
LGTM

PR #437, opened 2021-12-15
feat(wizards/reportcontrol): add select wizard

### LINKED ISSUE #427: Show report control blocks
As a system specification engineer, I want to see if any and how many `ReportControl` element(s) are defined in the project and within an IED element.

**Requirements**
- allow filtering for `ReportControl` blocks for name attribute
- allow filtering for parents LDevice, IED and LN/LN0

JakobVogelsang:
Closes #427

### DIFF
--- a/src/wizards/reportcontrol.ts
+++ b/src/wizards/reportcontrol.ts
@@ -13,6 +13,8 @@ import {
   EditorAction,
   getReference,
   getValue,
+  identity,
+  isPublic,
   SimpleAction,
   Wizard,
   WizardActor,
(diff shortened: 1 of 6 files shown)


JakobVogelsang:
IsPublic checks if a element has anchestor element Private. This is theoretically possible and this is one method to do this. Unfortunately there are three ways we are dealing with this problem in our code base. IsPublic is one and the preferred on. The other one is still in Communication wizards where we defined selectors that defined a clear path using `>` in selectors instead of ` `. The other one ist to not deal with it at all :). In the future we will have an own function selector function that will that have build in.

So to answer your question ReportControl element `within` Private section shall be ignored.
Am 15. Dez. 2021, 23:02 +0100 schrieb Rob Tjalma ***@***.***>:

Flurb:
> IsPublic checks if a element has anchestor element Private. This is theoretically possible and this is one method to do this. Unfortunately there are three ways we are dealing with this problem in our code base. IsPublic is one and the preferred on. The other one is still in Communication wizards where we defined selectors that defined a clear path using `>` in selectors instead of ` `. The other one ist to not deal with it at all :). In the future we will have an own function selector function that will that have build in. So to answer your question ReportControl element `within` Private section shall be ignored. Am 15. Dez. 2021, 23:02 +0100 schrieb Rob Tjalma ***@***.***>:
> […](#)
> @Flurb commented on this pull request. In src/wizards/reportcontrol.ts: > @@ -217,3 +219,28 @@ export function editReportControlWizard(element: Element): Wizard { }, ]; } + +export function selectReportControlWizard(element: Element): Wizard { + const reportControls = Array.from( + element.querySelectorAll('ReportControl') + ).filter(isPublic); For my information: ReportControl elements with a private element within should totally be ignored? — You are receiving this because you were assigned. Reply to this email directly, view it on GitHub, or unsubscribe.

Thanks for the answer! For the rest: Looks good!

PR #446, opened 2021-12-17
fix(wizards/fcda): remove incorrect iedName from FCDA

### LINKED ISSUE #445: Create FCDA is falsely writing iedName attribute to FCDA element
The element `FCDA` must not have an attribute `iedName`.

JakobVogelsang:
When creating a `FCDA` element, we add an `ìedName` attribute to this element which is not defined within `FCDA` and therefor is schema invalid.

Closes #445


ca-d (approved):
LGTM, but why did we do this in the first place?

JakobVogelsang:
This was my mistake. I just did not look it up, as was sure this is a defined attribute :(.

Issue #375, opened 2021-11-12
Increase padding in communication screen to see buttons and IED names


danyill:
In the communication screen at the moment there is insufficient padding to see the IED name or the interface buttons when there is a reasonable number of IEDs:

See screenshot (Firefox) from the current `main` branch:

![image](https://user-images.githubusercontent.com/674783/141531623-3a9df2f1-16fc-4829-a44b-f38c0f17bb22.png)

danyill:
Adopting a similar layout as the Communication tab might be one good option:

![image](https://user-images.githubusercontent.com/674783/141531900-b1b4d359-70a0-4486-ad64-26fea2b6a817.png)

JakobVogelsang:
Hi thanks for the comment, 

You are right and it is unfortunate this is the case at the moment. Just so that you can track it: We are already on it. The branch `action-icon` is a common web-component that we will be used for everything where actions are hidden behind an icon. With this we want to make sure that we have a common UI. This I think will solve the issue.

JakobVogelsang:
This is now in main. We have added some other styling and are very curious about your feedback.

danyill:
I'm really impressed, good job on the new styling :+1: it feels clean and fresh 

It's probably not great commenting on a closed issue and I do not wish to nitpick (!) but I would prefer enough padding to see IED names with up to say, 12 characters as a minimum. For our in-house naming we typically have:

* 3 characters for a substation
* 4 characters for a bay
* 2-3 characters to identify device function

I think this is fairly typical.

Here's an example in the Substation editor:

![image](https://user-images.githubusercontent.com/674783/146666642-a79ff568-6514-4d4c-b25a-5d59e391b985.png)

The Communication editor really gets this right:

![image](https://user-images.githubusercontent.com/674783/146666669-0cf982e1-5fa4-4ec4-9fb5-c034b2f4a64d.png)

PR #447, opened 2021-12-19
style(zeroline/ied-editor): allow longer IED name

danyill:
I wanted to allow a longer IED name in the substation editor (about 12 characters). In my experience, IED names can be long, and not all substations have primary equipment in the SCD file and the substation editor is an important entry point in open-scd.

See comments in #375.

Result:

image

I'm not very up with modern CSS frameworks and so I may be breaking a host of rules here in which case this PR can be closed without me being very concerned :wink:

### DIFF
--- a/src/action-icon.ts
+++ b/src/action-icon.ts
@@ -157,7 +157,7 @@ export class ActionIcon extends LitElement {
       margin: 0px;
       text-align: center;
       align-self: center;
-      max-width: 64px;
+      max-width: 120px;
       direction: rtl;
     }
 
--- a/src/action-icon.ts
+++ b/src/action-icon.ts
@@ -157,7 +157,7 @@ export class ActionIcon extends LitElement {
       margin: 0px;
       text-align: center;
       align-self: center;
-      max-width: 120px;
+      max-width: 100%;
       direction: rtl;
     }
 
--- a/src/zeroline-pane.ts
+++ b/src/zeroline-pane.ts
@@ -175,7 +175,7 @@ export class ZerolinePane extends LitElement {
       grid-gap: 12px;
       padding: 8px 12px 16px;
       box-sizing: border-box;
-      grid-template-columns: repeat(auto-fit, minmax(64px, auto));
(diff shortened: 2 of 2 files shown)


JakobVogelsang:
@danyill Thank you for the PR. This is why having users be part of the development is so important. You are the one using the software on a daily basis. I don't see anything that could break the style in other places. Still, as we don't have pixel tests, I would take time and manually test some places. So please give me some time to review this one.

JakobVogelsang:
Hmmm this will lead to overlapping in other element like.
grafik

src/action-icon.ts
```diff
@@ -157,7 +157,7 @@ export class ActionIcon extends LitElement {
       margin: 0px;
       text-align: center;
       align-self: center;
-      max-width: 64px;
+      max-width: 120px;
```
JakobVogelsang, line 160: I think I got the solution to this one: change the max-width rule in `action-icon` to 100%. Then the footer will adapt to the grid definitions you did in `ied-editor` and leave the rest untouched.

src/zeroline-pane.ts
```diff
@@ -175,7 +175,7 @@ export class ZerolinePane extends LitElement {
       grid-gap: 12px;
       padding: 8px 12px 16px;
       box-sizing: border-box;
-      grid-template-columns: repeat(auto-fit, minmax(64px, auto));
+      grid-template-columns: repeat(auto-fit, minmax(128px, auto));
```
JakobVogelsang, line 178: May the change in action-icon is already enough? Or do you think IED should not be so close together in any case?

JakobVogelsang (changes requested):

src/action-icon.ts
danyill: :+1: Quite right.

src/zeroline-pane.ts
danyill: I think they should be spaced further apart, similar to the Communication editor.
Also, this change is required even if `footer` is set to `100%`, otherwise the text still overflows the column width:

image

JakobVogelsang (approved):
LGTM

PR #449, opened 2021-12-20
feat(editors/SingleLineDiagram): allow selecting the Substation element

### LINKED ISSUE #416: Select substation in SLD editor
As a engineer,
I would like to select a substation to show in the Single line diagram
So I other substations as well (not only the first one)

Background
An SCD file can contain multiple substations.

dlabordus:
- Added Substation Selector in the SLD Editor. This was needed, because with multiple substations the elements are overlapping, because all substations start drawing from position 0,0.
- Fixed drawing Connectivity Nodes when there is no SubstationName/VoltgaeLevelName/BayName attribute on the Terminal. These attributes are optional in the XML and should only be check if available.
- Refactoring drawing the SVG for a Substation. 

Pull Request resolves #416

### DIFF
--- a/src/editors/SingleLineDiagram.ts
+++ b/src/editors/SingleLineDiagram.ts
@@ -4,12 +4,18 @@ import {
   LitElement,
   property,
   query,
+  state,
   TemplateResult,
 } from 'lit-element';
 
 import panzoom from 'panzoom';
 
-import { identity, newWizardEvent, SCLTag } from '../foundation.js';
+import {
+  compareNames,
+  identity,
+  newWizardEvent,
+  SCLTag
+} from '../foundation.js';
 import {
   getAbsolutePosition,
   createTerminalElement,
@@ -31,8 +37,17 @@ import {
   isBusBar,
   getConnectedTerminals,
   getPathNameAttribute,
+  getNameAttribute,
+  getDescriptionAttribute,
 } from './singlelinediagram/foundation.js';
+import {isSCLNamespace} from "../schemas.js";
 import { wizards } from '../wizards/wizard-library.js';
+import {SingleSelectedEvent} from "@material/mwc-list/mwc-list-foundation";
+import {translate} from "lit-translate";
+
+import '@material/mwc-list/mwc-list-item';
+import '@material/mwc-select';
+import '@material/mwc-textfield';
 
 /**
  * Main class plugin for Single Line Diagram editor.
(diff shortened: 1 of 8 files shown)


src/schemas.ts
```diff
@@ -15,6 +15,16 @@ export function newEmptySCD(
   return new DOMParser().parseFromString(markup, 'application/xml');
 }
 
+export const SCL_NAMESPACE= "http://www.iec.ch/61850/2003/SCL";
+
+/**
+ * Check if the namespace of tjhe passed element is the standard SCL Namespace.
```
JakobVogelsang, line 21: typo

src/editors/SingleLineDiagram.ts
```diff
@@ -42,206 +57,288 @@ export default class SingleLineDiagramPlugin extends LitElement {
...
+  selectedSubstation: Element | undefined;
+
+  // Container for giving the panzoom to.
+  @query('#panzoom') panzoomContainer!: HTMLElement;
+  // The main canvas to draw everything on.
+  @query('#svg') svg!: SVGElement;
+
+  private get substations() : Element[] {
+    return Array.from(this.doc.querySelectorAll(':root > Substation'))
+      .filter(isSCLNamespace)
```
JakobVogelsang, line 70: I don't get why you filter for `isSCLNamespace`. What do I miss?

src/editors/SingleLineDiagram.ts
```diff
@@ -42,206 +57,288 @@ export default class SingleLineDiagramPlugin extends LitElement {
...
+      .filter(element => element.querySelector(`Terminal[connectivityNode="${pathName}"]`));
+  }
 
-      this.addElementToGroup(bayElement, identity(bay.parentElement));
-    });
+  /**
+   * Draw all equipment and connections of the selected Substation.
+   */
+  private drawSubstation(): void {
+    const substationGroup = createVoltageLevelElement(this.selectedSubstation!);
```
JakobVogelsang, line 143: The naming in this line is missleading :)

src/editors/SingleLineDiagram.ts
```diff
@@ -310,21 +413,96 @@ export default class SingleLineDiagramPlugin extends LitElement {
...
+      color: var(--mdc-theme-on-surface);
+      font-family: 'Roboto', sans-serif;
+      font-weight: 300;
+      overflow: hidden;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+      margin: 0px;
+      line-height: 48px;
+      padding-left: 0.3em;
+      transition: background-color 150ms linear;
```
JakobVogelsang, line 495: This rule is not used as there is no transition.

src/editors/SingleLineDiagram.ts
```diff
@@ -310,21 +413,96 @@ export default class SingleLineDiagramPlugin extends LitElement {
...
+    }
+
+    #substationSelector {
```
JakobVogelsang, line 498: I would prefer to have a margin around the selector at least top and left I woudl say

JakobVogelsang (changes requested):
Some very minor changes would be nice. Other than that LGTM :)

src/editors/SingleLineDiagram.ts
dlabordus: For this query it can be removed, because at root level there is only SCL Substation possible.

But the reason I'm doing it at other level is because RTE is talking about adding a Compas Extension for Bay/VoltageLevel/.... This means we will get multiple Namespaces using the same Element names. QuerySelector doesn't look at the namespace, so it would be cleaner to only pickup the SCL elements and not all kind of privates that are there with the same name.

src/editors/SingleLineDiagram.ts
dlabordus: I saw it also when working at the editors (other issue). Will fix this.

src/editors/SingleLineDiagram.ts
dlabordus: This is currently copied from the Substation Editor. What example is better?

src/schemas.ts
dlabordus: Fixed

src/editors/SingleLineDiagram.ts
dlabordus: I just removed the transition. Remark: all editor still contain this style :-)

JakobVogelsang: Yes I know :). This is not completely clean there too. It is only the one line.

JakobVogelsang (approved):
LGTM

PR #452, opened 2021-12-21
feat(wizards/reportcontrol): test new UI variant

JakobVogelsang:
@ca-d and I were discussing the user-friendliness with the `GSEControl ` wizards. Here you have to do 3 of clicks to get for example to the wizards to add some data. We were thinking to maybe change the access to the wizards and reduce the number of click. This is an experiment that does this directly in the filtered list with `ReportControl` elements. 

I would love you to test and hope for some feedback. :) Maybe you can have a look as well @danyill 

In order to test, compare the filtered lists of Report `R` and GOOSE `G`.

### DIFF
--- a/src/wizard-selector.ts
+++ b/src/wizard-selector.ts
@@ -1,0 +1,211 @@
+import {
+  customElement,
+  css,
+  queryAll,
+  LitElement,
+  property,
+  internalProperty,
+  TemplateResult,
+  html,
+  query,
+  state,
+} from 'lit-element';
+import { get, translate } from 'lit-translate';
+
+import '@material/mwc-button';
+import '@material/mwc-dialog';
+import '@material/mwc-icon-button-toggle';
+import { Dialog } from '@material/mwc-dialog';
+import { List } from '@material/mwc-list';
+
+import 'ace-custom-element';
+import {
+  newActionEvent,
+  Wizard,
+  WizardInput,
+  WizardPage,
+  newWizardEvent,
+  WizardActor,
+  wizardInputSelector,
+  isWizard,
+  checkValidity,
+  reportValidity,
+  Delete,
+  Create,
+  identity,
+  SCLTag,
+  selector,
+} from './foundation.js';
+import { FilteredList } from './filtered-list.js';
+import { wizards } from './wizards/wizard-library.js';
+import { ListItem } from '@material/mwc-list/mwc-list-item';
+
+/** A wizard style dialog consisting of several pages committing some
+ * [[`EditorAction`]] on completion and aborting on dismissal. */
+@customElement('wizard-selector')
+export class WizardSelector extends LitElement {
+  @property({ type: Object })
+  scope!: Element | XMLDocument;
+
+  @property({ type: String })
+  targetTag!: string;
+
+  @property({ type: Array })
+  childTags: string[] = [];
+
+  @state()
+  get targets(): Element[] {
+    return Array.from(this.scope.querySelectorAll(this.targetTag));
+  }
+
+  @state()
+  get element(): Element | null {
+    if (!this.list) return null;
+    const identity = (<ListItem>this?.list?.selected).value ?? '';
+    return this.scope.querySelector(selector(<SCLTag>this.targetTag, identity));
+  }
+
+  @query('filtered-list') list!: FilteredList;
+  /** Commits `action` if all inputs are valid, reports validity otherwise. */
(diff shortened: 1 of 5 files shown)


Sander3003:
We will look at this as well.

JakobVogelsang:
@Sander3003 I was hoping so :)

Sander3003:
We prefer the the Select ReportControl option. This looks the most usefull.

danyill:
Thank you for inviting feedback :smile:

UI is really challenging and I have very limited experience. It's easy to imagine different alternative solutions all of which may be viable and may take considerable time to implement and I'm also quite new here... :flushed:

(BTW I do really like the search bar implementation which does a nice "smart select" on the IED name/report name or other data fields which is very friendly.)

I see this PR as giving an advantage in providing all the edit options adjacent to the report.

However (and at the risk of being off-topic) I am a bit cautious about the extensive use of dialog UI elements and nesting/chaining of these:

* Each dialog use causes a context switch for users which can generate confusion and is a cognitive burden. In general modals/popups should have quite a limited scope according to the Material Design guidelines because [they are "interruptive"](https://material.io/components/dialogs#usage).
  * They are interruptive because they prevent any other activity (looking at other IEDs, comparing different devices for instance and carrying out other tasks)
  * They interrupt a logical display because they just "sit on top of it"
  * They should be used for decision making which is not how we are using them here. We are nesting them and opening up dialogs on top of dialogs and it is our "normal" approach to configuration (although this may change with IED editor or SLD editor)
 
* For instance here we are providing an extensive set of options and then as we go into each detailed screen we are using a different dialog (or sometimes expanding the dialog interactively e.g. when adding an FCDA by resizing it for example). The dialogs are not "connected" in the UI and a cancel option takes you back "to the start" (should it take you back one layer or are you asking to go back to the "main screen"?).

As a thought experiment: an alternative approach could be to have a pane or report window as a single screen with a side sheet/draw for the report attributes. Clicking on each report might allow direct editing and viewing of the dataset (or are datasets a separate concern that should have a different window given they are common with other functions e.g. GOOSE?).

I wonder if you would be interested in using something like [Figma](https://www.figma.com) and the [Material System](https://setproduct.gumroad.com/l/materialsystem) for prototyping UI concerns (faster turnaround on ideas and testing of interfaces and less coding re-work).

JakobVogelsang:
@danyill: Wizards being too interruptive to be used within each other is a very good point. I have added another prototype to show an example that works similar to the finder-list (Add data to data set)

@Sander3003 What do you think about this one?

danyill:
I think it looks great.

* This kind of refactor helps answer a question like the following: "If I edit in the first dialog and then click to enter the second dialog, have I made a commitment to save the state?".
* Navigation is more intuitive and more powerful within a single screen and the user retains immediately apparent control of the context.
* It is now very easy to compare options set on particular reports.

So :100: from me! 

Incidentally I notice that when I edit a parameter and select another report the change transfers across to the new report which may not be as intended. Perhaps we need to show a little icon indicating "saved" or "changed" after a parameter is modified in this interface?

A little note, when we feel we want to nest dialogs I suggest we consider having a "Back" button on the left-hand side so the user is not forced to go back to the start, perhaps something like this:

image

randy993:
We had a discussion last week about this and as Sander said we preferred the RC option. However an engineer on our side mentioned it would be very usefull when all the editable parameters are instantly shown when clicking on a RC/GOOSE. I have added a draft as a screenshot for an example.
design1 (003)

I think we should look into showing as much parameters as possible upon selecting a 'connection'. We can maybe even use more columns, since we have enough room on the screen (assuming responsiveness is not a big issue?).

JakobVogelsang:
@randy993 Just so that we are on the same page. What you proposed does sound like my proposal in the last commit. Is this correct or did I miss your point?

randy993:
@JakobVogelsang I see your latest commit is after we had our discussion so I have to check your latest version. Will do that today.

JakobVogelsang:
We had a discussion on this topic today. We decided to go with the "old" handling in the wizards for now, with some improvements like a button to navigate back and forth between the different wizards. This decision has been made for several reasons:

1. Faster development for this critically needed features
2. We were not sure, if configuring in wizard is the best UI at the end for this feature or maybe another editor is even better. So we did not want to make too drastic changes just to later change it again. 
3. Wizards are closely tied together with the wizard-library (UNFORTUNATELY) We need to change that anyway, but this is something we want to do in a later stage.

PR #454, opened 2021-12-22
feat(editors/ied): add DO elements to IED editor

Flurb:


src/editors/ied/do-container.ts
```diff
@@ -0,0 +1,87 @@
...
+      return html`${name}${desc ? html` &mdash; ${desc}` : nothing}`;
+    }
+  }
+
+  /**
+   * Get the DOType of this DO.
+   * @param doType - The type of this DO.
+   * @returns The DOType section for this DO.
+   */
+   private getDOType(): Element | null {
```
ca-d, line 45: Do you see something speaking against using js getters to access these as properties? It may fit better with the rest of the codebase.

ca-d (approved):
LGTM

src/editors/ied/access-point-container.ts
```diff
@@ -28,7 +27,7 @@ export class AccessPointContainer extends LitElement {
 
   render(): TemplateResult {
     return html`<action-pane .label="${this.header()}">
-    ${Array.from(this.element.querySelectorAll(IEDSelector.Server)).map(
+    ${Array.from(this.element.querySelectorAll(':scope > Server')).map(
```
JakobVogelsang, line 30: Nice, I did not know `:scope` :)

src/editors/ied/do-container.ts
JakobVogelsang: I also would suggest that. You can also combine getNestedDOElement and getDOType to something like

```
private getNestedDOElements(): Element[] {
    const id = this.element.getAttribute('type');

    const doType = this.element
      .closest('DataTypeTemplates')
      ?.querySelector(`DOType[id="${id}"]`);

    return Array.from(doType?.querySelectorAll('SDO') ?? []);
  }
```

src/editors/ied/do-container.ts
```diff
@@ -0,0 +1,87 @@
...
+    const doType = this.element.getAttribute('type') ?? undefined;
+    return this.element.closest('SCL')!.querySelector(`:root > DataTypeTemplates > DOType[id="${doType}"]`);
```
JakobVogelsang, line 47: closest to DataTypeTemplates is enough. This improves performance of the querySelector

src/editors/ied/do-container.ts
```diff
@@ -0,0 +1,87 @@
...
+      ${this.getNestedDOElements().map(dO =>
+        html`<do-container
+          .element=${dO}
+          .instanceElement=${this.getInstanceElement(dO)}>
+        </do-container>`)}
+    </action-pane>
+    `;
+  }
+
+  static styles = css``;
```
JakobVogelsang, line 86: styles are not used here. So long not this line superfluous.

src/editors/ied/ln-container.ts
```diff
@@ -27,8 +28,34 @@ export class LNContainer extends LitElement {
...
+   */
+  private getInstanceElement(dO: Element): Element | null {
+    const doName = getNameAttribute(dO);
+    return this.element.querySelector(`:scope > DOI[name="${doName}"]`)
+  }
+
   render(): TemplateResult {
     return html`<action-pane .label="${this.header()}">
+      ${Array.from(this.getLNodeType() ? this.getLNodeType()!.querySelectorAll(':scope > DO') : [])
+        .sort((a,b) => compareNames(a,b))
```
JakobVogelsang, line 53: For performance reason, I would leave out the sorting. The biggest IED I have was loading 12s :(

src/editors/ied/ln-container.ts
```diff
@@ -27,8 +28,34 @@ export class LNContainer extends LitElement {
...
+   * @returns The optional DOI object.
+   */
+  private getInstanceElement(dO: Element): Element | null {
+    const doName = getNameAttribute(dO);
+    return this.element.querySelector(`:scope > DOI[name="${doName}"]`)
+  }
+
   render(): TemplateResult {
     return html`<action-pane .label="${this.header()}">
+      ${Array.from(this.getLNodeType() ? this.getLNodeType()!.querySelectorAll(':scope > DO') : [])
```
JakobVogelsang, line 52: Why not getting DOs directly in the method getLNodeType. Of course the name does not fit :)

src/editors/ied/ln-container.ts
```diff
@@ -27,8 +28,34 @@ export class LNContainer extends LitElement {
             ${inst ? html` &mdash; ${inst}` : nothing}`;
   }
 
+  /**
+   * Get the LNodeType of this LN(0) section.
+   * @returns The LNodeType section, or null if not found.
+   */
+  private getLNodeType(): Element | null {
+    const lnType = this.element.getAttribute('lnType') ?? undefined;
+    return this.element.closest('SCL')!.querySelector(`:root > DataTypeTemplates > LNodeType[id="${lnType}"]`);
```
JakobVogelsang, line 37: closest to DataTypeTemplates is enough. Improves performance of the querySelector

src/editors/ied/do-container.ts
JakobVogelsang: Just saw this is a wrong comment

src/editors/ied/do-container.ts
Flurb: I agree with the removal of getDOType, because it's only usage is in getNestedDOElements :) So I will merge them!

I was wondering how you guys would implement it as a getter (the getNestedDOElements for example).
I think getters and setters should apply to properties, so you can control the access of a certain property. That's why I'm a bit confused about how it's being used right now in open-scd, for example in the voltage-level-editor.ts, I see a `get voltage()` without a property (the voltage method itself is handled as a property). In my opinion this is a wrong usage, but maybe my Typescript expertise is lacking here :)

Is this how we want to use it and how we want this getNestedDOElements as well? :) If so, maybe we can discuss it!

src/editors/ied/access-point-container.ts
Flurb: Always nice to learn something!

src/editors/ied/do-container.ts
Flurb: good one!

src/editors/ied/ln-container.ts
Flurb: I will leave it out for now.
Is sorted LNs a thing that we want in the future? @JakobVogelsang @ca-d

src/editors/ied/ln-container.ts
Flurb: I don't think this is correct, `this.element` here is a LN/LN0, `DataTypeTemplates` is not a mutual parent. The closest mutual parent here is `SCL`, but correct me if I'm wrong :)

JakobVogelsang (commented):
LGTM

Issue #456, opened 2021-12-23
Allow movement of Bays in SLD via sxy coordinates


danyill:
It's possibly a little premature but the good work in #455 of allowing coordinate movements in the SLD editor made me want to (at least prototype) the possibility of changing bay coordinates to shift bays around. It seemed like the same kind of thing.

Why? I have sometimes hand-crafted an `ssd` file to replicate bays and after this I wish to move them apart graphically but I find it tedious to modify `sxy` coordinates.

I'd like to do this by clicking on the bay and setting coordinates on it.

Then, within the bay the coordinates should be relative to the bay object according to the standard (they are nested and relative according to IEC 61850-6 Ed 2.1, Annex C).

> If coordinates are defined at different SCL tag hierarchy levels, then each level contains
> coordinates relative to the higher level. The absolute coordinate of a lower level is thus
> calculated by summing up all higher level coordinates, and the object coordinates themselves.
> If there are no coordinates defined at a higher level, then (x,y) = (0,0) is assumed.

Prototyping PR coming up momentarily. 

I have a limited amount of SSD files which work in the SLD editor at the moment (possibly due to unrelated reasons or poor constructions of the SSD or some kind of looping condition in the logic with open-scd's rendering, I have not investigated yet). My example for examining this is the attached:

[HRP.ssd.txt](https://github.com/openscd/open-scd/files/7771658/HRP.ssd.txt)

JakobVogelsang:
@danyill is this still relevant?

PR #457, opened 2021-12-23
feat(editors/SingleLineDiagram) Allow movement of bays

danyill:
Allow movement of bays, see #456 

I will be very happy to take a review and make corrections. I will be happy to be educated. Perhaps I could convince @dlabordus to take a look (hi! :wink:).

Note that I have used `SVGGraphicsElement` instead of `SVGElement` because many methods associated with graphical objects are not available in the TS types for `SVGElement` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement)). I have resisted the urge to replace this more widely than required to satisfy the compiler.

It seemed for the time being that the best way to access bay properties was on the bay text rather than the rectangle drawn for the bay as this overlaps other objects. This is not consistent with what is done elsewhere where we click on the graphic. Happy to take some ideas about a better approach.

The `ssd` file included on the issue might be helpful for tests.

https://user-images.githubusercontent.com/674783/147298144-cfeb23db-4afd-404e-892e-cbc1d0299c75.mp4

I'm not sure how to write a test but I'm happy to have a go.

### DIFF
--- a/src/editors/SingleLineDiagram.ts
+++ b/src/editors/SingleLineDiagram.ts
@@ -22,6 +22,7 @@ import {
   createBusBarElement,
   createVoltageLevelElement,
   createBayElement,
+  createBayOutlineElement,
   createConductingEquipmentElement,
   createConnectivityNodeElement,
   getBusBarLength,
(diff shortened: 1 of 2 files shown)


dlabordus:
Sorry Daniel, we already merged my pull request to the main branch, so the pull request is invalid now. :-(
Can you make a new pull request?

danyill:
> Can you make a new pull request? Thanks.

I'll rebase on main

Issue #267, opened 2021-07-22
Check for `SwTyp` when displaying conducting equipment


JakobVogelsang:
As a user of OpenSCD I want to see the type of disconnector I am working on. 

1.) If conducting equipment of type `DIS` is connected to a logical node `XSWI`
2.) You can get the `SwTyp` data object of `XSWI` which is an enumeration and mandatory
3.) And from that the `stVal` data attribute that CAN have a pre-defined `Val` which ultimately gives you more information about the type

danyill:
I'll take a look at this.

JakobVogelsang:
Is implemented!

PR #460, opened 2021-12-29
refactor(editors/substation/conductingequipment): add earth-switch symbol

### LINKED ISSUE #459: Show earth-switch symbol in SLD and Substation editor
Currently we only show the disconnector symbol for both the SLD and substation editor.

Probably because this hasn't been implemented and the IEC 61850-6 Ed 2.1 specified that an earth switch was (oddly) a disconnector with one terminal associated with a connectivity node whose name was defined as "grounded":

> There is one predefined connectivity node with the name grounded. This is used to model
> earth potential. Thus, an earthing switch is an isolator (equipment type DIS) that is connected
> on one side to the connectivity node grounded. It is up to the generating tool to decide if
> grounded is one single node for the whole substation, or a separate node at each place
> where connected, or something in between, for example per bay or voltage level, by
> generating appropriate pathNames.

Something like this:

```xml
<ConductingEquipment name="ES260" sxy:x="16" sxy:y="0" virtual="false" type="DIS">
   <LNode lnClass="XSWI" lnInst="3" prefix="ES260"/>
   <LNode lnClass="CSWI" lnInst="4" prefix="ES260"/>
   <LNode lnClass="CILO" lnInst="4" prefix="ES260"/>
   <Terminal bayName="B258268" cNodeName="L1" name="L1" substationName="HRP" voltageLevelName="V220" connectivityNode="HRP/V220/Bus_A2/L1"/>
   <Terminal bayName="B258268" cNodeName="grounded" name="L1" substationName="HRP" voltageLevelName="V220" connectivityNode="HRP/V220/B258268/grounded"/>
</ConductingEquipment>
<ConnectivityNode name="grounded" pathName="HRP/V220/B258268/grounded"/>
```
At present we don't have an icon in [icons.ts](https://github.com/openscd/open-scd/blob/main/src/icons.ts), the earth switch icon is a copy of the disconnector icon.

However the (repeated) logic used for identifying earth switches also doesn't seem quite right to me. In the edit wizard, conducting equipment it is showing the type as "Disconnector" rather than "Earth Switch" for my testing ssd file.

This appears to be because it is presumed that the first terminal is the one with the name grounded which may not always be the case. I think we should check all terminals.

https://github.com/openscd/open-scd/blob/c3f2ab7e173f72fbc896420fdbf9ffc74bbdff95/src/zeroline/foundation.ts#L226-L235

PR for this coming in a moment.

danyill:
Closes #459 by providing an earth switch icon and using this in the various editors and wizards.

The logic for ESR detection shouldn't be in multiple places but it may be like this for a very good reason so I have not modified it.

In the SLD:

image

In the Substation Editor:

image

I'll be happy to take a review and make modifications.

I've tried to keep the icon square while retaining the stroke width but happy to take some feedback as to whether this is visually acceptable. Another option to introduce spacing might be to make the switch portion smaller or the earth symbol smaller.

### DIFF
--- a/src/icons.ts
+++ b/src/icons.ts
@@ -455,66 +455,72 @@ export const voltageTransformerIcon = html`<svg
 </svg>`;
 
 export const earthSwitchIcon = html`<svg
-viewBox="0 0 25 25"
-version="1.1"
-xmlns="http://www.w3.org/2000/svg"
-xmlns:svg="http://www.w3.org/2000/svg">
-<line
-  x1="12.5"
-  y1="19.25"
-  x2="12.5"
-  y2="16.25"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
-<line
-  x1="12.5"
-  y1="1.25"
-  x2="12.5"
-  y2="6.25"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
-<line
-  x1="12.5"
-  y1="6.25"
-  x2="17"
-  y2="15.25"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
-<line
-  x1="13.5"
-  y1="16.25"
-  x2="11.5"
-  y2="16.25"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
-<line
-  x1="17"
-  y1="19.25"
-  x2="8"
-  y2="19.25"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />>
-<line
-  x1="15.5"
-  y1="21.45"
-  x2="9.5"
-  y2="21.45"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
-<line
-  x1="14.5"
-  y1="23.5"
-  x2="10.5"
-  y2="23.5"
-  stroke="currentColor"
-  stroke-width="1.5"
-  stroke-linecap="round" />
+  viewBox="0 0 25 25"
+  xmlns="http://www.w3.org/2000/svg"
+>
+  <line
+    x1="12.5"
+    x2="12.5"
+    y1="19.2"
+    y2="16.2"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="12.5"
+    x2="12.5"
+    y1="1.25"
+    y2="6.25"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="12.5"
+    x2="8"
+    y1="16.2"
+    y2="7.25"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="13.5"
+    x2="11.5"
+    y1="6.25"
+    y2="6.25"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="17"
+    x2="8"
+    y1="19.2"
+    y2="19.2"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="15.5"
+    x2="9.5"
+    y1="21.4"
+    y2="21.4"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
+  <line
+    x1="14.5"
+    x2="10.5"
+    y1="23.5"
+    y2="23.5"
+    stroke="currentColor"
+    stroke-linecap="round"
+    stroke-width="1.5"
+  />
 </svg>`;
 
 export const generalConductingEquipmentIcon = html`<svg
--- a/src/icons.ts
+++ b/src/icons.ts
@@ -456,7 +456,7 @@ export const voltageTransformerIcon = html`<svg
 
 export const earthSwitchIcon = html`<svg
   xmlns="http://www.w3.org/2000/svg"
-  viewBox="0 0 25 35"
+  viewBox="0 0 25 25"
 >  
   <line
     x1="12.5"
--- a/src/icons.ts
+++ b/src/icons.ts
@@ -468,15 +468,6 @@ export const earthSwitchIcon = html`<svg
     stroke-linecap="round"
   />
   <line
-    x1="11"
-    y1="23.5"
-    x2="14"
-    y2="23.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-  />
-  <line
     x1="12.5"
     y1="23.5"
     x2="12.5"
--- a/src/icons.ts
+++ b/src/icons.ts
@@ -455,73 +455,67 @@ export const voltageTransformerIcon = html`<svg
 </svg>`;
 
 export const earthSwitchIcon = html`<svg
-  xmlns="http://www.w3.org/2000/svg"
-  viewBox="0 0 25 25"
->  
-  <line
-    x1="12.5"
-    y1="5.5"
-    x2="12.5"
-    y2="8.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-  />
-  <line
-    x1="12.5"
-    y1="23.5"
-    x2="12.5"
-    y2="18.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-  />
-  <line
-    x1="12.5"
-    y1="18.5"
-    x2="8"
-    y2="9.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-  />
-  <line
-    x1="11.5"
-    y1="8.5"
-    x2="13.5"
-    y2="8.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-    />
-  <line
-    x1="8"
-    y1="5.5"
-    x2="17"
-    y2="5.5"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-    />
-  <line
-    x1="9.5"
-    y1="3.3"
-    x2="15.5"
-    y2="3.3"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-    />
-  <line
-    x1="10.5"
-    y1="1.25"
-    x2="14.5"
-    y2="1.25"
-    stroke="currentColor"
-    stroke-width="1.5"
-    stroke-linecap="round"
-  />     
-</svg> `;
+viewBox="0 0 25 25"
+version="1.1"
+xmlns="http://www.w3.org/2000/svg"
+xmlns:svg="http://www.w3.org/2000/svg">
+<line
+  x1="12.5"
+  y1="19.25"
+  x2="12.5"
+  y2="16.25"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+<line
+  x1="12.5"
+  y1="1.25"
+  x2="12.5"
+  y2="6.25"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+<line
+  x1="12.5"
+  y1="6.25"
+  x2="17"
+  y2="15.25"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+<line
+  x1="13.5"
+  y1="16.25"
+  x2="11.5"
+  y2="16.25"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+<line
+  x1="17"
+  y1="19.25"
+  x2="8"
+  y2="19.25"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />>
+<line
+  x1="15.5"
+  y1="21.45"
+  x2="9.5"
+  y2="21.45"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+<line
+  x1="14.5"
+  y1="23.5"
+  x2="10.5"
+  y2="23.5"
+  stroke="currentColor"
+  stroke-width="1.5"
+  stroke-linecap="round" />
+</svg>`;
 
 export const generalConductingEquipmentIcon = html`<svg
   xmlns="http://www.w3.org/2000/svg"
--- a/src/icons.ts
+++ b/src/icons.ts
@@ -456,53 +456,80 @@ export const voltageTransformerIcon = html`<svg
 
 export const earthSwitchIcon = html`<svg
   xmlns="http://www.w3.org/2000/svg"
-  viewBox="0 0 25 25"
->
+  viewBox="0 0 25 35"
+>  
   <line
     x1="12.5"
-    y1="2"
+    y1="5.5"
     x2="12.5"
-    y2="8"
+    y2="8.5"
     stroke="currentColor"
     stroke-width="1.5"
     stroke-linecap="round"
   />
   <line
     x1="11"
-    y1="23"
+    y1="23.5"
     x2="14"
-    y2="23"
+    y2="23.5"
     stroke="currentColor"
     stroke-width="1.5"
     stroke-linecap="round"
   />
   <line
     x1="12.5"
-    y1="23"
+    y1="23.5"
     x2="12.5"
-    y2="18"
+    y2="18.5"
     stroke="currentColor"
     stroke-width="1.5"
     stroke-linecap="round"
   />
   <line
     x1="12.5"
-    y1="18"
+    y1="18.5"
     x2="8"
-    y2="9"
+    y2="9.5"
     stroke="currentColor"
     stroke-width="1.5"
     stroke-linecap="round"
   />
   <line
     x1="11.5"
-    y1="8"
+    y1="8.5"
     x2="13.5"
-    y2="8"
+    y2="8.5"
     stroke="currentColor"
     stroke-width="1.5"
     stroke-linecap="round"
-  />
+    />
+  <line
+    x1="8"
+    y1="5.5"
+    x2="17"
+    y2="5.5"
+    stroke="currentColor"
+    stroke-width="1.5"
+    stroke-linecap="round"
+    />
+  <line
+    x1="9.5"
+    y1="3.3"
+    x2="15.5"
+    y2="3.3"
+    stroke="currentColor"
+    stroke-width="1.5"
+    stroke-linecap="round"
+    />
+  <line
+    x1="10.5"
+    y1="1.25"
+    x2="14.5"
+    y2="1.25"
+    stroke="currentColor"
+    stroke-width="1.5"
+    stroke-linecap="round"
+  />     
 </svg> `;
 
 export const generalConductingEquipmentIcon = html`<svg
--- a/src/wizards/conductingequipment.ts
+++ b/src/wizards/conductingequipment.ts
@@ -120,7 +120,7 @@ function containsEarthSwitchDefinition(condEq: Element): boolean {
     const swTypVal = getDataAttributeValue(lNode, 'SwTyp', undefined, 'stVal');
     return swTypVal ? ['Earthing Switch', 'High Speed Earthing Switch'].includes(swTypVal) : false;
   } else {
-  return false; 
+    return false; 
   }
 }
 
--- a/src/wizards/conductingequipment.ts
+++ b/src/wizards/conductingequipment.ts
@@ -7,6 +7,7 @@ import '@material/mwc-select';
 import '../wizard-textfield.js';
 import {
   createElement,
+  crossProduct,
   EditorAction,
   getValue,
   isPublic,
@@ -15,7 +16,6 @@ import {
   WizardInput,
 } from '../foundation.js';
 import { updateNamingAction } from './foundation/actions.js';
-import { getDirections } from '../editors/singlelinediagram/sld-drawing.js';
 
 const types: Partial<Record<string, string>> = {
   // standard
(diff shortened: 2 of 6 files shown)


JakobVogelsang:
I am wondering where you got the icon from. Initially, the vertical line at the bottom is representing the earth connection. You see that with the voltage transformer for example as well (`VTR`). Like you are proposing it at the moment, I would say it is earthed on both terminals. You said there is a standard for such elements: is it how it is recommended there?

JakobVogelsang:
Also, as you are on this topic already I also would encourage you to add some more logic into distinguishing earth switches, that is proposed in this issue #267.

danyill:
> I am wondering where you got the icon from. Initially, the vertical line at the bottom is representing the earth connection.

Ah, that was not at all obvious to me. Perhaps that's a European standard I'm unfamiliar with:

FWIW, IEC 60617 uses the following for a VT:

image

And for a combined disconnector/earthing switch:

image

And for a disconnector:

image

But I cannot find a clean combined symbol called an "earth switch" or "earthing switch". So the symbol is literally intended to connect the two symbols.

Our in-house standard symbol for an earth switch is:

image

I've combined the existing disconnector icon in open-scd with the standard "earth" symbol but I can see that the bar could be removed.

How do you feel about: 

image

I had a hunt for an international standard for say, "HMI symbols" but couldn't find anything.
 
Possibly there would be something in IEC 60417:2002 DB or ISO 7000 but I don't find anything new here: https://www.iso.org/obp/ui/#iso:pub:PUB400001:en

> already I also would encourage you to add some more logic into distinguishing earth switches, that is proposed in this issue #267.

:+1: I'll take a look.

JakobVogelsang:
Yeah, sorry that was my bet. Had a look as well, and your proposal is the one most of the world is using. One small thing from my side though. Let's put the grounding symbol at the bottom. This would look better in the Substation editor.

danyill:
> Let's put the grounding symbol at the bottom. This would look better in the Substation editor.

Sure, will do

src/zeroline/foundation.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { BayEditor } from './bay-editor.js';
 import { SubstationEditor } from './substation-editor.js';
 import { VoltageLevelEditor } from './voltage-level-editor.js';
+import { typeStr } from '../wizards/conductingequipment.js';
```
danyill, line 15: I decided to pick one to pull from the other for this function rather than the current implementation where the function is duplicated but this may be for good reason. Should it go in `src/foundation.ts` instead and be imported from both places?

test/testfiles/conductingequipmentwizard.scd
```diff
@@ -0,0 +1,697 @@
...
+				<ConductingEquipment name="ES269" sxy:dir="vertical" sxy:x="1" sxy:y="1" virtual="false" type="DIS">
+					<LNode iedName="BC1" ldInst="Application" lnClass="XSWI" lnInst="2" lnType="LNType_XSWI_Ctl_2" />
+					<Terminal bayName="B262" cNodeName="L1" name="L1" substationName="ERE" voltageLevelName="V220" connectivityNode="ERE/V220/B262/L1" />
+					<Terminal bayName="B262" cNodeName="L2" name="L2" substationName="ERE" voltageLevelName="V220" connectivityNode="ERE/V220/B262/L2" />
+				</ConductingEquipment>
+				<ConnectivityNode name="L2" pathName="ERE/V220/B262/L2" />
+			</Bay>
+			<Bay name="B272" sxy:dir="vertical" sxy:x="5">
+				<ConductingEquipment name="ES279" sxy:dir="vertical" sxy:x="1" sxy:y="1" virtual="false" type="DIS">
+					<SubEquipment name="A" phase="A" virtual="false">
```
danyill, line 42: When this scd file is opened in open-scd it fails validation, but I think this `SubEquipment` definition is correct :confused:

test/testfiles/conductingequipmentwizard.scd
```diff
@@ -0,0 +1,697 @@
...
+									<Val>0</Val>
+								</DAI>
+								<DAI name="numPls">
+									<Val>1</Val>
+								</DAI>
+							</SDI>
+						</DOI>
+						<DOI name="BlkOpn" desc="Block opening" />
+						<DOI name="BlkCls" desc="Block closing" />
+						<DOI name="SwTyp" desc="Switching-device type">
```
danyill, line 151: This has been hand-crafted by me and is not the output of a tool. Please consider if it is a correct or sane thing to write in SCL.

test/testfiles/conductingequipmentwizard.scd
```diff
@@ -0,0 +1,697 @@
...
+							<DAI name="stVal">
+								<Val>Earthing Switch</Val>
```
danyill, line 184: This has been hand-crafted by me and is not the output of a tool. Please consider if it is a correct or sane thing to write in SCL.

test/testfiles/conductingequipmentwizard.scd
```diff
@@ -0,0 +1,697 @@
...
+			<DA fc="CO" name="Oper" bType="Struct" type="DAType_Oper_Boolean" />
+		</DOType>
+		<DOType id="DOType_ENS_SwitchType" cdc="ENS">
+			<DA dchg="true" fc="ST" name="stVal" bType="Enum" type="EnType_SwitchType" />
+			<DA qchg="true" fc="ST" name="q" bType="Quality" />
+			<DA fc="ST" name="t" bType="Timestamp" />
+		</DOType>
+		<DOType id="DOType_ENS_SwitchType_2" cdc="ENS">
+			<DA dchg="true" fc="ST" name="stVal" bType="Enum" type="EnType_SwitchType">
+				<Val>Earthing Switch</Val>
```
danyill, line 381: This has been hand-crafted by me and is not the output of a tool. Please consider if it is a correct or sane thing to write in SCL.

test/testfiles/conductingequipmentwizard.scd
```diff
@@ -0,0 +1,697 @@
...
+			<Bay name="Bus_A1" desc="Bus A1" sxy:x="0" sxy:y="3">
+				<ConnectivityNode name="L1_232" pathName="ERE/V220/B232/L1" />
+				<ConnectivityNode name="L1_242" pathName="ERE/V220/B242/L1" />
+				<ConnectivityNode name="L1_252" pathName="ERE/V220/B252/L1" />
+				<ConnectivityNode name="L1_262" pathName="ERE/V220/B262/L1" />
+				<ConnectivityNode name="L1_272" pathName="ERE/V220/B272/L1" />
+			  </Bay>
+		</VoltageLevel>
+	</Substation>
+	<IED desc="ERS Bay Controller" name="BC1" type="ERS_Bay_Controller" manufacturer="FreedomInc" configVersion="V00.01.00" originalSclVersion="2007" originalSclRevision="B" owner="IEC station 1" originalSclRelease="4">
```
danyill, line 59: I've bodged the output from Siemens DIGSI 5 with a few of my own ideas. Hopefully it makes sense.

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
+ * @param lNode - LNode within the IED section.
+ * @param doName - name for data object (e.g. SwTyp).
+ * @param sdName - id for sub data object.
+ * @param daName - name for data attribute (e.g. stVal).
+ * @returns - value of type as a string.
+ */
+function getDataAttributeValue(lNode: Element, doName: string, sdName: string | undefined, daName: string): string | undefined {
+  const sdDef = sdName ? ` > SDI[name='${sdName}']` : '';
+  const daInstantiated = lNode.querySelector(`DOI[name='${doName}']${sdDef} > DAI[name='${daName}'`);
+  if (daInstantiated) {
```
danyill, line 81: I object to my nested `if` statements here but I can't currently see a more elegant approach. Happy to receive instruction.

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
+    // data attribute is fully instantiated within a DOI/SDI, look for it within: DOI > ?SDI > DAI
+    return daInstantiated.querySelector('Val')?.innerHTML.trim();
+  } else {
+    const rootNode = <Element>lNode?.getRootNode();
```
danyill, line 85: is `getRootNode` safe?
Is it better that I do: `lNode?.ownerDocument;`?

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
+ * includes SwTyp and if stVal indicates an Earth/Earthing Switch.
+ * @param condEq - SCL ConductingEquipment
+ * @returns true if an earth switch is found, false otherwise.
+ */
+function containsEarthSwitchDefinition(condEq: Element): boolean {
+  const lNodeXSWI = condEq.querySelector("LNode[lnClass='XSWI']");
+  const lNode = getLNodefromIED(lNodeXSWI);
+  if (lNode) {
+    const swTypVal = getDataAttributeValue(lNode, 'SwTyp', undefined, 'stVal');
+    return swTypVal ? ['Earthing Switch', 'High Speed Earthing Switch'].includes(swTypVal) : false;
```
danyill, line 134: Is it acceptable for these values be hard-coded?

danyill:
I'd be very pleased to take a review of this, please be kind, my TypeScript ability is modest :blush:

JakobVogelsang:
@danyill I am on the review, please be patient with me. It will take this week, still. Sorry the delay :)

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
+
+/**
+ * Finds data attribute by inspecting an IED's logical node or types.
+ * @param lNode - LNode within the IED section.
+ * @param doName - name for data object (e.g. SwTyp).
+ * @param sdName - id for sub data object.
+ * @param daName - name for data attribute (e.g. stVal).
+ * @returns - value of type as a string.
+ */
+function getDataAttributeValue(lNode: Element, doName: string, sdName: string | undefined, daName: string): string | undefined {
```
JakobVogelsang, line 78: The function works and does that it has to do. I would argue that you have created a very flexible function allowing you to get Val for any give do or sdo, da. This is not really neccassary here. We know from IEC 61850-7-4 that `SwTyp` is a direct child of `XSWI`. This will reduce the complexity in the function a lot. 
If you want to write something more general I thing using recursive functions are better sutted here. 

In this case I would prefer to have a dedicated function for what we want to achive. I don't know something like 

`getSwitchTypeValue(ln:Element): string | null { ... }`

test/unit/editors/substation/conductingequipmentwizard.test.ts
```diff
@@ -0,0 +1,56 @@
+import { expect } from '@open-wc/testing';
```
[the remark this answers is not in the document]
JakobVogelsang: I very much like your test. I this is how I know: A started to review and after 1 min I was done :) Nice job on this one!

test/testfiles/conductingequipmentwizard.scd
JakobVogelsang, line 59: `FreedomInc` love that :)

src/wizards/conductingequipment.ts
```diff
@@ -15,6 +15,7 @@ import {
   WizardInput,
 } from '../foundation.js';
 import { updateNamingAction } from './foundation/actions.js';
+import { getDirections } from '../editors/singlelinediagram/sld-drawing.js';
```
JakobVogelsang, line 18: What is this import for?

JakobVogelsang (commented):
I have added some some comments but nothing for the functionality, I would like to test it with some files I have. In generall however Looks Good To Me (LGTM)

danyill:
Thank you for this review and your constructive comments. I shall improve over the next few days.

src/wizards/conductingequipment.ts
```diff
@@ -51,7 +51,9 @@ const types: Partial<Record<string, string>> = {
 
 function typeStr(condEq: Element): string {
   return condEq.getAttribute('type') === 'DIS' &&
-    condEq.querySelector('Terminal')?.getAttribute('cNodeName') === 'grounded'
+    Array.from(condEq.querySelectorAll('Terminal'))
+      .map(t => t.getAttribute('cNodeName'))
```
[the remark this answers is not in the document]
danyill: Nice. Thanks.

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
-  return condEq.getAttribute('type') === 'DIS' &&
-    condEq.querySelector('Terminal')?.getAttribute('cNodeName') === 'grounded'
-    ? 'ERS'
-    : condEq.getAttribute('type') ?? '';
+/**
+ * Finds Logical Node from IED based on Logical Node in the Substation SCL section.
+ * @param lNode - Logical Node in Substation SCL section.
+ * @returns - Logical Node in the IED SCL section.
+ */
+function getLNodefromIED(lNode: Element | null): Element | null {
```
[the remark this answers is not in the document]
danyill: You are so right, I missed this and it was clearly part of 6.3. This is a good distinction to make here.

> NOTE Figure 6 shows with its LNode class the LN object, whose instances can be referenced or represented in
> SCL in two ways. The LNode element resides in the Substation structure, while the LN element resides in the IED
> structure.

src/wizards/conductingequipment.ts
danyill: It shall be as you wish. I'll avoid writing a perfectly general function as it may stretch my understanding too far and cause other problems.

src/wizards/conductingequipment.ts
danyill: Whoops, taken from some other playing around. Good spotting :clap:

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,105 @@ const types: Partial<Record<string, string>> = {
...
+  if (!lNode) return null;
+  const [iedName, ldInst, lnClass, lnInst, lnType] = [
+    'iedName',
+    'ldInst',
+    'lnClass',
+    'lnInst',
+    'lnType',
+  ].map(attribute => lNode?.getAttribute(attribute));
+  return (<Element>lNode?.getRootNode()).querySelector(`IED[name='${iedName}'] > AccessPoint > Server > LDevice[inst='${ldInst}'] > LN[inst='${lnInst}'][lnType='${lnType}'][lnClass='${lnClass}']`);
```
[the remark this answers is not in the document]
danyill: > 1.) I would argue it is not complete as the prefix is missing. I don't know if we ca relay the lnInst makes the logical node complete.

Good point. Thanks. :+1: 

Looking at Table 8 of IEC 61850-6, I notice this is discussed:

> The LN prefix used in the IED (if needed; default, if not specified, is the empty string). Can
> be used for more detailed function specification than possible by LN class alone, if the LN
> is not allocated to an IED

But yes, if specified it must be handled. I shall add support for this.

I notice there is also this possibility discussed a little below which this PR really should address:

> For specifications where iedName=”None” everywhere, the combination of the other attributes
> must be unique within the same level. This means e.g. that the prefix or lnInst should be
> different if several LNs with identical lnClass are used within the same substation part (i.e.
> same bay). This should also be the case for IED names, if application related naming (see
> 8.5.4) shall be used. This is NOT checked by the SCL schema, therefore it is the
> responsibility of the project engineer or system tool, if application related naming shall be
> used additionally to product (IED)-related naming.

I will use the rather delightful `crossProduct` to do this also.

> 2.) Taken prefix into account we ususally handele missing prefix null and missing prefix empty string as the same. You can use or in selcetors with , and then [:not(prefix)] [prefix=""]. Chris has writen a nice function crossProduct that creates a nice selector.

It took me some :hourglass: to understand, but it is nice :grinning: 

> 3.) using lNode.ownerDocument saves you the type cast

:+1: 

> 4.) Use a formatter to hadle code formatting for you. I use VSCode and prettier. This is not only bit more readable for reiviews but helps you when working in code. :)

Will do. I also use VSCode and can use prettier.

src/zeroline/foundation.ts
JakobVogelsang: This is ok for me. I think the function is very specific to `ConductingEqipment` to not be part of faoundation. :)

src/icons.ts
```diff
@@ -455,55 +455,67 @@ export const voltageTransformerIcon = html`<svg
 </svg>`;
 
 export const earthSwitchIcon = html`<svg
-  xmlns="http://www.w3.org/2000/svg"
```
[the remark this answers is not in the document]
JakobVogelsang: There are some question that I see and I'll try to answer those one after the other. If I missed one please give me a hint:
1.) I like your proposal to group CSS rules either within the `icon` or outside. The second one we need to define a class that shall be applied to all `ConductingEquipment` type icon.
2.) We can try to play around with the CSS. I can imagine that it would be very benefitial when icons become small as in the SLD to not have round edges.
For both the questions I would propose to have it in a extra issue and PR. For me it woud be great to have the same rules and code style applied to all icons. I would propese to combine it with adding more icons to the code base. This would be very beneficial for the end user I fully aggree!! I opened an [Issue](#509) for that, hope this is ok for you.

One this within the PR I would like to ask you to change. The ""switch" is the other way around for the CBR and DIS. This jupms into my eye. I don't know which one is correct, but would love to be them in the same direction. 
grafik

src/wizards/conductingequipment.ts
```diff
@@ -49,11 +50,118 @@ const types: Partial<Record<string, string>> = {
...
+    'ldInst',
+    'prefix',
+    'lnType',
+  ].map(attribute => lNode?.getAttribute(attribute) || '');
+  const iedSelector = [`IED[name="${iedName}"]`, 'IED'];
+  const lDevicePath = ['AccessPoint > Server'];
+  const lDeviceSelector = [
+    `LDevice[inst="${ldInst}"] > LN[inst="${lnInst}"][lnType="${lnType}"][lnClass="${lnClass}"]`,
+  ];
+  const lDevicePrefixSelector = [`[prefix="${prefix}"]`, ':not(prefix)'];
```
JakobVogelsang, line 73: Here are two things:
1.) the name is missleading as prefix is an attribute of the logical node
2.) I think the way crossProduct is sopposed to be used is to look for a prefix, if prefix is there and not a empty string than look for the definition. In all other cases allow it to be missing or an empty string.

```
const ... = prefix && prefix !=='' ? [`prefix="${prefix}"`, ['prefix=""',':not(prefix)']]
```

test/testfiles/conductingequipmentwizard.scd
JakobVogelsang: The order of the elements is defined through the schema as well. For the `complexType` `tConductingEquipment` you have everyting defined within the abstract type `tAbstractConductingEquipment` and the `EqFunction`.

```
<xs:sequence>
                      <xs:element form="qualified" maxOccurs="unbounded" minOccurs="0"
                          name="EqFunction" type="scl:tEqFunction">
                          <xs:unique name="uniqueLNodeInFuncForCE">
                              <xs:selector xpath="./scl:LNode"/>
                              <xs:field xpath="@lnInst"/>
                              <xs:field xpath="@lnClass"/>
                              <xs:field xpath="@iedName"/>
                              <xs:field xpath="@ldInst"/>
                              <xs:field xpath="@prefix"/>
                          </xs:unique>
                          <xs:unique name="uniqueChildNameInFuncForCE">
                              <xs:selector xpath="./*"/>
                              <xs:field xpath="@name"/>
                          </xs:unique>
                      </xs:element>
                  </xs:sequence>
```
Lokking deeper into `TAbstractConductingEquipment` we see that we have everything within `tEqipment` and then `Terminal` followed by `SubEquipment`. And this is the reason why the schema validator does alarm you. We have a extra function `getReference` in the foundation to make sure the sequence is correct when adding elements to the SCD. 

```
<xs:complexType abstract="true" name="tAbstractConductingEquipment">
          <xs:complexContent>
              <xs:extension base="tEquipment">
                  <xs:sequence>
                      <xs:element form="qualified" maxOccurs="2" minOccurs="0" name="Terminal"
                          type="tTerminal"/>
                      <xs:element form="qualified" maxOccurs="unbounded" minOccurs="0"
                          name="SubEquipment" type="tSubEquipment">
                          <xs:unique name="uniqueLNodeInSubEquipment">
                              <xs:selector xpath="./scl:LNode"/>
                              <xs:field xpath="@lnInst"/>
                              <xs:field xpath="@lnClass"/>
                              <xs:field xpath="@iedName"/>
                              <xs:field xpath="@ldInst"/>
                              <xs:field xpath="@prefix"/>
                          </xs:unique>
                          <xs:unique name="uniqueChildNameInACESubEquipment">
                              <xs:selector xpath="./*"/>
                              <xs:field xpath="@name"/>
                          </xs:unique>
                      </xs:element>
                  </xs:sequence>
              </xs:extension>
          </xs:complexContent>
      </xs:complexType>
```

test/unit/editors/substation/conductingequipmentwizard.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    it('has no grounded connectivityNodes but has an XSWI LN and DOI:SWTyp > DAI:stVal defining an "Earthing Switch"', () => {
+      expect(
+        typeStr(
+          doc.querySelector('ConductingEquipment[name="ES259"]')!
+        )
+      ).to.equal('ERS');
+    });
+    
+    it('has no grounded connectivityNodes but has an XSWI LN and DOI:SWTyp > DOType > DA:stVal defining an "Earthing Switch"', () => {
```
JakobVogelsang, line 39: The description of this test is missleading `DIO:SwType > DOType` especially. I really dont' know what this means. DOType is never a child element of DOI.

test/unit/editors/substation/conductingequipmentwizard.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    it('has no grounded connectivityNodes but has an XSWI LN within SubEquipment and DOI:SWTyp > DAI:stVal defining an "Earthing Switch"', () => {
+      expect(
+        typeStr(
+          doc.querySelector('ConductingEquipment[name="ES279"]')!
+        )
+      ).to.equal('ERS');
+    });
+
+    it('has no grounded connectivityNodes but has an IED with name "None" and an XSWI LN with a prefix and DOI:SWTyp > DAI:stVal defining an "Earthing Switch"', () => {
```
JakobVogelsang, line 55: Tthis test should ne work. The attribute `iedName` can only be `None` if the IED does not exist and then DOI and DAI cannot exist either. What can happen in this case is that `LNode` has a `lnType` defined that is referencing a `LNodeType` element that can have a `SwType` and so on...

test/unit/editors/substation/conductingequipmentwizard.test.ts
```diff
@@ -0,0 +1,64 @@
+import { expect } from '@open-wc/testing';
+
+import { typeStr } from '../../../../src/wizards/conductingequipment.js';
+
+describe('conductingequipmentwizard', () => {
```
JakobVogelsang, line 5: From what I can see you test in all following tests that the Terminal is `grounded` of that the `SwType` is defined within the `IED` section but not within the `DataTypeTemplates` section. Otherwise you would have seen a faulty test.

JakobVogelsang (changes requested):
This time I have tested your PR with some files I have found and found some bugs. I don't know if they were introduced triggered by my review :( I fthis is the case sorry for that. Your time is limited and I force you to spend so much time...the more I thank you for comtributing on your free time!!! For me that is a sign that my contribution on a Sunday is worth it.

src/wizards/conductingequipment.ts
danyill: Thanks, I understand better now.

test/unit/editors/substation/conductingequipmentwizard.test.ts
danyill: It should have been:

```diff
- 'has no grounded connectivityNodes but has an XSWI LN and DOI:SWTyp > DOType > DA:stVal defining an "Earthing Switch"'
+ 'has no grounded connectivityNodes but has an XSWI LN and LNodeType > DOType > DA:stVal defining an "Earthing Switch"'
```
I've made this more obvious in the test because I think I had left too much information in the scd file for the test to function correctly.

test/unit/editors/substation/conductingequipmentwizard.test.ts
danyill: Ah. Good, then we must also support this.

test/unit/editors/substation/conductingequipmentwizard.test.ts
danyill: I have corrected this.

src/icons.ts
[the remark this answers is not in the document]
danyill: As per separate issue.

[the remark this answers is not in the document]
danyill: I think it is probably correct to have the switch arm connected to earth (most ESes that I ever saw "reach out" from the earthed side) and I think whether it is on left or right is "house rules" so that seems good to me.

danyill:
Thank you for this thorough review. If you have patience I have time :wink: plus it's a holiday today.

I have added support for an `LNode` definition only via a `LNodeType` with a test and made other corrections and improvements as requested, found a few other errors and I hope we're getting close.

JakobVogelsang (approved):
LGTM! This works perfectly and I like the icon change!

PR #461, opened 2022-01-03
feat(editors/ied): add DA elements to IED editor

Flurb:


src/editors/ied/da-container.ts
```diff
@@ -0,0 +1,118 @@
...
+  render(): TemplateResult {
+    const bType = this.element!.getAttribute('bType');
+    return html`<action-pane .label="${this.header()}" icon="${this.instanceElement != null ? 'done' : ''}">
+      <h6>${this.renderValue()}</h6>
+      ${bType == 'Struct' ? this.getBDAElements().map(element =>
+        html`<da-container
+          .element=${element}>
+        </da-container>`) : nothing}
+      ${bType == 'Enum' ? this.getEnumElements().map(element =>
+        html`<enum-container
```
JakobVogelsang, line 98: I would suggest leaving the `enum-container` out here. Logically, it does not belong to the IED section. It is a restraint when initializing a value and therefore should be displayed to the user when selecting or changing the value (`Val`). Other than that, it adds unnecessary complexity.

src/editors/ied/da-container.ts
```diff
@@ -0,0 +1,118 @@
...
+    `;
+  }
+
+  static styles = css`
+    h6 {
+      color: var(--mdc-theme-on-surface);
+      font-family: 'Roboto', sans-serif;
+      font-weight: 500;
+      font-size: 0.8em;
+      overflow: visible;
```
JakobVogelsang, line 111: I would hide the value if overflows. Otherwise, is gets into conflict with the outline of the parent action-pane. If you want, you can put the value into a `abbr` field - tooltip. Please see the CSS we use for `abbr`

src/editors/ied/enum-container.ts
```diff
@@ -0,0 +1,24 @@
+import {
```
JakobVogelsang, line 1: Again, would leave it out. But let's discuss this in a bigger round.

JakobVogelsang (commented):
LGTM. The code is nice and very clean. Special thank for the comments in your code.

I would leave the enumeration container out, as commented in several lines. I think belongs logically into the Template editor rather than in the IED editor.

src/editors/ied/da-container.ts
```diff
@@ -0,0 +1,118 @@
...
+      return Array.from(doType!.querySelectorAll(':scope > BDA'))
+    }
+    return [];
+  }
+
+  /**
+   * Get the nested EnumVal element(s) if available.
+   * @returns The nested EnumVal element(s) of this (B)DA container.
+   */
+  private getEnumElements(): Element[] {
```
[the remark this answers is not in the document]
Flurb: That's okay, I will remove the possible Enum values!

Flurb:
Hi @JakobVogelsang, I removed the Enum values.
I also added a little action on the LN element and below, I'm interested in your opinion.
It's an action to toggle child elements, what do you think about it?

https://user-images.githubusercontent.com/16406106/148461315-54c1a986-f2e7-4d38-a62c-2066479af3c6.mov

JakobVogelsang:
I love it. Btw. that was a very, very fast response. If you ask me, we can go with this!! Thanks

Flurb:
Cool!
Going to add some tests as well, committing it and then it can be merged.

JakobVogelsang:
I just tried it out and it works very nice with chrome but not with firefox.

Flurb:
People actually use Firefox? Just kidding.
In Firefox, the `toggle` function is not called when clicking the toggle action.
Let's see if I can fix it.

Flurb:
The `:scope` seems not being supported in Firefox. If I remove them, it works.
According to the :scope documentation (https://developer.mozilla.org/en-US/docs/Web/CSS/:scope), it should work.

Shall I make an issue about this and leave it for now? @JakobVogelsang

Flurb:
Btw, if someone has a better idea for defining one `toggle` method instead of defining it in every container, let me know.
There some duplicate code now which I don't really like (but it works!) :)

Flurb:
Already added an issue about `:scope` issues:
https://github.com/openscd/open-scd/issues/463

JakobVogelsang:
I did add a proposal for another way to toggle. I added it to the level DO a below. On the logical node level, I would for performance issues use another functionality. Although, `hidden` already makes a big impact as we save the render. However, it seems to me that still every DO and DA is walked through. I therefore would propose to create a `do-container` at all. This would then resolve in something like that 

```
render (): ... {
    ....
    ${this.toggleButton.on ? html`<do-container>...` : nothing}
}
```

instead of 

```
render (): ... {
    ....
    ${... html`<do-container hidden=...>...}
}
```

JakobVogelsang (approved):
LGTM

PR #464, opened 2022-01-08
refactor(ied/da-container,do-container): code improvements

JakobVogelsang:
@Flurb  as you did not like the toggle function I tried to refactor. What do you think.

I do get snapshot updates that I cannot interpret. Maybe you can have a look.

### DIFF
--- a/web-test-runner.config.mjs
+++ b/web-test-runner.config.mjs
@@ -12,7 +12,7 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
    * Plugins have a fix URL and do not fit to the file structure in test environment.
    * Creating open-scd in the tests leads to error in the browser log - we had to disable the browser log
   */
-  browserLogs: false,
+  browserLogs: true,
 
   /** specify groups for unit and integrations tests 
    * hint: no --group definition runs all groups
--- a/web-test-runner.config.mjs
+++ b/web-test-runner.config.mjs
@@ -12,7 +12,7 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
    * Plugins have a fix URL and do not fit to the file structure in test environment.
    * Creating open-scd in the tests leads to error in the browser log - we had to disable the browser log
   */
-  browserLogs: true,
+  browserLogs: false,
 
   /** specify groups for unit and integrations tests 
    * hint: no --group definition runs all groups
--- a/src/editors/ied/da-container.ts
+++ b/src/editors/ied/da-container.ts
@@ -10,6 +10,9 @@ import {
 import { nothing } from 'lit-html';
 import { translate } from 'lit-translate';
 
+import '@material/mwc-icon-button-toggle';
+import { IconButtonToggle } from '@material/mwc-icon-button-toggle';
+
 import '../../action-pane.js';
 import { getNameAttribute } from '../../foundation.js';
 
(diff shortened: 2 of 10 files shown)


Flurb:
Way better! Using standard components instead of your own implementation is always better.
I couldn't find one, but I think I might have searched wrong, mwc-icon-button-toggle is perfect here.

Flurb (changes requested):
Looks good, but during testing is see that the icon isn't changing when toggling.

Flurb (changes requested):
Also, the same toggle principle can be added to the ln-container :)

JakobVogelsang:
> Also, the same toggle principle can be added to the ln-container :)

This was on purpose from my side (see #461). My opinion is that we need some other mechanism here. Do you agree? Do I miss something?

JakobVogelsang:
> Way better! Using standard components instead of your own implementation is always better. I couldn't find one, but I think I might have searched wrong, mwc-icon-button-toggle is perfect here.

This is bug of this component that only is present in debug mode. Just for you to compare, see the buttons in the log.

Flurb:
> > Way better! Using standard components instead of your own implementation is always better. I couldn't find one, but I think I might have searched wrong, mwc-icon-button-toggle is perfect here.
> 
> This is bug of this component that only is present in debug mode. Just for you to compare, see the buttons in the log.

Then it's fine, hope it will be fixed soon!

Flurb:
> > Also, the same toggle principle can be added to the ln-container :)
> 
> This was on purpose from my side (see #461). My opinion is that we need some other mechanism here. Do you agree? Do I miss something?

Why would this functionality be different for a LN container compared to DO/DA containers?

JakobVogelsang:
> > > Also, the same toggle principle can be added to the ln-container :)
> > 
> > 
> > This was on purpose from my side (see #461). My opinion is that we need some other mechanism here. Do you agree? Do I miss something?
> 
> Why would this functionality be different for a LN container compared to DO/DA containers?

Because rendering the data model for a complicated IED does take some seconds which in my opinion is not a good user experience. The `hidden` attribute is a performance problem, but as is, we still walk through every DO and DA there is in a data model of an IED. We don't need to do that. I think we should on first render only consider the data model up until the logical nodes and only on toggle button trigger should dive into the DOs and DAs. :)

Flurb (approved):

Flurb (changes requested):
I'm going to implement a solution for all three the containers, should not be a problem :)

Flurb:
@JakobVogelsang (or anyone else) please review :)

web-test-runner.config.mjs
```diff
@@ -12,7 +12,7 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
    * Plugins have a fix URL and do not fit to the file structure in test environment.
    * Creating open-scd in the tests leads to error in the browser log - we had to disable the browser log
   */
-  browserLogs: false,
+  browserLogs: true,
```
JakobVogelsang, line 15: Just out of curiosity. Why this? Do we need it for something? Due to our plugin architecture, the browser log shows a misleading error when calling `Hosting`. This makes the log hard to read. But if you think we require it for something, I am fine with it.

Flurb: No we don't, I just like to use console.logs during testing. For this we need to set it to true, and it should be reverted which I didn't do :)

JakobVogelsang:
LGTM @Flurb

Flurb (approved):

PR #477, opened 2022-01-12
feat(zeroline): show SampledValueControl for IED and whole project

### LINKED ISSUE #426: Show sampled value control blocks
As a system specification engineer, I want to see if any and how many `SampledValueControl` element are defined in the project and within an IED element.

**Requirements**
- allow filtering for `SampledValueControl` blocks for name attribute
- allow filtering for parents LDevice, IED and LN/LN0

JakobVogelsang:
With `SampledValueControl ` element, I thought to make smaller steps immediately usable by the user. Showing `SampledValueControl` element is already interesting enough :)

Closes #426

### DIFF
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -66,6 +66,7 @@ export const de: Translations = {
     iedsloading: 'IEDs werden geladen...',
     commmap: 'Kommunikationszuordnung',
     gsecontrol: 'GOOSEs anzeigen',
+    smvcontrol: 'Sampled Values anzeigen',
   },
   editing: {
     created: '{{ name }} hinzugefügt',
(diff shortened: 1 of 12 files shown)


src/translations/de.ts
```diff
@@ -162,7 +163,7 @@ export const de: Translations = {
     searchHelper: 'IED auswählen',
     searchHelperDesc: '({{description}})',
     missing: 'Kein IED vorhanden',
-    toggleChildElements: "???"
```
dlabordus, line 165: Shouldn't this be translated? Don't know how it came here originally.

dlabordus (approved):

src/translations/de.ts
JakobVogelsang: I missed it during the review. There is a very similar bug to fix the translation within an abbreviation. I will add there and leave it in this PR.

Issue #467, opened 2022-01-11
Change icon of the IED editor


JakobVogelsang:
The icon of the IED editor is miss-leading, if you ask me. In the substation editor, the icon for an editor is a different one.

Sander3003:
Any suggestion what the Icon of the IED editor should be?

JakobVogelsang:
I personally like the one we are using in the Substation section for the IED that is a material design icon: https://fonts.google.com/icons?icon.query=developer+board

JakobVogelsang:
Using icons from this library assures that we follow the material design guidelines and this one fit quite well to the IED I think :)

Sander3003:
Sounds good to me

dlabordus:
Small remark when the pull request is merge. A reset of the extensions is needed to reload the Editor with the new icon.

PR #481, opened 2022-01-13
feat(editors/ied): Changed icon of IED Editor

dlabordus:


JakobVogelsang:
It is a bit strange that the second commit is the whole snapshot file. I would have guessed that only the icon name is changed. Did you not have the snapshot on your machine? 


Closes #467 .

Flurb (approved):

PR #489, opened 2022-01-17
refactor(editor/communication): move wizards to wizard library

JakobVogelsang:
- Mainly moved wizards to wizards-library. While doing this also improved code with wizards and added more extensive tests.
- In cooperated `action-pane` with have a similar look and feel like the Substation and IED editor type plug-ins.


src/editors/communication/subnetwork-editor.ts
```diff
@@ -84,30 +84,8 @@ export class SubNetworkEditor extends LitElement {
   }
 
   private renderHeader(): TemplateResult {
```
ca-d, line 108: Why do we still use a `renderHeader` method returning a `TemplateResult` here? Isn't all we need a string? Does the `mdash` thing still work within the attribute?

src/editors/Communication.ts
```diff
@@ -60,24 +60,32 @@ export default class CommunicationPlugin extends LitElement {
...
+            html`<subnetwork-editor .element=${subnetwork}></subnetwork-editor>`
+        )}
+      </section> `;
   }
 
   static styles = css`
-    ${styles}
+    :host {
+      width: 100vw;
+    }
```
ca-d, line 74: empty line disappeared here

src/editors/communication/subnetwork-editor.ts
```diff
@@ -114,6 +101,19 @@ export class SubNetworkEditor extends LitElement {
       );
   }
 
+  private renderSubNetworkSpecs(): TemplateResult {
```
ca-d, line 100: I don't understand this pattern of using `TemplateResult`s and passing them to our components as a `.label` property, where they will be wrapped in a string. I'm really stunned that any of this works, especially the `&mdash;` - does it work in practice or does it render a regular dash?

src/editors/communication/subnetwork-editor.ts
```diff
@@ -76,17 +76,19 @@ export class SubNetworkEditor extends LitElement {
   }
 
   private renderIedContainer(): TemplateResult[] {
-    return Array.from(this.element.querySelectorAll('ConnectedAP') ?? [])
+    return Array.from(
+      this.element.querySelectorAll(':scope > ConnectedAP') ?? []
```
ca-d: Great idea using `:scope` here! 👍

src/wizards/subnetwork.ts
```diff
@@ -234,3 +194,113 @@ export function subNetworkWizard(options: WizardOptions): Wizard {
     },
   ];
 }
+
+function getBitRateAction(
```
ca-d, line 124: I can't currently really review the `update edit wizard` commit, since it moves code it possibly changes, so the diff is not very useful to me. I hope you can help me understand this change later.

src/editors/communication/subnetwork-editor.ts
```diff
@@ -45,12 +44,12 @@ export class SubNetworkEditor extends LitElement {
   /** SubNetwork child elements BitRate label */
   @property()
   get bitrate(): string | null {
-    const V = this.element.querySelector('BitRate');
-    if (V === null) return null;
-    const v = V.textContent ?? '';
-    const m = V.getAttribute('multiplier');
+    const Br = this.element.querySelector('BitRate');
```
ca-d, line 46: When reading a constant called `Br` and another called `br` I can't tell the difference in meaning from the difference in name.

Also, the idiomatic way to write `BitRate` as a constant name would be `bitRate` and its abbreviation would then be `bR`, since constants that don't refer to constructors always start with a lower-case letter.

ca-d (changes requested):
I'm still not done with the review, as I noted I have trouble with one of the commits.

As we talked about earlier, this review made clear to me once again how much easier reviewing a branch is when the branch in question does just one thing. If this branch contained *only* the refactor (as it says in the title) and no drive-by fixes and style improvements, the review would have been done in a few minutes. The same would then probably be true for the *separate* `fix:` and `style:` branches containing the fixes.

In general, I don't enjoy this mode of reviewing branches by individual commits. What I want to review is the result of your work, not the process you took to get there.

src/wizards/subnetwork.ts
ca-d: Thanks for clarifying that this change consisted only in introducing `ContentOptions`.

JakobVogelsang:
> 

This is a good hint, and I get what you mean. I in the future will create one branch per single issue/step to allow you to review faster.

src/editors/communication/subnetwork-editor.ts
JakobVogelsang: That was @Flurb idea in some of his PRs :)

src/editors/communication/subnetwork-editor.ts
```diff
@@ -23,21 +23,25 @@ import { subNetworkWizard } from '../../wizards/subnetwork.js';
...
-  @property()
+  /** SCL element SubNetwork */
+  @property({ attribute: false })
   element!: Element;
-
+  /** SubNetwork attribute name */
   @property()
   get name(): string {
-    return this.element.getAttribute('name') ?? '';
+    return this.element.getAttribute('name') ?? 'UNDEFINED';
```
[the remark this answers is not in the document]
JakobVogelsang: I would argue not. The `name` is a required attribute, and normally it should not happen.

ca-d (approved):
LGTM

PR #494, opened 2022-01-18
feat(wizards/reportcontrol): added new IED wizard to update name/description

dlabordus:
Also all SCL elements referencing that IED Name are updated.

### DIFF
--- a/src/wizards/foundation/references.ts
+++ b/src/wizards/foundation/references.ts
@@ -16,20 +16,24 @@ const referenceInfos: Record<
     attribute: string | null;
   }[]
 > = {
-  IED: [{
-      elementQuery: `ConnectedAP`,
-      attribute: 'iedName'
-    }, {
-      elementQuery: `Server > Association`,
+  IED:
+    [{
+      elementQuery: `Association`,
       attribute: 'iedName'
     }, {
       elementQuery: `ClientLN`,
       attribute: 'iedName'
     }, {
-      elementQuery: `LNode`,
+      elementQuery: `ConnectedAP`,
       attribute: 'iedName'
     }, {
-      elementQuery: `Inputs > ExtRef`,
+      elementQuery: `ExtRef`,
+      attribute: 'iedName'
+    }, {
+      elementQuery: `KDC`,
+      attribute: 'iedName'
+    }, {
+      elementQuery: `LNode`,
       attribute: 'iedName'
     }, {
       elementQuery: `GSEControl > IEDName`,
--- a/src/wizards/foundation/references.ts
+++ b/src/wizards/foundation/references.ts
@@ -1,4 +1,4 @@
-import {EditorAction, isPublic} from "../../foundation.js";
+import {isPublic, SimpleAction} from "../../foundation.js";
 
 const referenceInfoTags = ['IED'] as const;
 type ReferencesInfoTag = typeof referenceInfoTags[number];
@@ -60,7 +60,7 @@ function cloneElementAndTextContent(element: Element, value: string | null): Ele
   return newElement;
 }
 
-export function updateReferences(element: Element, oldValue: string | null, newValue: string): EditorAction[] {
+export function updateReferences(element: Element, oldValue: string | null, newValue: string): SimpleAction[] {
   if (oldValue === newValue) {
     return [];
   }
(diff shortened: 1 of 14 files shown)


src/wizards/ied.ts
```diff
@@ -0,0 +1,76 @@
...
+  desc: string | null,
+  reservedNames: string[]
+): TemplateResult[] {
+  return [
+    html`<wizard-textfield
+      label="name"
+      .maybeValue=${name}
+      helper="${translate('ied.wizard.nameHelper')}"
+      required
+      validationMessage="${translate('textfield.required')}"
```
JakobVogelsang, line 37: I would suggest adding a pattern here. I think it is defined in the issue. For the average user, it might not be so clear what the limitation to this attribute might be.

src/wizards/ied.ts
```diff
@@ -0,0 +1,76 @@
...
+    ></wizard-textfield>`,
+  ];
+}
+
+export function reservedNamesIED(currentElement: Element): string[] {
+  return Array.from(
+    currentElement.parentNode!.querySelectorAll('IED')
+  )
+    .filter(isPublic)
+    .map(cNode => cNode.getAttribute('name') ?? '')
```
JakobVogelsang, line 55: Why `cNode`? This jumped into my eye. So far, I can see something like `ied` would fit better.

src/wizards/foundation/references.ts
```diff
@@ -0,0 +1,93 @@
...
+  const referenceInfo = referenceInfos[<ReferencesInfoTag>element.tagName];
+  if (referenceInfo === undefined) {
+    return [];
+  }
+
+  const actions: EditorAction[] = [];
+  referenceInfo.forEach(info => {
+    if (info.attribute !== null) {
+      Array.from(element.ownerDocument.querySelectorAll(`${info.elementQuery}[${info.attribute}="${oldValue}"]`))
+        .filter(element => isPublic(element))
```
JakobVogelsang, line 77: `element => isPublic(element)` can be shortened down to `isPublic`. This is very picky and not necessary to change. Your solution is working as well.

src/wizards/foundation/references.ts
```diff
@@ -0,0 +1,93 @@
...
+        .forEach(element => {
+          const newElement = cloneElement(element, info.attribute!, newValue);
+          actions.push({old: {element}, new: {element: newElement}});
+        })
+    } else {
+      Array.from(element.ownerDocument.querySelectorAll(`${info.elementQuery}`))
+        .filter(element => element.textContent === oldValue)
+        .filter(element => isPublic(element))
```
JakobVogelsang, line 85: The same hint here too.

JakobVogelsang (changes requested):
In general, look good. Two things I would says might be better:
1. complexAction instead of simpleActions
2. Add pattern to the wizard-textfield. At the moment for example you can type in things like `None` which is not schema valid

src/wizards/foundation/references.ts
```diff
@@ -0,0 +1,93 @@
...
+    }, {
+      elementQuery: `GSEControl > IEDName`,
+      attribute: null
+    }, {
+      elementQuery: `SampledValueControl > IEDName`,
+      attribute: null
+    }]
+}
+
+function cloneElement(element: Element, attributeName: string, value: string | null): Element {
```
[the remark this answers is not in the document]
dlabordus: The record solution in foundation is not working, because the name can't be passed as variable. The name of the variable 'attributeName' is then used as the name of the attribute in the XML. 
The way cloneElement is used in the other solutions the name of the variable is the name of the attribute.

src/wizards/ied.ts
dlabordus: Solved, copy / paste issue probably :-)

src/wizards/ied.ts
```diff
@@ -0,0 +1,76 @@
...
+    if ( name === oldName &&
+         desc === element.getAttribute('desc')) {
+      return [];
+    }
+
+    const newElement = cloneElement(element, { name, desc });
+    const actions: EditorAction[] = [];
+    actions.push({ old: { element }, new: { element: newElement } });
+    actions.push(...updateReferences(element, oldName, name));
+    return actions;
```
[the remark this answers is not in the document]
dlabordus: I couldn't find ComplexAction in ImportIEDs, but found a example in connectedap.ts in wizards.
Is this also a good example?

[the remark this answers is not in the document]
dlabordus: Fixed it using example from connectedap.ts :-)

src/wizards/ied.ts
dlabordus: Done. Also added pattern to descr field.

JakobVogelsang (approved):
LGTM

PR #495, opened 2022-01-18
fix(editors/SingleLineDiagram): added check if a substation is available/selected before drawing

dlabordus:


JakobVogelsang (approved):
LGTM

PR #501, opened 2022-01-19
fix(editors): In IED and SLD Editors fixed preserving the selection (IED or Substation)

dlabordus:
If a IED or Substation was selected in the drop-down box and next a Element was edited using one of the Wizards the selection was lost and the first element of the list was again selected.

The solution for now is to save the selection outside the LitElement, preventing it being lost after being re-created.

### DIFF
--- a/src/editors/IED.ts
+++ b/src/editors/IED.ts
@@ -28,8 +28,10 @@ export default class IedPlugin extends LitElement {
   doc!: XMLDocument;
 
   private get alphabeticOrderedIeds() : Element[] {
-    return Array.from(this.doc?.querySelectorAll(':root > IED'))
-                .sort((a,b) => compareNames(a,b));
+    return (this.doc)
+      ? Array.from(this.doc.querySelectorAll(':root > IED'))
+             .sort((a,b) => compareNames(a,b))
+      : [];
   }
 
   @state()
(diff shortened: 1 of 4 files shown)


src/editors/SingleLineDiagram.ts
```diff
@@ -45,6 +52,16 @@ import '@material/mwc-list/mwc-list-item';
 import '@material/mwc-select';
 import '@material/mwc-textfield';
 
+/*
+ * We need a variable outside the plugin to save the selected substation, because the Plugin is created
+ * more than once during working with the SLD, for instance when opening a Wizard to edit equipment.
+ */
+let sldEditorSelectedSubstationName: string | undefined;
+function onOpenDocResetSelectedSubstation() {
```
JakobVogelsang, line 60: Same here a short comment why this function is here? Both a comment. If you think this is obvious, leave it out.

JakobVogelsang (approved):
LGTM. The two comments are completely optional

src/editors/IED.ts
```diff
@@ -8,65 +8,82 @@ import '../zeroline-pane.js';
...
-import { Select } from '@material/mwc-select';
 import { SingleSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
 import { compareNames, getDescriptionAttribute, getNameAttribute } from '../foundation.js';
 
+/*
+ * We need a variable outside the plugin to save the selected IED, because the Plugin is created
+ * more than once during working with the IED Editor, for instance when opening a Wizard to edit the IED.
+ */
+let iedEditorSelectedIedName: string | undefined;
+function onOpenDocResetSelectedIed() {
```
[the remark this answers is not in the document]
dlabordus: Added some comments, is indeed nicer.

src/editors/SingleLineDiagram.ts
dlabordus: Added some comments, is indeed nicer.

PR #502, opened 2022-01-19
feat(settings): load nsdoc to local storage

Flurb:


JakobVogelsang:
@Flurb I had a first look on your PR and code. I need another round to go through your code, but have two things that I would remark:
1. I would love to have `multi` file import. I think some will want to load all files at once, not one after another.
2. I can imagine that list-item's being `noninteractive` - and maybe even `disabled` - so long there is no file loaded would make it clearer for the user.

Flurb:
> @Flurb I had a first look on your PR and code. I need another round to go through your code, but have two things that I would remark:
> 
> 1. I would love to have `multi` file import. I think some will want to load all files at once, not one after another.
> 2. I can imagine that list-item's being `noninteractive` - and maybe even `disabled` - so long there is no file loaded would make it clearer for the user.

Implemented both!

src/Setting.ts
```diff
@@ -13,21 +13,67 @@ import { Switch } from '@material/mwc-switch';
 
 import { ifImplemented, LitElementConstructor, Mixin } from './foundation.js';
 import { Language, languages, loader } from './translations/loader.js';
-import { WizardDialog } from './wizard-dialog.js';
 
-export type Settings = {
+import './Divider.js';
```
JakobVogelsang, line 17: We usuaelly add our imported web-components above our class, function imports

src/Divider.ts
```diff
@@ -0,0 +1,22 @@
+import {css, customElement, html, LitElement, TemplateResult} from "lit-element";
```
JakobVogelsang, line 1: The files name does not fit to the way other `web-components` are defined. We usuaelly do use the same name as the custem tag. Here it would be `openscd-devicer`. Also but I cannot say that looking at the code purelly if this is desinged to be used within dialogs/wizards. Then we should call it i `wizard-devider`

src/Setting.ts
```diff
@@ -90,6 +120,72 @@ export function Setting<TBase extends LitElementConstructor>(Base: TBase) {
...
+      
+      if (nsdSetting) {
+        const nsdoc = this.parseToXmlObject(nsdSetting)!.querySelector('NSDoc');
+        nsdVersion = nsdoc?.getAttribute('version');
+        nsdRevision = nsdoc?.getAttribute('revision');
+      }
+
+      return html`<mwc-list-item id=${key} graphic="avatar" hasMeta twoline .disabled=${!nsdSetting}>
+        <span>${key}</span>
+        ${nsdSetting ? html`<span slot="secondary">${nsdVersion}${nsdRevision}</span>` :
```
JakobVogelsang, line 173: Is release missing? For the experienced user there is a bit difference between `2007B` and `2007B4`.

src/translations/de.ts
```diff
@@ -42,6 +42,8 @@ export const de: Translations = {
     dark: 'Dunkles Design',
     mode: 'Profimodus',
     showieds: 'Zeige IEDs im Substation-Editor',
+    selectFileButton: '???',
```
JakobVogelsang, line 45: `Datei auswählen` would be german term here

src/Setting.ts
Flurb: Where should it be placed then?
I thought it should be added on the bottom of all imports :)

src/Divider.ts
Flurb: WizardDivider would be more appropriate here :)

src/Setting.ts
Flurb: Added!

src/translations/de.ts
Flurb: Done!

src/editors/IED.ts
```diff
@@ -31,6 +32,13 @@ export default class IedPlugin extends LitElement {
   @property()
   doc!: XMLDocument;
 
+  /** All the nsdoc files that are being uploaded via the settings. */
+  @property()
+  nsdoc!: Nsdoc;
```
dlabordus, line 37: Where is this used? Shouldn't it be passed to the ied-container?

src/foundation.ts
```diff
@@ -8,6 +8,7 @@ import AceEditor from 'ace-custom-element';
 
 import { WizardTextField } from './wizard-textfield.js';
 import { WizardSelect } from './wizard-select.js';
+import { iec6185074 } from './validators/templates/foundation.js';
```
dlabordus, line 11: Is this import used? No other code has changed.

src/editors/IED.ts
```diff
@@ -31,6 +32,13 @@ export default class IedPlugin extends LitElement {
...
+
+  /** Query holding the current selected IEDs. */
```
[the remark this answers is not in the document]
Flurb: Probably during merge, good call!

test/unit/Setting.test.ts
```diff
@@ -49,6 +48,44 @@ describe('SettingElement', () => {
...
+  
+    const nsdocFile = await fetch('/test/testfiles/settingTest.nsdoc')
+      .then(response => response.text())
+
+    element.setSetting('IEC 61850-7-2', nsdocFile);
+    
+    await element.requestUpdate();
+    await element.updateComplete;
+
+    expect(element).shadowDom.to.equalSnapshot();
```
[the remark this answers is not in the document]
Flurb: No doing it like below was intended.
Checking if the file in localStorage is what we expected.

Added it!

src/Setting.ts
```diff
@@ -90,6 +118,71 @@ export function Setting<TBase extends LitElementConstructor>(Base: TBase) {
...
+    private async loadNsdocFile(evt: Event): Promise<void> {
+      const files = Array.from(
+        (<HTMLInputElement | null>evt.target)?.files ?? []
+      );
+      
+      if (files.length == 0) return;
+      files.forEach(async file => {
+        const text = await file.text();
+        const id = this.parseToXmlObject(text).querySelector('NSDoc')?.getAttribute('id');
+        if (!id) return;
```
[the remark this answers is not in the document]
Flurb: Added!

src/foundation/nsdoc.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export async function initializeNsdoc(): Promise<Nsdoc> {
+  const nsdoc72 = localStorage.getItem('IEC 61850-7-2') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-7-2')!, 'application/xml') : undefined;
+  const nsdoc73 = localStorage.getItem('IEC 61850-7-3') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-7-3')!, 'application/xml') : undefined;
+  const nsdoc74 = localStorage.getItem('IEC 61850-7-4') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-7-4')!, 'application/xml') : undefined;
+  const nsdoc81 = localStorage.getItem('IEC 61850-8-1') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-8-1')!, 'application/xml') : undefined;
+
+  const nsd74 = await iec6185074;
+
+  const referenceInfoTags = ['LN', 'LN0'] as const;
```
dlabordus, line 23: referenceInfoTags is maybe not the best name :-)
I think a copy from original.

dlabordus (changes requested):

src/foundation/nsdoc.ts
Flurb: What about just calling it `iedElementTagNames`?

JakobVogelsang (approved):
LGTM. Some changes to the NSDoc object that I think are necessary can be done in the pending PR #522

dlabordus (approved):

Issue #407, opened 2021-12-01
Make a VoltageLevel more visible in SLD editor


Flurb:
As a CoMPAS user
I want to see what's a VoltagLevel in the SLD editor
So I can search quicker within the SLD for certain bays for example.

Only valuable if end users need this!

Sander3003:
@Flurb is this still an issue?

Flurb:
@Sander3003 We haven't implemented it yet, so it's still something to do :)
If it's still an issue, that's another question.

What we wanted to apply, is the same principle as selecting a Bay in the SLD Editor like this:

![Schermafbeelding 2022-03-18 om 15 21 47](https://user-images.githubusercontent.com/16406106/159020430-9c955423-49e3-41f5-b44b-03d91996aefe.png)

Issue #406, opened 2021-12-01
Make a Bay more visible in SLD editor


Flurb:
As a CoMPAS user
I want to see what's a Bay in the SLD editor
So I can search quicker within the SLD for certain equipments for example.

Only valuable if end users need this!

PR #504, opened 2022-01-20
feat(editors/SingleLineDiagram) Allow edit/move of bays

dlabordus:
Based on the idea from @danyill created this implementation.

When hovering over the Bay there will be a dotted line around it and the label is shown (with a Edit icon).
Clicking on the Label (or Edit icon) will open the Edit Bay window, so the user can update the X and Y Coordinates.

Also the Busbar is changed a little bit. The editing is also only possible by clicking on the Label (or Edit button)
to change the X and Y Coordinates.


JakobVogelsang:
I had not a closer look on it from code perspective. Just ran some test with files I had. The basic idea is nice. At least on my machine it work ok with busb bars. With other bays it is very tricky to get to the corner without the focus disapearing. I am not sure where this is comming from. Maybe my file has something very special inside that is causing this issue.

Sander3003:
This looks related to #407 #406 #456.

Flurb (approved):

Issue #475, opened 2022-01-12
Incorrect abbreviation for the IED toggle button in the Substation editor


JakobVogelsang:
The abbreviation is wrong and the same as for the button next to it.

PR #505, opened 2022-01-20
feat(wizards/reportcontrol): allow basic ReportControl manipulation capability

### LINKED ISSUE #506: Incorrect rendering of FCDA elements in DataSet
There is a difference when rendering a DataSet located in a LN0 and one located in LN. The first shows the path from logical node to the data the second shows the logical device name. My guess is that the identity string differs for the two.

JakobVogelsang:
Also closes #506 #475


JakobVogelsang:
@danyill @Sander3003 @randy993 we have added something we call dynamic sub-wizards. This allows you to navigate more fluently between wizards that were open within other wizards. You can test it here with the ReportControl wizards that we will add very soon. What do you think about this UI?

dlabordus (approved):

Sander3003:
The reporting block shows the wrong tooltip (GOOSE instead of Reporting):
image

JakobVogelsang:
Shoot :( I will change that.

Sander3003:
image
We would suggest to change the edit button into a "ok" button.

Sander3003:
The UI goods look to me!

JakobVogelsang:
How about `save` as in the edit wizard for the control block itself?

Sander3003:
Save could work as well as.

JakobVogelsang:
> Save could work as well as.

I will put it this PR #507  as this wizard is commonly used.

PR #510, opened 2022-01-23
feat(wizards/sampledvaluecontrol): add edit wizards accessable from selection

### LINKED ISSUE #428: Add wizard to edit SampleValueControl attributes
As a system configuration engineer, I want to edit attributes of the SCL element `SampledValueControl`. 

**Requirement**:
- shall be accessible from the select wizard for SampledValueControl (#426 )
- shall be accessible from the Substation editor type plugin
- shall be accessible from the IED editor type plugin
- the attribute `confRev` shall not be editable but shall be updated automatically +10000 each time an update action of this control block or its child elements is triggered

**Type description**: 
attribute | description |  
 --- | --- |
name | the name of the sampled value control block - must be unique within LN0 |   
desc | description |    
smvID | id for the sampled value control block |
multicast | whether the control block is of type multicast false is deprecated  |
smpRate | sample rate its interpretation depends on smpMode | 
nofASDU | number of samples per (Ethernet) packet | 
smpMode | the mode of the attribute smpRate  | 
securityEnable |  security status of the sampled value control block | 
confRev | READ_ONLY attribute updated by OpenSCD automatic |

**Restriction** :
- 

attribute | required |  pattern  | min length  | max length |  
 --- | --- | --- | --- | --- | 
name | yes | `tCBName  ` | 1  |  32 |   
desc | no | xs:normalizedString | 0 |    |    
smvID | yes | tMessageID | 1  | 129 |
multicast | yes | xs:boolean |   |    |
smpRate | yes  | xs:unsignedInt | 0 |  | 
nofASDU | yes  | xs:unsignedInt  | 0 |  | 
smpMode | no | `tSmpMod` |  |  | 
securityEnable |  no | `tPredefinedTypeOfSecurityEnum`  |  |  | 

**Type definition**:
type  |  pattern/value range  | min length  | max length |  
--- | --- | --- | --- | 
tCBName | `[A-Za-z][0-9A-Za-z_]*` | 1  |  32 |
tMessageID | `[&#x0020;-&#x007E;]*` | 1  |  60 |   
tSmpMod | SmpPerPeriod, SmpPerSec, SecPerSmp |  |  |
tPredefinedTypeOfSecurityEnum | None, Signature, SignatureAndEncryption  |  |  |

**Further Information Link**
- 

**Edition 1 vs Edition 2**
- the restrictions are valid for Ed2.1 and can differ for Ed1 and Ed2. This shall not be taken into account for not

JakobVogelsang:
Close #428


src/wizards/sampledvaluecontrol.ts
```diff
@@ -1,7 +1,187 @@
...
+      validationMessage="${translate('textfield.required')}"
+      pattern="${patterns.asciName}"
+      maxLength="${maxLength.cbName}"
+      dialogInitialFocus
+    ></wizard-textfield>`,
+    html`<wizard-textfield
+      label="desc"
+      .maybeValue=${options.desc}
+      nullable
+      helper="${translate('scl.desc')}"
```
dlabordus, line 56: Some description field have a pattern and some not. Should there always be a pattern, because all description are the same type?

src/zeroline-pane.ts
```diff
@@ -78,8 +78,11 @@ export class ZerolinePane extends LitElement {
   }
 
   openSampledValueControlSelection(): void {
-    const wizard = selectSampledValueControlWizard(this.doc.documentElement);
-    if (wizard) this.dispatchEvent(newWizardEvent(wizard));
+    this.dispatchEvent(
+      newSubWizardEvent(() =>
```
dlabordus, line 82: Is this not a normal Wizard? zeroline-pane.ts is used by the Substation Editor, correct?

src/zeroline/ied-editor.ts
```diff
@@ -51,8 +51,9 @@ export class IedEditor extends LitElement {
   }
 
   private openSmvControlSelection(): void {
-    const wizard = selectSampledValueControlWizard(this.element);
-    if (wizard) this.dispatchEvent(newWizardEvent(wizard));
+    this.dispatchEvent(
+      newSubWizardEvent(() => selectSampledValueControlWizard(this.element))
```
dlabordus, line 55: Same here, is this not a normal Wizard?

dlabordus (changes requested):

src/zeroline-pane.ts
JakobVogelsang: True! It is cleaner. Thanks!! In this particular case it does not make a difference, but I will adapt.

src/zeroline/ied-editor.ts
JakobVogelsang: True!

src/wizards/sampledvaluecontrol.ts
JakobVogelsang: It is true. For readablility reasos there should. For this particular attribute pattern is `xs:nomalizedString` and therefore I thinks it does not mattern if we forget to set it. Still thanks for the hint. The wizards library need some clean up anyway and this is one such point.

dlabordus (approved):

PR #512, opened 2022-01-24
fix(wizards/fcda): make sure lnInst is non empty string

### LINKED ISSUE #497: Schema validation issue on `FCDA` creation
When I create a new `FCDA` pointing to a `LLN0` the attribute `lnInst` is set to an empty string. The schema validator throws an issue saying that the empty string does not fit the predefined pattern `[0-9]{1,12}`. We need to check if this the same for all versions of the standard. And if not have to decide who to proceed.

JakobVogelsang:
Close #497 

Sorry @ca-d. I should have added a description of the PR here. With this fix I wanted to prevent that an empty string is set for the attribute `lnInst` within the element `FCDA`. To do so, the action function checks if the referenced `LN` or `LN0` has a `inst` defined AND if it is not an empty string. In these cases `lnInst` must be `inst`. In all other cases, `lnInst` must be missing within `FCDA`.

### DIFF
--- a/src/wizards/fcda.ts
+++ b/src/wizards/fcda.ts
@@ -31,7 +31,7 @@ function newFCDA(parent: Element, path: string[]): Element | undefined {
   const prefix = ln.getAttribute('prefix') ?? '';
   const lnClass = ln.getAttribute('lnClass');
   const lnInst =
-    ln.getAttribute('inst') && ln.getAttribute('inst') !== ''
+    (ln.getAttribute('inst') && ln.getAttribute('inst') !== '')
       ? ln.getAttribute('inst')
       : null;
 
--- a/src/wizards/fcda.ts
+++ b/src/wizards/fcda.ts
@@ -30,7 +30,10 @@ function newFCDA(parent: Element, path: string[]): Element | undefined {
   const ldInst = ln.closest('LDevice')?.getAttribute('inst');
   const prefix = ln.getAttribute('prefix') ?? '';
   const lnClass = ln.getAttribute('lnClass');
-  const lnInst = ln.getAttribute('inst') ?? '';
+  const lnInst =
+    ln.getAttribute('inst') && ln.getAttribute('inst') !== ''
+      ? ln.getAttribute('inst')
+      : null;
 
   let doName = '';
   let daName = '';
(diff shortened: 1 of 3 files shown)


src/wizards/fcda.ts
```diff
@@ -30,7 +30,10 @@ function newFCDA(parent: Element, path: string[]): Element | undefined {
   const ldInst = ln.closest('LDevice')?.getAttribute('inst');
   const prefix = ln.getAttribute('prefix') ?? '';
   const lnClass = ln.getAttribute('lnClass');
-  const lnInst = ln.getAttribute('inst') ?? '';
+  const lnInst =
+    ln.getAttribute('inst') && ln.getAttribute('inst') !== ''
```
ca-d, line 34: if `ln.getAttribute('inst') === ''`, then it's falsey and the `&&` will short circuit, meaning this will return `''`, which doesn't look like the intention. Is this behaviour tested correctly?

test/integration/wizards/fcda-wizarding-editing-integration.test.ts
```diff
@@ -46,14 +46,14 @@ describe('FCDA editing wizarding integration', () => {
       expect(
         doc.querySelector(
           'DataSet > FCDA[ldInst="CircuitBreaker_CB1"]' +
-            '[prefix=""][lnClass="LLN0"][lnInst=""][doName="Beh"][daName="stVal"][fc="ST"]'
+            '[prefix=""][lnClass="LLN0"]:not(lnInst)[doName="Beh"][daName="stVal"][fc="ST"]'
```
ca-d, line 49: Do we have to check for both `:not` and `=""` here in case of an `lnInst` we did not set ourselves? Also, does this have anything to do with the above change to `inst`?

ca-d (commented):
Sorry, I don't seem to understand this pull request at all, probably mostly because I don't really understand the bug this is meant to fix. Disregard my comments if they don't make sense.

src/wizards/fcda.ts
ca-d, line 34: This was really hard for me to read and understand, I got it now. Could you please add parentheses around the first two boolean operands if the linter will let you?

ca-d (commented):
Seems fine, not sure about the selector change though.

test/integration/wizards/fcda-wizarding-editing-integration.test.ts
JakobVogelsang: I would argue that we don't. The test explicitly tests that this attribute is not there, as this is what all versions of the standard require. Meaning, if we choose to reference to a `LLN0` with a `FCDA` there must not be a `lnInst`.

src/wizards/fcda.ts
JakobVogelsang: Sure :)

dlabordus (approved):

PR #516, opened 2022-01-25
feat(wizards/settings): Added Nsdoc plugin + Show LN names

Flurb:


src/editors/ied/ldevice-container.ts
```diff
@@ -48,8 +49,9 @@ export class LDeviceContainer extends LitElement {
         ></mwc-icon-button-toggle>
       </abbr>` : nothing}
       <div id="lnContainer">
-        ${this.toggleButton?.on ? lnElements.map(server => html`<ln-container
-            .element=${server}
+        ${this.toggleButton?.on ? lnElements.map(ln => html`<ln-container
+            .element=${ln}
+            .lnClassElement=${nsdoc.get74NsdLNClass(ln.getAttribute('lnClass')!)}
```
dlabordus, line 54: Why do this lookup here? We are passing the LN Element to the LNContainer.

src/Setting.ts
```diff
@@ -16,14 +16,38 @@ import { ifImplemented, LitElementConstructor, Mixin } from './foundation.js';
...
+      return nsd74.querySelector(`NS > LNClasses > LNClass[name="${name}"]`);
+    },
+
+    /**
+     * Get the documentation for a specific ID found in the IEC 61850-7-4 nsdoc file.
+     * @param id - The id to search for.
+     * @returns The documentation for the specific ID.
+     */
+    get74NsdocDocumentation(id: string): Element | null {
+      return parsedNsdoc74.querySelector(`NSDoc > Doc[id="${id}"]`);
```
dlabordus, line 45: Why not combine these two functions? Pass a LNClass Name, use that to search in NSD and next in NSDoc.

And maybe in state of adding 74 to the function name, pass some kind of constant which can be used to get the correct NSD/NSDOC variables.

dlabordus (changes requested):

JakobVogelsang:
I will merge this as there are some changes that revert changes in the merging branch. Hope this does not block you from reviewing @dlabordus

JakobVogelsang (approved):
Will be riewed in load-nsdoc-settings directly

Issue #498, opened 2022-01-19
Issue with sub-wizards in case wizards are depending on each other


JakobVogelsang:
The example - issue - we see in our solution is this:

1.) Open edit wizard for `DataSet` say from `ReportControl` element. 
2.) Change the name of the DataSet
3.) As the name of the DataSet is reference from the `ReportControl` element, we need to change the attribute `datSet` there
4.) When - in a subwizard based UI - we return to the edit wizard of the `ReportControl` it actually does not show the `ReportControl` in the SCL, but the "old" one that was replaced with the Update action. 

I would say a solution would be to not use elements as the input for the function that return `Wizard` but to use our identity string. 

For now, we have tom make sure to either not allow changing things that might affect the wizard before or to not use subwizards in such cases.

JakobVogelsang:
This is an issue with our Editing system. Specifically with our `onUpdate` function. What we do at the moment is to swap the `old` element with the `new` element. This function is very straight foreward and the undo does work in the same way, we just nieed to exchange `old` and `new`. 

Me an @ca-d as well as @dlabordus were discussing how we could solve the problem. @ca-d and me are seeing a long term solution and a potentially long term solution that I will open a new issue for. But in short:

*Potentially long-term*:
- do not use DOM objects at all for editing but objects and to identify the element in the DOM the identiy string we provide in the foundation

>Advantage: We have automatically serializable actions that we can store in the SCD - e.g. `Hitem`. 
>Disadvantage: The SCL does not have native id's for their elements and the id is a reference build from configurable attributes. This means we would need something event that triggers when a identity string changes :(

*Short-term*:
We leave the `Update` API as is `new:Element,old:Element` but do not swap in the `onUpdate` but we exchange attributes using the functions `removeAttributeNode` and `setAttributeNode`. Then we only need to change the `onUpdate` and everything else can stay as is.

>Advantages: clear reference through element, no big changes needed this can be implemented very fast
>Disadvantages: we relay with our API on DOM from the beginning, we can not undo history before file open

@dlabordus @Flurb @danyill and others what do you think about that

dlabordus:
For the short-term, how are you planning to do the undo history in this case?
Because that still means that you need to save a clone of the old element to make this possible.
And in the future it still could be possible to undo history from before opening the file, just replacing the attributes again...

JakobVogelsang:
[...]

PR #521, opened 2022-01-28
refactor(Editing): update action swaps attributes not elements

JakobVogelsang:
This is the refactor we have discussed in issue #498. 
- the main refactor is swapping attributes between old and new element of the update action
- in addition to that we need to swap the textContent as sometimes information is stored here, not in attributes. E.g. `<MinTime multiplier="m" unit="s">10</MinTime>`.
- textContent shall only be swapped if the new element has a textContent defined to avoid changing something we did not intend to change

> swapping attributes seems ugly to me as I needed to add a dummy element to store atrributes from one of the elements. But I did not see a better way to do it.

### DIFF
--- a/src/Editing.ts
+++ b/src/Editing.ts
@@ -219,8 +219,10 @@ export function Editing<TBase extends LitElementConstructor>(Base: TBase) {
       oldAttrs.forEach(attr => newElement.setAttributeNode(attr));
       newAttrs.forEach(attr => oldElement.setAttributeNode(attr));
 
+      if (!action.new.childNodes) return true;
+
       const newChildNodes = action.new.childNodes;
-      const oldChildNodes = action.old.childNodes;
+      const oldChildNodes = Array.from(action.old.element.childNodes);
 
       oldChildNodes?.forEach(child => oldElement.removeChild(child));
 
--- a/src/Editing.ts
+++ b/src/Editing.ts
@@ -210,28 +210,19 @@ export function Editing<TBase extends LitElementConstructor>(Base: TBase) {
 
       const oldElement = action.old.element;
       const newElement = action.new.element;
-      const oldAttrs = action.old.element.attributes;
-      const newAttrs = action.new.element.attributes;
+      const oldAttrs = Array.from(action.old.element.attributes);
+      const newAttrs = Array.from(action.new.element.attributes);
 
-      const intElement =
-        action.new.element.ownerDocument.createElement('Clone');
+      oldAttrs.forEach(attr => oldElement.removeAttributeNode(attr));
+      newAttrs.forEach(attr => newElement.removeAttributeNode(attr));
 
-      while (oldAttrs.length > 0)
-        intElement.setAttributeNode(
-          oldElement.removeAttributeNode(oldAttrs[0])
-        );
+      oldAttrs.forEach(attr => newElement.setAttributeNode(attr));
+      newAttrs.forEach(attr => oldElement.setAttributeNode(attr));
 
-      while (newAttrs.length > 0)
-        oldElement.setAttributeNode(
-          newElement.removeAttributeNode(newAttrs[0])
-        );
-
-      while (intElement.attributes.length > 0)
-        newElement.setAttributeNode(
-          intElement.removeAttributeNode(intElement.attributes[0])
-        );
-
-      if (action.new.element.textContent) {
+      if (
+        action.new.element.childNodes.length === 1 &&
+        action.new.element.children.length === 0
+      ) {
         const oldTextContent = action.old.element.textContent;
         action.old.element.textContent = action.new.element.textContent;
         action.new.element.textContent = oldTextContent;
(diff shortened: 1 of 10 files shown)


JakobVogelsang:
Thinking about the `textContent` part in more detail, I think we have to extend the API of `Update` with an optional `textContent` key in `old` and `new`. This will still allow us to use the invert function as is.  
```
export interface Update {
  old: { element: Element; textContent?: string | null };
  new: { element: Element; textContent?: string | null };
  derived?: boolean;
  checkValidity?: () => boolean;
}
``` 

This also would allow manipulating nonstandard content with `Update` actions.

ca-d:
Maybe try `childNodes?: Node[]` instead of `textContent?: string | null`?

JakobVogelsang:
@ca-d I have implemented the proposal using optional `childNodes` in the update action. I think this could be a way to go! It is straight forward, and we know it will work with the invert function. 

There is only one drawback, that I see: Defining the action is a bit bumpy. At least it feels for me. If you have time have a look at the last commit and especially the way the update action is defined e.g. for the `Voltage` element.

src/Editing.ts
```diff
@@ -208,8 +208,28 @@ export function Editing<TBase extends LitElementConstructor>(Base: TBase) {
...
+
+      oldAttrs.forEach(attr => oldElement.removeAttributeNode(attr));
+      newAttrs.forEach(attr => newElement.removeAttributeNode(attr));
+
+      oldAttrs.forEach(attr => newElement.setAttributeNode(attr));
+      newAttrs.forEach(attr => oldElement.setAttributeNode(attr));
+
+      const newChildNodes = action.new.childNodes;
+      const oldChildNodes = action.old.childNodes;
+
```
dlabordus, line 224: Will swapping the ChildNodes only work for TextNodes?
Looking at the places where ChildNodes are passed, there are only TextNodes.

Because you can pass a array of Nodes it's a little bit confusing for me.
Shouldn't be better to be able to just pass a TextNode and also only remove these.
Maybe then you even don't have to pass any extra parameters, just search for the text nodes and swap these.

JakobVogelsang: This peace of code wil swap all child `Node`s. So far I understood Node can be everything. Element for example is a Node as well. So far we only use text Node's but one might swap all children using this API. The reason we only swap text nodes is the definition of those elements like 

```
<Voltage unit="V" multiplier="k">110.1</Voltage>
```

where text nde is the only allowed child.

dlabordus: Shouldn't the swap than work the same as with Attributes?
First remove all child nodes from **both** old and new element. 
This is now only done on the oldElement.

And as Christian also mentions above, is this working for more Complex structures?

src/Editing.ts
```diff
@@ -208,8 +208,28 @@ export function Editing<TBase extends LitElementConstructor>(Base: TBase) {
...
+      const oldAttrs = Array.from(action.old.element.attributes);
+      const newAttrs = Array.from(action.new.element.attributes);
+
+      oldAttrs.forEach(attr => oldElement.removeAttributeNode(attr));
+      newAttrs.forEach(attr => newElement.removeAttributeNode(attr));
+
+      oldAttrs.forEach(attr => newElement.setAttributeNode(attr));
+      newAttrs.forEach(attr => oldElement.setAttributeNode(attr));
+
+      const newChildNodes = action.new.childNodes;
```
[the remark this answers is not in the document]
JakobVogelsang: Hmmm good point. As I did it in this PR `action.old.childNodes` have been created using `Array.from()`. This is why it works, but making sure that childNode's are not a live NodeList in the `OnUpdate´ is not a bad idea nor to have specific tests for that. Thanks

src/Editing.ts
JakobVogelsang: You are correct I it should follow the same approach as the attributes. Thanks for the hint!

Manipulating more cmplex structure is possible I think with the downside AND advantage that the behaviour can be manipulated by the user with the childNodes attribute in the `Update` action. That means you as a user are very flexible in what you want to add/remove from the old element, but it is easy to make mistakes. Chris am me were discussing that as well. We both favour a solution that is not bound to the SCL schema but allows any kind of update in the DOM.

src/Editing.ts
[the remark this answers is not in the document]
ca-d: Actually, I just misread `action.new.childNodes` as `action.new.element.childNodes`. Please disgregard my comment, the code is perfect as is.

src/Editing.ts
JakobVogelsang: @dlabordus please have a look at the tests. I added more detail to it and this gives you a chances to see how this is working at the moment.

src/foundation.ts
```diff
@@ -39,8 +39,8 @@ export interface Move {
 }
 /** Replaces `old.element` with `new.element`, keeping element children. */
```
ca-d, line 40: We currently still allow the user to give us `old.childNodes` in the action, even though that won't do anything. Maybe we should clarify that in this comment?

ca-d (approved):
LGTM, apart from one minor comment on documentation and one on style.

JakobVogelsang:
We decidet to not mix and match update of elements attributes with swapping childNodes. Instead we will add an EXTRA `EditorAction` to just update the attributes and will allow to delete and create `Node`s not only Elements

JakobVogelsang:
All the reset will be done in other PRs

PR #522, opened 2022-01-31
Display ied data attributes

Flurb:


src/foundation/nsdoc.ts
```diff
@@ -35,35 +35,133 @@ export async function initializeNsdoc(): Promise<Nsdoc> {
...
+      label: getNsdocDocumentation(nsdoc74!, lnClass?.getAttribute('titleID')) ?? lnClassAttribute
     };
   }
 
+  /**
+   * Getting data descriptions for DO(I) elements out of the IEC 61850-7-4 .nsdoc file.
+   * @param elements - The elements to use for searching the DO description.
+   * @returns Documentation from the .nsdoc file for this DO(I) file, or the name attribute in case no description can be found.
+   */
+  function getDODataDescription(elements: Element[]): { label: string; } {
```
JakobVogelsang, line 78: I would argue that we don't need to have the parent element forwarded to `getDescription`. I would make the API cleaner and get the parent element from the element itsself like that for example:
const parent = element.closest('LN0') ? element.closest('LN0') : element.closest('LN');

src/foundation/nsdoc.ts
```diff
@@ -1,28 +1,28 @@
...
+    localStorage.getItem('IEC 61850-7-3') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-7-3')!, 'application/xml') : undefined,
+    localStorage.getItem('IEC 61850-7-4') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-7-4')!, 'application/xml') : undefined,
+    localStorage.getItem('IEC 61850-8-1') ? new DOMParser().parseFromString(localStorage.getItem('IEC 61850-8-1')!, 'application/xml') : undefined
+  ]
 
-  const iedElementTagNames = ['LN', 'LN0'] as const;
+  const iedElementTagNames = ['LN', 'LN0', 'DO', 'SDO', 'DOI', 'DA', 'BDA', 'DAI'] as const;
   type IEDElementTagNames = typeof iedElementTagNames[number];
-  type GetDataDescription = (element: Element) => { label: string; };
+  type GetDataDescription = (elements: Element[]) => { label: string; };
```
JakobVogelsang, line 25: I would love to clean the API a bit more. I see why you want to have the parent in it as well. With nested BDAs it is tricky to navigate through the `NSD` files. But it is possible and if you as me the task of the individual get..Description and not of the `da-container` to save that. 

Looking through the code, I also would love to have an optional `attribute` in there. For now, it is clear which attribute we want need to look for each element, but this is not always the case. In addition to that we need to distinguish between attribute and textContent for Enums anyway.

I also think we need to have two outputs, one `short` and one `long`. For data objects it is clear that there are two types of information. The same is true for data attributes. You must not need it in the PR but I would already specify it for later use.

src/editors/ied/da-container.ts
```diff
@@ -94,7 +107,9 @@ export class DAContainer extends LitElement {
       <h6>${this.renderValue()}</h6>
       ${this.toggleButton?.on && bType == 'Struct' ? this.getBDAElements().map(element =>
         html`<da-container
-          .element=${element}>
+          .element=${element}
+          .nsdoc=${this.nsdoc}
+          .daParent=${this.daParent ?? this.element}>
```
JakobVogelsang, line 112: This should ne be saved here but can be done by clever navigation in the `nsd` files.

src/foundation/nsdoc.ts
```diff
@@ -35,35 +35,133 @@ export async function initializeNsdoc(): Promise<Nsdoc> {
...
       getDataDescription: getLNDataDescription
+    },
+    DO: {
+      getDataDescription: getDODataDescription
+    },
+    SDO: {
+      getDataDescription: getDODataDescription
+    },
+    DOI: {
+      getDataDescription: getDODataDescription
```
JakobVogelsang, line 46: This need an extra function. I see it with the logical node MMXU and the DO `PhV` and it's `SDO` phsA. This is not property taken from the nsdoc. I would suggest something like:
```
const elementsName = element.getAttribute('name')!;
const cdc =  element.parentElement?.getAttribute('cdc');
const long =  nsd73.querySelector(`CDCs > CDC[name="${cdc}"] > SubDataObject[name="${elementsName}"]`).getAttribute...
...
```

src/foundation/nsdoc.ts
```diff
@@ -35,35 +35,133 @@ export async function initializeNsdoc(): Promise<Nsdoc> {
...
+  }
+
+  /**
+   * Getting data descriptions for BDA elements out of the IEC 61850-7-3 and IEC 61850-8-1 .nsdoc file.
+   * @param elements - The elements to use for searching the BDA description.
+   * @returns Documentation from the .nsdoc file for this BDA element, or the name attribute in case no description can be found.
+   */
+  function getBDADataDescription(elements: Element[]): { label: string; } {
+    const bdaElementName = elements[0].getAttribute('name')!;
+    const bdaParent = elements[1];
```
JakobVogelsang, line 121: Can be solved with a recursive function that works something like:

1.) get the parent and its `id`
2.) element.ownerDocument.querySelector(`DOType > DA[type="${id}"]`) 
3.) found something ... return the DA
4.) did not found anything try element.ownerDocument.querySelector(`DAType > BDA[type="${id}"]`)
5.) found something still BDA
6.) did not found anything cannot find parent...


I know that this does not give back the actual parent. But for navigation within the NSD we don't need that.

JakobVogelsang (changes requested):
Mostly ok. Works already good for most data. Some still don't work with e.g `SDO` and some nested `BDA`.

src/foundation/nsdoc.ts
Flurb: The only function where we need the parent is for BDA's, because we need to have the `cdc` attribute in the `ServiceConstructedAttribute` in the IEC 61850-8-1. This cannot be resolved with `.closest`, because none of the parents have a `cdc` attribute.

Flurb: But please correct me if I'm wrong, because it would be prettier if we only need 1 parameter :D

src/foundation/nsdoc.ts
Flurb: The attribute parameter can be added later when we need it. For now, it's fine. :)

Flurb: And I love to hear your solution for using a single element instead of an array.

src/editors/ied/da-container.ts
Flurb: I'd love to see how you fix `getBDADataDescription` then, because we need the parent `DOType` for the `cdc` attribute :)

src/foundation/nsdoc.ts
Flurb: This is right, I adjusted it!

JakobVogelsang:
Hi @Flurb, after playing around a bit I found a solution that gets one valid ancestor combination with the DA as last element:

```
/**
   * @param element - BDA/DA element in the DataTypeTemplate
   * @returns one valid array ancestor combination 
   */
  function daPath(element: Element): Element[] {
    if (element.tagName === 'BDA') {
      const id = element.parentElement?.getAttribute('id');
      const reference = element.ownerDocument.querySelector<Element>(
        `DataTypeTemplates > DAType > BDA[type="${id}"], DataTypeTemplates > DOType > DA[type="${id}"]`
      );

      return reference ? [element, ...daPath(reference)] : [element];
    }

    if (element.tagName === 'DA') return [element];

    return [];
  }
```


Something like `getPath(element).pop()` called to line `137` of `nsdoc.ts` would deliver the same results as what you have now. 

HOWEVER, I do not like the solution either. So far I could see we don't need the actual ancestor to navigate in the nsd-files, but still I does not feel good. 

I would therefor propose to add an optional `path?: Element[]` or `ancestors?: Element[]`. This is extendable but more readable compared to elements:Element[]. The second reason why I think this might be interesting is that we might need the path or ancestor array in the IED editor for other features as well.

Flurb:
> ancestors?: Element[]

Having a second optional parameter `ancestors` is fine for me, also because it's optional and don't have to be added to the other `getDataDescription` functions. For me, this solution is pretty clean and should work. I will implement it :)

JakobVogelsang:
The code looks good. There is still a missing description with nested `BDA`'s. An example is  MMXU -> A -> phsA -> instCVal -> mag -> f. The last `f` is not reached with the current implementation of the getBDADescription function.

Flurb:
> The code looks good. There is still a missing description with nested `BDA`'s. An example is MMXU -> A -> phsA -> instCVal -> mag -> f. The last `f` is not reached with the current implementation of the getBDADescription function.

Fixed!

JakobVogelsang (approved):
LGTM

PR #527, opened 2022-02-02
style: make landing buttons reactive

ca-d:
The fact that the landing buttons only look nice if I have exactly two of them has been bothering me for a long time.

This patch finally fixes that by making the display of the landing buttons reactive.


JakobVogelsang (approved):
LGTM

PR #528, opened 2022-02-02
style(action-pane): clip heading label

### LINKED ISSUE #526: Header label in `action-pane` is not clipped
This is due to the CSS rule `overflow: visible`. We should change that to `overflow: clip visible`.

JakobVogelsang:
Close #526


ca-d (approved):
LGTM, thanks!

PR #536, opened 2022-02-08
feat(wizard/sampledvaluecontrol): allow removing incl. referenced elements

### LINKED ISSUE #431: Add remove button to edit wizard for SampledValueControl
As a system configuration engineer, I want to remove SampledValueControl element from the project. 

**Requirements**
- on removing check if the connected DataSet element is used within this control block only. If so remove the data set as well.
- on removing remove the connected SMV element as well.
- use complex action as undo will add DataSet and SMV back again

JakobVogelsang:
Closes #431

### DIFF
--- a/src/wizards/sampledvaluecontrol.ts
+++ b/src/wizards/sampledvaluecontrol.ts
@@ -11,11 +11,14 @@ import '../wizard-select.js';
 import '../wizard-textfield.js';
 import {
   cloneElement,
+  Delete,
   EditorAction,
   getValue,
   identity,
   isPublic,
+  newActionEvent,
   newSubWizardEvent,
+  newWizardEvent,
   selector,
   Wizard,
   WizardActor,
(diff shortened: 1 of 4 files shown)


src/wizards/sampledvaluecontrol.ts
```diff
@@ -211,6 +260,17 @@ export function editSampledValueControlWizard(element: Element): Wizard {
               }}}"
             ></mwc-button>`
           : html``,
+        html`<mwc-button
+          label="${translate('remove')}"
+          icon="delete"
+          @click=${(e: MouseEvent) => {
+            const deleteActions = removeSampledValueControlAction(element);
+            deleteActions.forEach(deleteAction =>
```
dlabordus, line 268: Shouldn't this be a Complex Action to make the Redo work as 1 unit?

dlabordus (changes requested):

src/wizards/sampledvaluecontrol.ts
JakobVogelsang: Yeah I was thinking that too yesterday. I deceded to not do it, because we don't have a diff view right now in the log and complex actions a bit hide what has changed in detail. 

In any case I would love to do it for all control blocks or non of them. Report and GOOSE control block removal at the moment works with up to 3 SimpleActions as well. 

I will open a [Issue](https://github.com/openscd/open-scd/issues/540) for it to not forget this. Is this a way to proceed for you?

dlabordus (approved):

PR #537, opened 2022-02-08
feat(wizard-checkbox): web component for xs:boolean XML attributes

JakobVogelsang:
This was on my personal machine for some time now...until now we use `wizard-select` for xs:boolean type attributes. This is not the best UI and takes up too much space in my opinion. A good example is `OptFields`, a `ReportControl` subwizard, where all inputs are xs:boolean type. 

I did not want to have everything refactored right away, and applied this web component only in the `GSEControl` edit wizard for the attribute `fixedOffs`. I would be happy about review and comments on this component :).

### DIFF
--- a/src/editors/templates/foundation.ts
+++ b/src/editors/templates/foundation.ts
@@ -105,19 +105,6 @@ export function addReferencedDataTypes(
   return actions;
 }
 
-export function buildListFromStringArray(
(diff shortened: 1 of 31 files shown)


dlabordus (approved):

ca-d (approved):
LGTM

PR #538, opened 2022-02-08
refactor(Plugins): improve extentions UI

### LINKED ISSUE #533: Confusing Extensions userinterface
As a user
It is confusing to me when to press the add button
I now think that once I selected an extions I need to click the add button

image

Suggested solution, add a oke button for exiting the window and rename the add button to "add external plugin"

Background: this is what I saw multiple users making this mistake.

JakobVogelsang:
Based on the user feedback in #533 I have adapted the UI slightly. @Sander3003 what do you think about this?

Close #533


Sander3003 (approved):
Works well!

PR #541, opened 2022-02-09
feat(editors/IED): Handle difference in version before uploading nsdoc file

### LINKED ISSUE #513: Handle difference in version before uploading nsdoc file
As a user
I want to handle the version difference between nsd and nsdoc before uploading the nsdoc file
So I'm sure the functionality works fine

Flurb:
Resolves #513


src/foundation/nsdoc.ts
```diff
@@ -1,7 +1,6 @@
 import { iec6185072, iec6185073, iec6185074, iec6185081 } from "../validators/templates/foundation.js";
 
 export interface Nsdoc {
-  nsdoc72?: XMLDocument;
```
dlabordus, line 4: Why is this only removed from the interface and not completely from the settings if it's not used?
Maybe add this again for future uses...

Flurb: It should be added, not sure why it's deleted :)

Flurb: I should be more careful before committing I guess!

dlabordus (approved):

PR #544, opened 2022-02-14
feat(wizards/reportcontrol): add create wizard

JakobVogelsang:
Finally I can show you my attemed for this very important feature. The tricky part for me was the selction of the parent for the ReportControl element. My solution is a logical node pickert that allows you to select the parent `AnyLN`. When beeing on the IED already I decided to just put it into the first `LN0`. I would love to hear your review. 

I did not finish the tests yet, so until now just the usablitily would interest me. @danyill , @randy993 , @Sander3003, @ca-d  and others of course are wellcome to.


danyill:
Thank you for this important feature! :tada: 

Apologies for the delay in looking at this. 

I had a look at creating a new report from the 'Substation' editor 

My comments, as is often the case are a matter of taste. UI is hard, progress is important and the merits of these I can happily leave to your discretion since my ability to implement the harder ones is distant :wink:

Some of these relate to other things than the PR but I group them together (perhaps inappropriately, let me know if you want me to log issues for them and they are not relevant here) under the general usability of the feature. 

1. Use of abbreviations.

    From a user point of view have we made a specific decision to favour abbreviations in preference to the name of the attribute?

    image

    image

    My preference would be to have a plain description and the standard specific terminology bracketed, in a tooltip or formatted differently (in a code style). I think this improves the approachability of the product.

    Something like:

    image

    To connect the dialogue with a verb form, I would put underneath "Trigger Options", "Trigger when the following occurs".

    In the "Optional Fields" dialog, I would state below the title, something like: "Include the following in each report:" and then remove the words ("Add") and reverse the bracketed and non-bracketed terms.

2. User feedback. On completion of the newly created Report, I expect to see a confirmation that the action worked correctly by seeing the user returned to the "Select Report Control" with the cursor scrolled to the recently created report as a confirmation that the process has been completed (these blocking modals don't easily show a process or outcome as mentioned previously). This would be similar to what the "remove" button does at the moment.

3. From a user experience point of view, after adding an FCDA I probably need to add another one and it is very likely the next row or the same data object or LN. The UI allows this (but is a little challenging to use). To facilitate understanding could we remove the device name from this screen and only show it from the logical device, perhaps changing the title to `Add FCDA to <iedName>`?

    e.g. 
        
    image

    image

4. The UI binds the end-user to what I call an "order of operations problem". Consider the following:

    image

    The DOs and DAs are out of order with the LNs caused by the end-users selection. But 10 seconds after the edit, the end-user has forgotten their order of operations and should not need to know it.

     I think the DO and DAs associated with the LN should _match_ the order of the entries and not be modified by the user's selection order. The problem becomes severe when you realise 10 minutes after adding it to a dataset of 50 items you would like to remove it and can no longer find where it is in the list (unless doing so by methodical search or by exiting and removing it as a separate operation which you the end-user must then track out of band as a "to-do" task).

5. As an end-user I often wish to maintain a "logical order" of the items in the dataset. I would like to be able to drag and drop items in the dataset vertically when I am editing the dataset to reposition them.

6. I think the remove button for the report belongs with "cancel and save", it belongs to the report control "as a whole".

    image

7. Sometimes (in fact very often) on some IEDs (from SEL and Siemens) I get to the add Data Attributes dialog and have no options:

    image

    I think when this happens the report is created but I cannot later open or edit it:
    
   image

   I can investigate further if this is of interest.... I think this happens when I try to create a report control on an existing IED rather than using the toolbar shown below:

    image

    I can confirm that when trying to create a report from a conforming SCD file in SEL, Siemens and Micom if I start from the IED itself to create  new report control, I end up with no options in the "Add Data Attribute" screen.

8. Just fiddling I seem to be able to generate invalid options and produce diagnostic errors. I can work on some reproduction cases if these are not apparent.
    
    image

9. Often I only get asked once per instance of open-scd the following question and when creating later reports on other IEDs I am no longer asked the question:

    image

    This may be related to 7.

10. Is it the correct workflow to add "FCDAs" without an explicit "Add Dataset" step and have the dataset name allocated automatically based on the report name? I feel this doesn't follow the semantics of the IEC 61850 workflow and forces the user to correct/modify decisions made by the tool about the engineering workflow (a utility might want to a particular semantic for dataset names and this should be able to be done upon report creation).

JakobVogelsang:
I finally could proceed the work on this set of wizards, and can answer you to your questions/hints/user-feedback:

1. I pushed this in one extra issue, as this affects other wizards as well. With the `wizard-checkbox` at least, I would want to follow your recommendation and put the describing information first
2. I think this is something that can be done in the `Publisher` plugin very easy. With wizards, it is not easy to implement as they can be triggered from different sources. I would argue, that on a long run this wizards will be used to quickly add a missing report instead of being in the engineering step to define many `ReportControl` elements where a dedicated wizard would do a better job
3. Extra issue. The multi option as is, is hard to use, and we need to do some refactoring here.
4. Extra issue. The multi option as is, is hard to use, and we need to do some refactoring here.
5. This is something that a power user like you would say ;). I would not highly prioritize it. If we do your job right and the vendors tools do as well, the order of the data in the data set does not matter, I would say :). In any case, the wizard with its already existing limitation is not the right location to do that. That might fit better to the plugin.
6. True!! Opened a extra issue for this. We have that pattern very often in the code and I would like to change it for each the wizards
7. Bug that is fixed now
8. Bug that is fixed now
9. Bug that is fixed now
10. That is a topic that I knew would trigger discussion. Foremost, you are right kind of. Theoretically, you can create `DataSet`s without the need to connect those to any control blocks. However, I found that always the hardest to explain in my IEC 61805 trainings. The added value I THINKS is not so big. The configuration becomes more complex, and you don't save as much size AND you always have to be careful to not mix and match GSE/SMV and Report dataset as they theoretically can be too big for the first. Most tools I have seen just ignore the fact and behave as if those are strictly connected. I would love to further discuss this and open a discussion field [discussion](https://github.com/openscd/open-scd/discussions/new?category=iec-61850-related)

src/icons.ts
```diff
@@ -8,6 +8,12 @@ export const reportIcon = svg`<svg style="width:24px;height:24px" viewBox="0 0 2
     <path fill="currentColor" d="M9,7H13A2,2 0 0,1 15,9V11C15,11.84 14.5,12.55 13.76,12.85L15,17H13L11.8,13H11V17H9V7M11,9V11H13V9H11M12,2A10,10 0 0,1 22,12A10,10 0 0,1 12,22A10,10 0 0,1 2,12A10,10 0 0,1 12,2M12,4A8,8 0 0,0 4,12C4,16.41 7.58,20 12,20A8,8 0 0,0 20,12A8,8 0 0,0 12,4Z" />
 </svg>`;
 
+export const addReportIcon = svg`<svg style="width:24px;height:24px" viewBox="0 0 24 24">
```
Flurb, line 11: It's a good idea to split this icons up, I already made a suggestion in #570; I just added a icons folder to the root, so we can have separate `icons` files in there.

src/wizards/foundation/finder.ts
```diff
@@ -0,0 +1,80 @@
...
+function getIED(parent: Element): Element[] {
+  if (parent.tagName === 'SCL')
+    return Array.from(parent.querySelectorAll('IED')).filter(isPublic);
+
+  return [];
+}
+
+export function iEDPicker(doc: XMLDocument): TemplateResult {
+  return html`<finder-list
+    path="${JSON.stringify(['SCL: '])}"
```
Flurb, line 44: Why do we `stringify` here? If I look at `Help.ts`, we can just insert an object to the `path` property.

Flurb (changes requested):

src/wizards/foundation/finder.ts
JakobVogelsang: Yeah, that has to do with the lit-element library. When you put a `.` in front of it you set the `property` not the attribute. This cannot be done here because we will lose information navigating back and forth between pages using `next` and `prev`.

src/icons.ts
JakobVogelsang: I will rebase on #570. You are right, there are too many of them already :)

Flurb (approved):

src/icons.ts
JakobVogelsang: Wait, I just saw that I do not use this icon. I will add it to an extra issue. Then there will be no merge conflicts :)

PR #547, opened 2022-02-17
feat(wizards/smvopts): add edit wizard

### LINKED ISSUE #429: Add wizard to edit SmvOpts element
As a system configuration engineer, I want to edit attributes of the SCL element `SmvOpts` to configure sampled values 

**Requirement**:
- shall be accessible from edit wizard of SampledValueControl (#428 )
- allow returning back to edit wizard on secondary wizard action

**Type description**: 
attribute | description |  
 --- | --- |
refreshTime |  whether refresh time shall be included into the telegram |   
sampleSynchronized | whether the synch status shall be included into the telegram always true with >Ed2  |    
sampleRate | whether sample rate shall be included into the telegram  |
dataSet | whether data set name shall be included into the telegram  |
security | whether security configuration shall be included into the telegram  | 
dataRef | deprecated with Ed2 | 
timestamp | whether time stamp shall be included into the telegram  | 
synchSourceId | whether the synch source id see IEC61850-9-3 shall be included into the telegram  | 

**Restriction** :
- 

attribute | required |  pattern  | min length  | max length |  
 --- | --- | --- | --- | --- | 
refreshTime | no | xs:boolean |  |  |   
sampleSynchronized | no | xs:boolean |  |    |    
sampleRate | no | xs:boolean |   |  |
dataSet | no | xs:boolean |  |  |
security | no | xs:boolean |  |  | 
dataRef | no  | xs:boolean  |  |  | 
timestamp | no | xs:boolean |  |  | 
synchSourceId |  no | xs:boolean |  |  | 

**Type definition**:
- no further only boolean values

**Further Information Link**
- 

**Edition 1 vs Edition 2**
- the restrictions are valid for Ed2.1 and can differ for Ed1 and Ed2. This shall not be taken into account for not

JakobVogelsang:
Closes #429


test/integration/wizards/sampledvaluecontrol-wizarding-editing.test.ts
```diff
@@ -239,6 +240,28 @@ describe('Wizards for SCL element SampledValueControl', () => {
...
+        const editSmvOptsButton = <Button>(
+          element.wizardUI.dialog!.querySelector(
+            'mwc-button[id="editsmvopts"]'
+          )!
+        );
+        expect(editSmvOptsButton).to.exist;
+
+        await editSmvOptsButton.updateComplete;
+        editSmvOptsButton.click();
+        await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
Flurb, line 253: Why isn't `await editSmvOptsButton.updateComplete;` enough here, instead of a sleep (which is not recommended).

Using this should be enough and is a better solution compared to a sleep:
```
await editSmvOptsButton.updateComplete;
editSmvOptsButton.click();
await editSmvOptsButton.requestUpdate();
await editSmvOptsButton.updateComplete;
```

Too bad we need an extra `requestUpdate()`, I think because of the way tests are being set up and a `click()` doesn't automatically triggers a render?

Flurb (changes requested):

test/integration/wizards/sampledvaluecontrol-wizarding-editing.test.ts
JakobVogelsang: The reason is that we need to await an animation frame. With this test on button click one of the wizards closes and the other opens. We have that all over our tests and we have to found a solution to that until now.

I think the solution you have proposed would not work in all cases. It would cause timing issues depending on the power of the mchanine the tests are run on. The line `editSmvOptsButton.requestsUpdate()` eats up some time as this button disapears shotsly after clicking on this button (edit SampledValueControl wizards closes to make room or the edit SMV wizard).

src/translations/en.ts
```diff
@@ -38,6 +38,12 @@ export const en = {
     smpMod: 'Sample mode',
     smpRate: 'Sample rate',
     nofASDU: 'Samples per paket',
```
danyill, line 40:
```suggestion
    nofASDU: 'Samples per packet',
```

src/translations/en.ts
```diff
@@ -38,6 +38,12 @@ export const en = {
...
+    SmvOpts: 'Optional Information',
+    refreshTime: 'Add time stamp to SMV telegram',
+    sampleRate: 'Add sampled rete time to SMV telegram',
+    dataSet: 'Add DataSet name to SMV telegram',
+    security: 'Potential future use. E.g. dig signature',
+    synchSourceId: 'Add identity of sync source to SMV telegram',
```
danyill, line 46:
```suggestion
    SmvOpts: 'Optional Information',
    refreshTime: 'Add timestamp to SMV packet',
    sampleRate: 'Add sample rate to SMV packet',
    dataSet: 'Add DataSet name to SMV packet',
    security: 'Potential future use. e.g. digital signature',
    synchSourceId: 'Add sync source id to SMV packet',
```

danyill (changes requested):
A few minor translation comments.
Some of the comments on ReportControlBlocks are applicable here.

src/translations/en.ts
JakobVogelsang: Thanks for those hints. Sorry I am not really good at these things :) But together we will make it work :)

PR #549, opened 2022-02-21
feat(plugins/Subscription): Create 'Subscription' plugin for GOOSE subscriptions

### LINKED ISSUE #525: Create 'Subscription' plugin for GOOSE subscriptions
[Discussion topic](https://github.com/openscd/open-scd/discussions/520)

To create the subscription functionality, we need a new plugin called 'Subscription'. This plugin creates a new tab in the user interface for the functionality of subscribing IEDs to GOOSE messages. This issue contains the creation of the plugin and showing the published GOOSE messages and IEDs which can subscribe or are already subscribed. This will later be extended.

Since it is likely that there will be more plugins, depending on the method implemented, it is desirable that the Js file has a suffix of the subscribtion method used. For example:

```
  {
    name: 'Subscription',
    src: '/src/editors/Subscribtion_ABB.js',
    icon: 'link',
    kind: 'editor'
  }
```
image

Show all published GOOSE messages from all IEDs:
``` XML
<GSEControl datSet="DataSet" confRev="20001" appID="SIP1/VI3p1/LLN0/Control_DataSet" name="Control_DataSet" type="GOOSE">
	<IEDName apRef="E">SIP_Merg</IEDName>
</GSEControl>

```

When selecting an IED, show all `<extRef>` with serviceType = GOOSE of all `LN`s.
``` XML
<Inputs>
	<ExtRef iedName="SIP_Client" ldInst="VI3p1" lnClass="PTRC" lnInst="1" doName="Op" daName="general" serviceType="GOOSE" srcLDInst="VI3p1" srcLNClass="LLN0" srcCBName="Control_DataSet" pServT="GOOSE" desc="RxOp" intAddr="RxOp;/Op/general" pDO="Op" pDA="general" />
	<ExtRef iedName="SIP_Client" ldInst="VI3p1" lnClass="PTRC" lnInst="1" doName="Op" daName="q" serviceType="GOOSE" srcLDInst="VI3p1" srcLNClass="LLN0" srcCBName="Control_DataSet" pServT="GOOSE" desc="RxOp" intAddr="RxOp;/Op/q" pDO="Op" pDA="q" />
</Inputs>

```

Flurb:
Resolves #525


danyill:
This is exciting! :tada: 

 I'm not familiar with ABB IEDs. I thought I'd have a play anyway.

When there are many devices or transmit messages in the left pane, then it would be convenient if the right pane was "sticky" such that when the left-hand side is scrolled, and a new IED is selected then the right-hand side remains in view (the user doesn't have to scroll to the top on the RHS).

Otherwise, if I understand correctly, the user must constantly scroll up and down if working with IEDs at the bottom of the stack.

Here's an example file: [ArcFlash_IEC61850_GOOSE_2-section.scd.zip](https://github.com/openscd/open-scd/files/8120007/ArcFlash_IEC61850_GOOSE_2-section.scd.zip)

Flurb:
> This is exciting! 🎉
> 
> I'm not familiar with ABB IEDs. I thought I'd have a play anyway.
> 
> When there are many devices or transmit messages in the left pane, then it would be convenient if the right pane was "sticky" such that when the left-hand side is scrolled, and a new IED is selected then the right-hand side remains in view (the user doesn't have to scroll to the top on the RHS).
> 
> Otherwise, if I understand correctly, the user must constantly scroll up and down if working with IEDs at the bottom of the stack.
> 
> Here's an example file: [ArcFlash_IEC61850_GOOSE_2-section.scd.zip](https://github.com/openscd/open-scd/files/8120007/ArcFlash_IEC61850_GOOSE_2-section.scd.zip)

Hi @danyill , good suggestion! I can take a quick look at that!

Flurb:
@danyill I implemented the scrollable lists, I would love to have your feedback!

danyill:
Scrollabe lists are the :bomb: :+1: 
That makes this feel very usable.

I'm not sure what's in scope for this feature. Should we be showing additional fields? e.g. for all dataset items which are subscribed? For SEL relays it is useful to know the `intAddr` parameter (and I think ABB too).

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,219 @@
...
+      this.subscriberWrapper.scrollTo(0,0);
+    }
+  }
+
+  render(): TemplateResult {
+    const partialSubscribedIeds = this.availableIeds.filter(ied => ied.partial);
+
+    return html`
+      <h1>${translate('subscription.subscriberIed.title', {
+        selected: this.gseName ? this.iedName+'>'+this.gseName : 'IED'
```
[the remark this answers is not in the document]
danyill: Also I think we should put a space in here to facilitate wrapping.

```suggestion
        selected: this.gseName ? this.iedName + ' > ' + this.gseName : 'IED'
```

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  static styles = css`
+    h1 {
+      color: var(--mdc-theme-on-surface);
+      font-family: 'Roboto', sans-serif;
+      font-weight: 300;
+      overflow: hidden;
+      white-space: nowrap;
+      text-overflow: ellipsis;
```
danyill, line 203: If we're going to treat text-overflow like this can we add a tooltip.
I would prefer not to use an ellipsis because the end information is quite useful to the end user (I don't think its optional):

image

danyill (commented):
I thought I should read the code and as I did I had a  few minor comments, use as you wish :wink:

src/editors/Subscription.ts
```diff
@@ -0,0 +1,49 @@
+import { LitElement, html, TemplateResult, property, css } from 'lit-element';
+
+import '@material/mwc-fab';
+
+import './subscription/subscriber-ied-list.js';
+import './subscription/publisher-goose-list.js';
+
+/** An editor [[`plugin`]] for subscribing IEDs to GOOSE messages using the ABB subscription method. */
+export default class SubscriptionABBPlugin extends LitElement {
```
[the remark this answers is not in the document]
JakobVogelsang: This is how I know it from PCM600 where you do not connect individually. We had the discussion how to combine different types of subscriber processes. We decided to separate those through plugins. That means that the way you are used to working - with individual FCDA connecting to individual ExtRef I guess. And the other one called `later binding`

JakobVogelsang:
I have tried it out for the first time and have some UI remarks before diving into the code:
1. The two lists are not so visible. Having them in a slightly different color like we do with the `action-pane`s would be a helper already, I think
2. The header for the source would sound better `Publicher: GOOSE` or `GOOSE publisher`. The right side of the sink I would add a `of` making it `Subscriber IED of GOOSE: ....`
3. I also think the headers should not be clipped. I, this has to be you can clip at the beginning not at the end as the valuable information is at the end
4. Using `filtered-list` could be a nice option to reduce complexity. You have to imagine that there are project that have 250+ IED in the file. There you see that you don't see.
5. Maybe reuse the `gooseIcon` for the list-items that represent a `GSEControl` :)
6. This plugin can be a default from the beginning. When we add other subscriber plugins later, we can decide to change that. But for now!

src/editors/subscription/publisher-goose-list.ts
```diff
@@ -0,0 +1,79 @@
...
+  doc!: XMLDocument;
+
+  private get ieds() : Element[] {
+    return (this.doc)
+      ? Array.from(this.doc.querySelectorAll(':root > IED')).sort((a,b) => compareNames(a,b))
+      : [];
+  }
+
+  /**
+   * Get all the published GOOSE messages, ABB method.
```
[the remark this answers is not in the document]
Flurb: Can be removed here, no added value indeed.

src/translations/en.ts
```diff
@@ -261,6 +261,19 @@ export const en = {
...
+    none: 'None',
+    publisherGoose: {
+      title: 'Publisher GOOSE'
+    },
+    subscriberIed: {
+      title: 'Subscriber {{ selected }}',
+      subscribed: 'Subscribed',
+      availableToSubscribe: 'Available to subscribe',
+      partiallySubscribed: 'Partially subscribed',
+      noGooseMessageSelected: 'No GOOSE message selected.'
```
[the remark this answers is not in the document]
Flurb: Fine for me!

src/editors/Subscription.ts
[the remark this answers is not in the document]
Flurb: Did this answer your question @danyill ?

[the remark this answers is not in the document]
danyill: Yes, thanks both.

src/editors/subscription/subscriber-ied-list.ts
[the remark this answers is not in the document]
Flurb: Putting a space at the beginning and end of the `>` is nice, but not consistent.
I would also suggest to do that, but then we need to do it in the entire OpenSCD application.

@JakobVogelsang worth an issue?

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,219 @@
...
+              : ``
+          }${
+            fcda.getAttribute('doName')
+              ? `[doName="${fcda.getAttribute('doName')}"]`
+              : ``
+          }${
+            fcda.getAttribute('daName')
+              ? `[daName="${fcda.getAttribute('daName')}"]`
+              : ``
+          }`)) {
```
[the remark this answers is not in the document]
Flurb: Good suggestion! Makes the code a bit more clean :)

src/editors/subscription/subscriber-ied-list.ts
Flurb: This is also something we use throughout the OpenSCD application.
If we want to change this, it needs to be changed consistently. 

@JakobVogelsang , worth an issue?

Sander3003:
As @Flurb mentioned small improvements and bugs can be solved in this pull-request. Bigger improvements require a seperate issues. By doing so we have small improvements (more easy to review).

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
     missing: 'DataTypeTemplates fehlen',
     add: 'DataTypeTemplates hinzufügen',
   },
+  subscription: {
+    none: '???',
```
JakobVogelsang, line 268: Keine Verbindung vorhanden

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+    publisherGoose: {
+      title: '???'
```
JakobVogelsang, line 270: GOOSE-Publizierer

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+    },
+    subscriberIed: {
+      title: '???',
```
JakobVogelsang, line 273: Verbunden mit {{ selected }}

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+      subscribed: '???',
```
JakobVogelsang, line 274: Verbunden

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+      availableToSubscribe: '???',
```
JakobVogelsang, line 275: Kann verbunden werden

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+      partiallySubscribed: '???',
```
JakobVogelsang, line 276: Teilweise verbunden

src/translations/de.ts
```diff
@@ -264,6 +264,19 @@ export const de: Translations = {
...
+      noGooseMessageSelected: '???'
```
JakobVogelsang, line 277: Keine GOOSE ausgewählt

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,221 @@
...
+  private async onGOOSEDataSetEvent(event: GOOSEDataSetEvent) {
+    this.iedName = event.detail.iedName;
+    this.gseName = event.detail.gseName;
+
+    const dataSet = event.detail.dataset;
+
+    this.clearIedLists();
+
+    Array.from(this.doc.querySelectorAll(':root > IED')).forEach(ied => {
+      const inputs = ied.querySelector(`LN0[lnClass="LLN0"] > Inputs`);
```
JakobVogelsang, line 79: I think this is not completely correct. `GSEControl` are limited to `LN0` as parent, but they can be stored in any of the `LN0`. Basically, you can store them in any of the logical devices. The query down below takes the first `LN0` that has a `Input` not misses others that potentially are allocated to another logical device in the same IED.

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (!inputs) {
+        this.availableIeds.push({element: ied});
+        return;
+      }
+
+      /**
+       * Count all the linked ExtRefs.
+       */
+      dataSet.querySelectorAll('FCDA').forEach(fcda => {
+        if(inputs.querySelector(`ExtRef[iedName=${event.detail.iedName}][serviceType="GOOSE"]` +
```
JakobVogelsang, line 95: I would leave out the limitation `[serviceType="GOOSE"]`. This is not present in Ed1 and logically not neccassery here. When the user clicks on one of the GOOSEs, he wants to see all valid connections, not only those taged as GOOSE.

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  /** Current selected GOOSE message. */
+  gseName!: string;
+  
+  @query('div') subscriberWrapper!: Element;
+
+  constructor() {
+    super();
+    this.onGOOSEDataSetEvent = this.onGOOSEDataSetEvent.bind(this);
+    const openScdElement = document.querySelector('open-scd');
```
JakobVogelsang, line 59: Why do you dispatch and listen on open-scd? My gud feeling would say we can dispatch in the object where we click on, and we can listen on the GOOSE tab/Subscriber.

We do use `open-scd` with menu type plugin as well or better we have to as on re-render the connection to `open-scd` gets lost. Is this an issue as well here? If so I would ask you to add a comment `TODO: ...`

src/foundation.ts
```diff
@@ -280,6 +280,26 @@ export interface ResetDetail {
   kind: 'reset';
 }
 
+export interface GOOSEDataSetDetail {
```
JakobVogelsang, line 283: 1. Do we need to define this event in foundation? So far I think we have events defined that are use globally be the whole application. However, I am not sure if we can move it to `editors/subscriber/foundation`.

2. The name is bothering me. The other event can of tell what they do or have been triggered by. I cannot think of a better name than `GOOSeSelectEvent` - not perfect either :(

src/foundation.ts
```diff
@@ -280,6 +280,26 @@ export interface ResetDetail {
...
+  iedName: string;
+  gseName: string;
```
JakobVogelsang, line 285: I would make it gseControl: Element. The name is part of it, and we can use it for other purposes later in the application.

src/foundation.ts
```diff
@@ -2637,6 +2657,7 @@ declare global {
     ['open-doc']: OpenDocEvent;
     ['wizard']: WizardEvent;
     ['validate']: ValidateEvent;
+    ['goose-dataset']: GOOSEDataSetEvent;
```
JakobVogelsang, line 2660: I am not sure if we can declare this somewhere else. So far we need it only within the `Subscriber` plugin and so log it is the case we should try to define it there. However, not sure it this can be done. @ca-d can you help out?

JakobVogelsang (changes requested):
The plugin is nice!! Thank you for that. I found a bug and would love to have the german translation added. Everything else are comments.

src/editors/subscription/subscriber-ied-list.ts
Flurb: Ah I thought every IED only has 1 LN0 element. This is not correct?

src/editors/subscription/subscriber-ied-list.ts
Flurb: I don't think we want to listen in the plugin itself, because the SubscriberIEDLIst is the only element it's interested in. How can I add an EventListener here without using the open-scd element?

src/foundation.ts
Flurb: I also wanted it somewhere within the `subscription` plugin for example, but it only works if it's placed here :)
A prefer a better solution!

src/foundation.ts
Flurb: Good idea, in the manage goose services I also changed iedName to full IED element :)

src/foundation.ts
Flurb: I see the interfaces always ends with `-Detail`, why should this be different? `GOOSESelectDetail` is fine by me as well.

src/editors/subscription/subscriber-ied-list.ts
JakobVogelsang: You must have one LN0 per LDevice but you can have multiple LDevices per Server and thus multiple LN0 per Server

JakobVogelsang: True Inputs is a single element within one LN0

JakobVogelsang: I think to fix it you need something like `ied.querySelectorAll(LN0 > Inputs)` the `[lnClass="LLN0" ]` is not neccassery as is always the case for `LN0`

danyill:
Thanks for this @Flurb this seems like a very important feature :tada: :trumpet: 

Should I review this or #569?

Off the top of my head from a UI point of view

* I would like to see a little edit option for each GSEControl so that I can look for further information from the same screen. That doesn't seem too difficult since the Edit GSEControl wizard I think already does this.
* Perhaps that's sufficient, but I also wonder if we should the description field in a tooltip or beneath the GSEControl in the UI?
* Should the scroll bars of the GOOSE Publisher and Subscriber be invisible/width zero?
* How do we show partial subscriptions of a GOOSE dataset?

I'm quite interested in having a subscription plugin for SEL relays where the `intAddr` would need to be visible/settable for the dataset item which is subscribed.

Flurb:
> Thanks for this @Flurb this seems like a very important feature 🎉 🎺
> 
> Should I review this or #569?
> 
> Off the top of my head from a UI point of view
> 
> * I would like to see a little edit option for each GSEControl so that I can look for further information from the same screen. That doesn't seem too difficult since the Edit GSEControl wizard I think already does this.
> * Perhaps that's sufficient, but I also wonder if we should the description field in a tooltip or beneath the GSEControl in the UI?
> * Should the scroll bars of the GOOSE Publisher and Subscriber be invisible/width zero?
> * How do we show partial subscriptions of a GOOSE dataset?
> 
> I'm quite interested in having a subscription plugin for SEL relays where the `intAddr` would need to be visible/settable for the dataset item which is subscribed.

I think @JakobVogelsang already reviewed both, am I right Jakob?
For your other suggestions:

> I would like to see a little edit option for each GSEControl so that I can look for further information from the same screen. That doesn't seem too difficult since the Edit GSEControl wizard I think already does this.

This was already the plan, I'm not sure if we already have an issue for that. Can you check this? :)

> Perhaps that's sufficient, but I also wonder if we should the description field in a tooltip or beneath the GSEControl in the UI?

Good suggestion, create an issue for that :)

> Should the scroll bars of the GOOSE Publisher and Subscriber be invisible/width zero?

You mean while still being scrollable? Why do you want that? Or maybe I'm not getting it :)

> How do we show partial subscriptions of a GOOSE dataset?

This is already implemented, we have 3 lists on the right: Fully subscribed, partially subscribed and not subscribed. IEDs that are inside the not subscribed/partially subscribed can be fully subscribed. IEDs in the Fully subscribed section can be unsubscribed.

I hope this answers all your questions :)

danyill:
Thanks @Flurb I have created some issues. @JakobVogelsang asked me to review... but I may have missed the moment.

Anyway, from a style/user-facing perspective:

>>    Should the scroll bars of the GOOSE Publisher and Subscriber be invisible/width zero?
>You mean while still being scrollable? Why do you want that? Or maybe I'm not getting it :)

The height of the GOOSE publisher on Chrome is greater than the contained elements.

image

This gives us a scrollbar on the publisher where we don't need it _and_ a scrollbar on the overall window which is too many scrollbars. They are also the default scrollbars which are intrusive. Perhaps not zero width but less width might be nice.

Probably a CSS change is required which would remove the scrollbars for this case.
Perhaps a calc function should be used to subtract the height of the `mwc-top-app-bar-fixed` element...

Even so, there is the case with more IEDs/GSEControls it will extend.

In this case:
* the scrollbar appears to start partway done the publisher pane and abruptly so it doesn't "look right" to my (somewhat unkind) :eye: 
* Scrollbars as shown here are used in OpenSCD so this is consistent with the overall UI. In the wider world there is flamewars around scrollbars (do you need them, are they a good UI element etc.). Material Design [suggests](https://material.io/components/menus#behavior) there should be scrollbars but the modern take is much more minimalist than the scrollbars we have here.
** Browser support for CSS scrollbars is still weak https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
** But neither do we (at least me) want IE 5.5 scrollbars!
** And the [webkit option](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) is not on a standards track but is the only viable option

To be very concrete, what I was thinking of was something less visually intrusive, i.e.

```css  
    mwc-list::-webkit-scrollbar {
      width: 5px;
      height: 8px;
      background-color: var(--base3);
    }
    
    mwc-list::-webkit-scrollbar-thumb {
        background: var(--primary);
    }
```

This way the scrollbars blend into the selection at the top and for the subscribing pane (I may not have the colours exactly right).

image

I accept that I am a pedant and apologise and beg forgiveness :pray: 

A reasonable person will simply respond to this feedback with, "You didn't do that in #568 so why should I?"

JakobVogelsang:
> 

There will be an extra plugin for that, that will be similar but will allow you to connect single `FCDA` elements to one `ExtRef`. Combining all approaches people got used to with different vendors is too difficult. At least starting with it. When we see synergies later and think we can merge with one or the other together, we will do that.

JakobVogelsang:
> Thanks @Flurb I have created some issues. @JakobVogelsang asked me to review... but I may have missed the moment.
> 
> Anyway, from a style/user-facing perspective:
> 
> > > Should the scroll bars of the GOOSE Publisher and Subscriber be invisible/width zero?
> > > You mean while still being scrollable? Why do you want that? Or maybe I'm not getting it :)
> 
> The height of the GOOSE publisher on Chrome is greater than the contained elements.
> 
> image
> 
> This gives us a scrollbar on the publisher where we don't need it _and_ a scrollbar on the overall window which is too many scrollbars. They are also the default scrollbars which are intrusive. Perhaps not zero width but less width might be nice.
> 
> Probably a CSS change is required which would remove the scrollbars for this case. Perhaps a calc function should be used to subtract the height of the `mwc-top-app-bar-fixed` element...
> 
> Even so, there is the case with more IEDs/GSEControls it will extend.
> 
> In this case:
> 
>     * the scrollbar appears to start partway done the publisher pane and abruptly so it doesn't "look right" to my (somewhat unkind) 👁️
> 
>     * Scrollbars as shown here are used in OpenSCD so this is consistent with the overall UI. In the wider world there is flamewars around scrollbars (do you need them, are they a good UI element etc.). Material Design [suggests](https://material.io/components/menus#behavior) there should be scrollbars but the modern take is much more minimalist than the scrollbars we have here.
>       
>       * Browser support for CSS scrollbars is still weak https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
>       * But neither do we (at least me) want IE 5.5 scrollbars!
>       * And the [webkit option](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) is not on a standards track but is the only viable option
> 
> 
> To be very concrete, what I was thinking of was something less visually intrusive, i.e.
> 
> ```css
>     mwc-list::-webkit-scrollbar {
>       width: 5px;
>       height: 8px;
>       background-color: var(--base3);
>     }
>     
>     mwc-list::-webkit-scrollbar-thumb {
>         background: var(--primary);
>     }
> ```
> 
> This way the scrollbars blend into the selection at the top and for the subscribing pane (I may not have the colours exactly right).
> 
> image
> 
> I accept that I am a pedant and apologise and beg forgiveness 🙏
> 
> A reasonable person will simply respond to this feedback with, "You didn't do that in #568 so why should I?"

Thanks again for the very detailed review!!! As often, I think you are making a very good point. The scroll bares in windows are a pain. With iOS it is not a problem...don't know how it is with Linux based operating systems. 

I also think this should be handled globally in the `normalized.ts` as you pointed out already, the look and feel is the same in all plugins up until now. I know I am saying this a lot lately, but this should be an extra issue with an extra PR. In the PR especially, the complexity is already to big to be handled properly. I will open an issue with your comments as a starter :)

danyill (approved):
I reviewed it from a UI/functionality point of view. This is a super important feature which I expect to be looking at in detail in future, thanks! :tada:

dlabordus (approved):

PR #563, opened 2022-02-26
feat(foundation): add EditorAction type Update

JakobVogelsang:
We want to introduce a new EditorAction `Update` that allows to keep the `element` in the DOM and use its reference when the identity string cannot be use - e.g. name attribute must be exchanged. The action allows exchange the attributes of `element` with `newAttributes`. Updating children of `element` is out of scope, but can be done using `Create` and `Delete` as those can now be user for any kind of `Node`.

### DIFF
--- a/src/Editing.ts
+++ b/src/Editing.ts
@@ -21,6 +21,8 @@ import {
   SCLTag,
   SimpleAction,
   Replace,
+  Update,
+  isUpdate,
 } from './foundation.js';
 
 /** Mixin that edits an `XML` `doc`, listening to [[`EditorActionEvent`]]s */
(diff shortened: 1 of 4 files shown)


JakobVogelsang:
After trying to fiddle apart the two action types, one for the initial action the other one when undo and redo are triggered, I found it not worth it. 

The action system we have now has a straight forward logic where in my opinion the new `Update` action fits so long we have both `oldAttributes` and `newAttributes` be part of the API. 

For the user  it might be annoying to define `oldAttributes` when they are stored in `element` but I think we can more elegantly cope with that with the `createUpdateAction` function that does it for him/her. You can see how an action definition would look like in the unit and integrations tests for the `Update` action.

@ca-d @dlabordus @Flurb what would you say.

Flurb (approved):

Issue #290, opened 2021-08-10
Feature request, delete unused (not referenced GCBs)


BurTah00:
It is the case in MICOM Px40 series IEDs have no dataset referred or not defined in the communication sections GCBs. 
Each case I have to delete this GCBs manually by hand. 
Could you introduce a feature that deletes the GCBs if they have no dataset reference or they are not defined in the communication section or ...... (this can be extended with irrelevant cases that creates problem in StationScout) ?

Thanks!

JakobVogelsang:
I was thinking already about something like that. I would call it `clean up`. Can be used for report as well as sampled values and GOOSEs.

BurTah00:
Hi Jakob, could you give a rough date for this feature?

JakobVogelsang:
Hi Burak, sorry for the delay. I have prioritized this issue a bit higher. But being hones with you, I don't think we can make it until mid-December. I am confident however, we can include it until the end of the year. Do you have other suggestions for this `cleaning-up` feature? So far I can think of:
- GSEControl with no or invalid `datSet`
- SampledValueControl with no or invalid `datSet`
- DataSet that a not referenced by any control block
- `LNode`s that are invalid reference
- `LNodeType`, `DOType`, `DAType` and `EnumType` that are not referenced

BurTah00:
Hi Jakob, they are all fine and have use cases for different vendor implemantations.
What this story should not cover is:

-RCBControl with no  `datSet`

These RCBs would be filled out via a Client who creates a dynamic dataset.
Other than that your proposals make me happy.

danyill:
> @BurTah00 can you check if the clean-up tool of @danyill works for you?

Regrettably, the first task was not the one that was requested here (it's always the way). I'll do that next. I promise :handshake: 

> What this story should not cover is:
> -RCBControl with no datSet
>These RCBs would be filled out via a Client who creates a dynamic dataset.

As long as you can clearly see it's a ReportControl and filter on it this should be OK. Some will want to remove unused ReportControls and others will be using a dynamic allocation via the client.

BurTah00:
Hi,
@danyill thanks, I'll check this week hopefully when I have some time. Sorry for the delay.
You are right, RCBs with no datSet should not be wiped out, even I don't enjoy seeing crazy URCB101.....120 BRCB101...120 series in SIEMENS IEDs. That's the game...
Thanks again!

PR #568, opened 2022-02-28
feat(editors/cleanup) Cleanup editor

danyill:
This deals with one request of #290 and is a foundation to eventually incorporate the rest.

I have questions and problems which will need resolution before I can progress so this is very much WIP.

* [ ] GSEControl with no or invalid datSet
* [ ] SampledValueControl with no or invalid datSet
* [x] DataSet that are not referenced by any control block
* [ ] LNodes that are invalid reference
* [ ] LNodeType, DOType, DAType and EnumType that are not referenced

I've decided that a datatable would be a good way to address these. I want the ability to filter across different IED names, manufacturers, IED types and (e.g.) DataSets because I think a cleanup may deal with existing scd files with many IEDs or historical issues. It needs to be easy for the user to select large numbers of IEDs with fairly simple screening criteria.

Regrettably there is no material-web-component for a datatable and very little evidence that it might be arriving soon ([issue](https://github.com/material-components/material-web/issues/386) has been open for three years).

There is a very excellent article on the greatness of a good data-table element and features it might have here:

https://uxdesign.cc/design-better-data-tables-4ecc99d23356

The issue regarding data tables led me to look at [vadin-grid](https://vaadin.com/docs/latest/ds/components/grid) which offers (I think) lit-compatible Javascript front-end web components with a material theme. The particular "killer feature" I like is filter columns.

However snowpack (`3.8.6`) has an allergic reaction to this (unless I have done my imports or dependencies incorrectly which is ever so possible).
Interestingly, so does `3.8.8` but it does provide better error messages.
Intriguingly, `3.2.1` "just works". Some issues I found associated with snowpack suggest some recent regressions which (may) be related.

https://github.com/withastro/snowpack/issues/3724
https://github.com/withastro/snowpack/issues/3004
https://github.com/withastro/snowpack/discussions/3395
https://github.com/mui/material-ui/issues/26568

I am by no means finished with this, but I think it (mostly) works with the older version of snowpack.

https://user-images.githubusercontent.com/674783/155942498-04734edf-8e5b-4a66-b89c-249ec68d932c.mp4

`vaadin-grid` is license compatible (Apache 2) with Open-SCD but I don't want to simply assume we should add a dependency.
I think a data table is important for OpenSCD but there may be other (or better) (or lighter) options.

So... I am up for gentle mockery, advice and instruction :wink: 

* I'm not very good with type handling.
* I haven't written tests (yet) but plan to.
* I haven't fully researched whether the logic is sound.
* We need to be able to click on the dataset and open up the existing wizard to show the dataset.
* I am muddled most of the time :confused:


danyill:
> However snowpack (3.8.6) has an allergic reaction to this (unless I have done my imports or dependencies incorrectly which is ever so possible).
> Interestingly, so does 3.8.8 but it does provide better error messages.
> Intriguingly, 3.2.1 "just works". Some issues I found associated with snowpack suggest some recent regressions which (may) be related.

The error messages in 3.8.6 are typically when I click on the plugin I get a browser console error regarding cyclic references.

Sometimes instead I receive the following:

image

But it is not always the same error mesage.

JakobVogelsang:
The first impression is very good. The extra dependency is not nice, we were hoping to use the `mwc-data-table` that is on the roadmap for the component library we are using. `vaadin` is powerful, but adds I think something like 8 more dependencies to the repository. One thing that is important is: we have not merged to `lit2.0` and both the component libraries have this as a dependency. So you have to choose a version of `vaadin-grid` that still uses lit-element 2.xx and lit-hatm 1.xx.

JakobVogelsang:
Another solution would be to have a list for each of the things you want to clean up: In the list you can then use the secon span to add more information that you want to filter for something like:

```
dataset_name
IEDname (manufactorer) - iedType
```

When you put the identity string in the value attribute of the list-item, you even can filter for the device and other parents of the dataset.

Same list can be create for empty GSE,SMV and Report control blocks. Then you have all lists next to each other on a big screen. You have a lot of code where we use this type all the time and you can use filtered-list out of the box.

JakobVogelsang:
Don't get me wrong, I don't wont to restrict you to not work with tables. But I really think list might work better here :)

danyill:
:+1: Thanks for the feedback, I think it is fair. 

Let me have a go at doing as you suggested. This is a useful learning experience for me.

(:newspaper_roll: Material Web 3 is hopefully coming [late in 2022](https://github.com/material-components/material-web/issues/3050#issuecomment-1011447013) and perhaps that will deliver something. :relaxed: I guess I should be honest, my usage here is a little bit of a pretext - I ultimately want data tables for other things, [here are some incomplete ramblings](https://excalidraw.com/#json=DA7tb6hIHiIV_pqvl8xKC,uiWo0Mm-ZajU48vPJppC8A) from a little while ago)

danyill:
I have done something and now seek a preliminary review before I try to scale for the remaining tasks:

Some incomplete things which I would like some help with or to receive comments on:

1. image
    When I have no datasets, my button is not disabled. Fixing this currently escapes me.
2. Should the button be on the left-hand side? Or the right-hand side?
3. The button styling doesn't look correct but I'm unsure how best to improve. I think there should be more margin.
4. All German is courtesy of Google Translate and likely to be unintelligible.
5. I have not written tests before here and mocha is fairly new to me, I will take many comments. I have tried to generate a snapshot.

A few thoughts:

* I wanted to use a snackbar to indicated to the user that datasets were removed. But the validation snackbar interferes. Do we really need the validation snackbar to prevent any other snackbars? I think putting a number of errors on top of the validation icon might be an alternative to allow this to be used by other UI elements.
* I haven't carefully read the standard to see if this logic is correct but I'm confident others will be able to confirm/deny.

public/js/plugins.js
```diff
@@ -126,6 +126,13 @@ export const officialPlugins = [
     requireDoc: true,
     position: 'middle'
   },
+  {
+    name: 'Cleanup',
+    src: '/src/editors/Cleanup.js',
+    icon: 'cleaning_services',
+    default: false,
```
danyill, line 133: I think a cleanup is a specialised task I don't think this should clutter the UI. So the default should be false.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,274 @@
...
+
+/** An editor [[`plugin`]] for cleaning SCL references and definitions. */
+export default class Cleanup extends LitElement {
+  /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
+  @property()
+  doc!: XMLDocument;
+  @property()
+  disableDatasetClean = false;
+  @property()
+  gridRowsUnusedDatasets: iedData[] = [];
```
danyill, line 34: Are we able to store this property against the UI template in some fashion or is this correct to have an "aligned" object on the class?

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,274 @@
...
+          },
+        });
+      });
+    }
+    return actions;
+  }
+
+  async firstUpdated(): Promise<void> {
+    const checklist: List =
+      this.shadowRoot!.querySelector('#unuseddatasetlist')!;
```
danyill, line 234: Should our id be camel case? I guess so. Refactor to `#unusedDataSetList`

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,274 @@
...
+    checklist!.addEventListener('selected', () => {
+      this.disableDatasetClean =
+        (<Set<number>>(<MWCListIndex>checklist!.index)).size === 0 || (<MWCListIndex>checklist!.index === -1)
+    });
+  }
+
+  render(): TemplateResult {
+    return html`
+      <div id="containerTemplates">
+        <section tabindex="0">${this.getUnusedDatasets()}</section>
```
danyill, line 244: My plan is to add additional cleanup features as additional sections a lot like this.
Is there a better way -- a class per feature?

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  static styles = css`
+    ${styles}
+
+    :host {
+      width: 100vw;
+    }
+
+    #containerTemplates {
```
danyill, line 256: probably `#containerCleanup` would be better.

test/integration/editors/Cleanup.test.ts
```diff
@@ -0,0 +1,79 @@
+import { html, fixture, expect } from '@open-wc/testing';
```
danyill, line 1: I haven't provided a unit test but I will be happy to take guidance.

I am not sure what additional tests should be here? A snapshot with data for example...

test/integration/editors/Cleanup.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      )).click();
+      // two datasets should be deleted, leaving us with two
+      const dataSetCountCheck =
+        doc.querySelectorAll(
+          ':root > IED[name="IED1"] > AccessPoint > Server > LDevice > LN0 > DataSet'
+        ).length +
+        doc.querySelectorAll(
+          ':root > IED[name="IED3"] > AccessPoint > Server > LDevice > LN0[lnClass="LLN0"] > DataSet'
+        ).length;
+      expect(dataSetCountCheck).to.equal(2);
```
danyill, line 76: These seems a bit untidy for a test and perhaps points to design issues, WDYT?

danyill (commented):
Apparently I can't request changes on my own PR!

I note I've somehow deleted some snapshots erroneously which I must fix.
I have posed one or two questions.

public/js/plugins.js
JakobVogelsang: True

src/editors/Cleanup.ts
JakobVogelsang: There is no technical reason to not use camel sase id's so far I know. There are style guides that suggest avoiding id and use class attributes instead. See here, for example https://google.github.io/styleguide/htmlcssguide.html#ID_Selectors

I do use only classes in my latest commits and usually do not use camel case

src/editors/Cleanup.ts
JakobVogelsang: I would say so. We used the web-component in the past to structure code like that, but in my opinion this add a unnecessary shadowRoot boarder into the html DOM. There are ways you can structure the code without using classes. 

1. Break down the render method into several render methods like so 
```

renderSubMethod1():TemplateResult{
return html`...`
}
render(): TemplateResult {
  return 
      html `<section class="dataset">
             ${renderSubMethod1()}
           </section>
           <section class="gsecontrol">
             ${renderSubMethod2()}
           </section>`
}
```

3. The action that are needed to remove stuff can be split into two parts. One that dispatches the `editor-action` event, and the other of that can be a function as well that creates the Delete array. The function then can be allocated to any file within OpenSCD although you should as close as possible to the file you are using the function in.

test/integration/editors/Cleanup.test.ts
JakobVogelsang: My gut feeling is that you need to test these things:

1. The code queries unreferenced DataSet element correctly
2. Display those in the form that you have chosen: lists
3. The button is per default disabled
4. The list items are per default not selected
5. The button is not disabled when selecting at least one item
6. The button again becomes disable when deselecting all list items
7. The selected and only the selected DataSet are removed from the SCL

This can be done with unit tests and integrations tests:
1 and 2 can be tested with the snapshot which a recommend to put to the unit tests. If you have constructed a good SCD file with edge cases you can see by the snapshot if the query works as well
3. Can be done in the unit test. One tipp have a look at the `@query` decorator from lit-element we use. The exception could be something like `except('variableNameOfPLugin.queryNameToTheButton').to.have.property('disabled','true')`
4. You can use a `@queryAll` on all 'mwc-list-item's or just on the list and the can use the selected property of the list to test that
5.  You can click within unit tests as well and the check again for the button disabled property. Should be false then
6. Same thing instead you deselect all and check the disabled property
7. This must be done using a Editing integration test. We have a mock for that. Make sure this is the parent for your plugin under test and then the editor-action will do something with the file and you can check by querying the document for the to be deleted DataSet elements

test/integration/editors/Cleanup.test.ts
JakobVogelsang: So comment above

src/editors/Cleanup.ts
JakobVogelsang: I am not sure what you mean by that :)

test/integration/editors/Cleanup.test.ts
danyill: :+1: I mostly understand this and I think it is best that I have a go and we do a further review. Thanks.

JakobVogelsang: Looking foreward to review. Shall I have a look at the productive code meanwhile?

danyill: If you have time I am grateful. I hope to spend some time on this over the weekend.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * Get all the DataSets for a specific IED
+   * @param ied - The SCL IED element to search through.
+   * @returns An Array containing the name of all DataSets.
+   */
+  private getDatasetNames(ied: Element): Element[] {
+    return Array.from(
+      ied.querySelectorAll(
+        `:scope > AccessPoint > Server > LDevice > LN0[lnClass="LLN0"] > DataSet`
```
danyill, line 69: I am not entirely clear from IEC 61850-6 Ed 2.1 sec. 9.3.5 whether DataSet can be defined outside `LN0` or any `LN`

pp. 91 of the standard seem unclear to me, when it states "The LN contains the following elements: DataSet (see 9.3.7)..." and I am not clear if this is a reference to LN0 or "other Logical Nodes".

Should it be `LN0 || LN`.

This is the way I might interpret `SCL_IED.xsd` without having done any testing as `tAnyLN` allows the `tDataSet` element.

danyill: OTOH https://github.com/openscd/open-scd/blob/2e11cc8c277d31dd693ae54656e1c3b4cd23f8f3/src/menu/SubscriberInfo.ts#L117-L124

test/integration/editors/Cleanup.test.ts
JakobVogelsang: Will do so :)

src/editors/Cleanup.ts
JakobVogelsang: The element `DataSet` can be allocated within the complex type `tAnyLN` that is kind of the parent class for both `LN0` and `LN`. Btw the best thing to look up stuff like that is the schema not the PDF.

The example you see above is dealing with DataSet's for `GSEControl`. The element `GSEControl` as well as the element `SampledValueControl` can only be allocated to `LN0`. And because the referenced `DataSet` must be in the same parent, the `DataSet`s used within GOOSE or SMV are only within `LN0`. For the feature you are building, I would argue you need to look into both LN0 and LN.

danyill: Good :+1: 

I think that means the only thing which is left is `ReportControl` which can be defined on an LN with its own `DataSet`. So yes, we must do both.

That implies that the code for the control type selector is insufficient:

```
         :scope > AccessPoint > Server > LDevice > LN0[lnClass="LLN0"] > ${controlType}
```

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,283 @@
...
+  private getUsedDatasets(ied: Element) {
+    const controlReport = this.getControlDatasets(ied, 'ReportControl');
+    const controlGSE = this.getControlDatasets(ied, 'GSEControl');
+    const controlSV = this.getControlDatasets(ied, 'SampledValueControl');
+    return controlReport.concat(controlGSE, controlSV);
+  }
+
+  /**
+   * Get a named dataset from a specific IED name in the SCL file.
+   * @param ied - An SCL IED element.
```
danyill, line 108: Not true, just the name value as a string...

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,283 @@
...
+   * @param dataset - The name of a DataSet.
+   * @returns - An SCL DataSet Element
+   */
+  private getDatasetFromIED(ied: string, dataset: string): Element | null {
```
danyill, line 112: From our previous discussion:

> The example you see above is dealing with DataSet's for GSEControl. The element GSEControl as well as the element SampledValueControl can only be allocated to LN0. And because the referenced DataSet must be in the same parent, the DataSets used within GOOSE or SMV are only within LN0. For the feature you are building, I would argue you need to look into both LN0 and LN.

I think this means that `ied.name` and `dataset.name` are not sufficient to fully qualify the reference. I must research this, possibly I need to store an LN parameter as well.

danyill:
I have made (I think) some improvements to the user interface:

* The header says the number of unused datasets
* The button the user clicks indicates the number of unused datasets
* The datasets are ordered according to the identity string (i.e. sorted by IED first)
* Button disabling works properly and allowed logic simplification
* The checklist should be close to the item which means on the left, not the right. We don't want mistakes in deleting things so the visual distance between checklist and item should be minimised. It's either this or the container size must be reduced (although I've done that too and it still seems too big - I am not sure on the best approach with this). This does mean the select all is on the opposite side (in my mind that is OK, a slight discordance in the UI when you decide to do something "severe" is not a bad thing, it reminds one that with great power should come greater fear). I haven't considered consistency with other UI elements.

We must not have a clean function which instead of sweeping the house eats the house instead!

_Question:_ Should we include the dataset description here in some way?

Also:

* Ids converted to classes
* I hadn't taken into account LogControl blocks, so I've added them. I wonder if I missed anything else.
* The selectors are (now, I hope) improved since a `DataSet` can be under an LN and more work was needed to ensure uniqueness. 

The next thing will be some work on the tests.

src/editors/Cleanup.ts
danyill: I hope this is what I've done but I am not sure I understood.

test/integration/editors/Cleanup.test.ts
danyill: I think I have done all of these.

I am only unsure about:

> 7. This must be done using a Editing integration test. We have a mock for that. Make sure this is the parent for your plugin under test and then the editor-action will do something with the file and you can check by querying the document for the to be deleted DataSet elements

danyill:
I think I am mostly done but I have a few problems:

First, when I click to look at a dataset I lose allow selection within the editor when I return. I think it's probably a change in context around `this` but I'm not sure how to correct it.

Second, styling:

>Should the button be on the left-hand side? Or the right-hand side?
>>Right hand would fit more to the wizard handling

>The button styling doesn't look correct but I'm unsure how best to improve. I think there should be more margin.
>> This is how we use it in the wizards. But I would say outlined could be a solution.

I have added some padding because there was zero. Alas, despite a valiant effort, my CSS skills have not allowed me to move this button to the right :angry: 

Because I've moved the checklists to the left this doesn't feel so problematic but I am interested...

Other than that, I've added some tests and no doubt more could be done. I would be grateful for a review.

JakobVogelsang:
Hi @danyll, to be honest with you, I was satisfied with the first version you have pushed UI wise, although I am glad now that you have pushed yourself a little further. E.g that the amount of to be deleted `DataSet`s are displayed on the remove button and other little things that make the feature close to production ready :)

The two question you have I could solve
1. Use this additional style attribute on the mwc-button ittself `style="float:right; margin-bottom: 10px"`
2. I tested the `outlined` property on the button and I like it a lot. Would look something like that
grafik
3. The responsive width of the whole thing does not work, but I don't know why. I will have a look with my review.

Thanks a lot for your effort :)

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Render a user selectable table of unused datasets if any exist, otherwise indicate this is not an issue.
+   * @returns html for table and action button.
+   */
+  private getUnusedDatasets() {
+    // Unused Datasets
+    const gridRowsUnusedDatasets: datasetInfo[] = [];
+    this.ieds.forEach(ied => {
+      const usedDatasets = this.getUsedDatasets(ied);
```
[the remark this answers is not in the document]
danyill: > The level of reference for DataSet's it the parent anyLN. That means that a control block can only be connected to DataSet within the same parent !!NOT to any DataSet in the IED. I think this is not correct at the moment in the code

I think I understand. From the standard:

* For a report control block: "The referenced data set must be in the same LN as the control block."
* For a log control block: "The referenced data set must reside in the same LN as the control block."
* For a GSE control block: "It resides in LLN0 like the control block."
* For a sampled value control block: "A referenced data set must reside in LLN0."

>The second on is the approach with I think I would have chosen to solve a bit different. You are looking for all used DataSet name and compare it to the defined DataSet. I propebelly would have checked if the DataSet name is there and if it is used by any of the control block in the parent. Both are good the later can be done with a query and one forEcah loop like that:

Indeed, they may both be good, but yours is much better and makes the code more elegant because we keep the `DataSet` element in scope throughout the code and it simplifies it. I was thinking very much in terms of a hierarchy involving IEDs but this is neither necessary nor helpful here.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+        const isReferenced = Array.from(
+          parent?.querySelectorAll(
+            'GSEControl, ReportControl, SampledValueControl, LogControl'
+          ) ?? []
+        ).some(cb => cb.getAttribute('datSet') === name);
+
+        if (parent && (!name || !isReferenced))
+          unreferencedDataSets.push(dataSet);
+      });
+    this.unreferencedDataSets = unreferencedDataSets.sort((a, b) => {
```
JakobVogelsang, line 76: I would add an empty line between two logical sections. This makes the code much more readable.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+      if (aId < bId) {
+        return -1;
+      }
+      if (aId > bId) {
+        return 1;
+      }
+      // names must be equal
+      return 0;
+    });
+    return html`
```
JakobVogelsang, line 89: Same here an empty line would be beneficial.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+        <mwc-button
+          outlined
+          icon="delete"
+          class="cleanupUnreferencedDataSetsDeleteButton"
+          label="${translate('cleanup.unreferencedDataSets.deleteButton')} (${(<
+            Set<number>
+          >this.selectedItems).size || '0'})"
+          ?disabled=${(<Set<number>>this.selectedItems).size === 0 ||
+          (Array.isArray(this.selectedItems) && !this.selectedItems.length)}
+          slot="secondaryAction"
```
JakobVogelsang, line 144: I think this is a left over?

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+    return actions;
+  }
+
+  async firstUpdated(): Promise<void> {
+    this._cleanUnreferencedDataSetsList?.addEventListener('selected', () => {
+      this.getSelectedUnreferencedDataSetItems();
+    });
+  }
+
+  render(): TemplateResult {
```
JakobVogelsang, line 185: The sequence of methods could be improved. I would put `renderUnreferencedDataSets` right above `render` as this is only called there. It is easier to find this way :)

src/translations/de.ts
```diff
@@ -50,8 +50,9 @@ export const de: Translations = {
     showieds: 'Zeige IEDs im Substation-Editor',
     selectFileButton: 'Datei auswählen',
     loadNsdTranslations: 'NSDoc-Dateien hochladen',
-    invalidFileNoIdFound: 'Ungültiges NSDoc; kein \'id\'-Attribut in der Datei gefunden',
-    invalidNsdocVersion: '???'
+    invalidFileNoIdFound:
+      "Ungültiges NSDoc; kein 'id'-Attribut in der Datei gefunden",
+    invalidNsdocVersion: '???',
```
JakobVogelsang, line 55: I think you have to rebase. This might be a merge conflict.

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+  css,
+  html,
+  LitElement,
+  property,
+  TemplateResult,
+  query,
+  queryAll,
+} from 'lit-element';
+import { translate } from 'lit-translate';
+import { Button } from '@material/mwc-button';
```
JakobVogelsang, line 13: To be safe within our dev environment, we need to import the web-component we use in the file as well not only the classes. Here I would expect 
- `import '@material/mwc-button'`
- `import '@material/mwc-list/mwc-checked-list-item'`

src/editors/Cleanup.ts
```diff
@@ -0,0 +1,230 @@
...
+import { List, MWCListIndex } from '@material/mwc-list';
+import { ListItem } from '@material/mwc-list/mwc-list-item.js';
+
+import {
```
JakobVogelsang, line 17: here the import of the filtered-list 
`import '../filtered-list'`

We need this imports as web-test-runner does only load the component it needs for testing. Missing component import might lead to test errors.

test/unit/editors/ied/__snapshots__/da-container.test.snap.js
```diff
@@ -1,33 +1,33 @@
 /* @web/test-runner snapshot v1 */
```
JakobVogelsang, line 1: This is probably a rebase issue?

JakobVogelsang (changes requested):
This is already very good. So please don't take all the comments too seriously. The only thing I would ask you to change are the import statements. 

Over all LGTM!! Thank you for your big effort!!!

src/translations/de.ts
danyill: I will do so but I worry about Terrible Things

src/editors/Cleanup.ts
danyill: Good spotting :wink:

test/unit/editors/ied/__snapshots__/da-container.test.snap.js
danyill: I don't know why this has happened. Each time I generate a snapshot it modified a couple of container tests. I have rebased but the rebase was not terribly clean, you may need to take a look.

FTR: Nothing good in my life ever came from a messy rebase and I worry that this will not be the first time...

src/editors/Cleanup.ts
danyill: Quite right.

danyill:
OK, I'm gonna push the rebase. I have a backup branch up my sleeve in case this ends poorly but all tests pass I believe.

For the remaining cleanup tasks, would you prefer to:

* merge this and then add to it via MRs onto this?
* leave this PR open and have me add MRs against it?

Thank you for reviewing this and providing guidance :clap:. I hope to need fewer review comments for additional tasks :wilted_flower:

JakobVogelsang:
> OK, I'm gonna push the rebase. I have a backup branch up my sleeve in case this ends poorly but all tests pass I believe.
> 
> For the remaining cleanup tasks, would you prefer to:
> 
>     * merge this to main and then add to it via separate PRs?
> 
>     * leave this PR open and have me add PRs against this branch?
> 
> 
> Thank you for reviewing this and providing guidance 👏. I hope to need fewer review comments for additional tasks 🥀

The first would be better. What you have has clearly an added value for users and the other PRs would be simpler that ways :)

danyill:
I think I've dealt with all of these, although after the force push I can't see, perhaps one wasn't resolved...

JakobVogelsang (approved):
LGTM

PR #569, opened 2022-02-28
feat(plugins/Subscription): Manage GOOSE service connections

Flurb:
Also adding some tests for the subscription plugin, which were not included in the subscriber-plugin branch.


Flurb:
Updating an element is done as follows:
```
private async replaceElement(original: Element, clone: Element) {
    const parent = original.parentElement;

    this.dispatchEvent(
      newActionEvent({
        old: {
          parent: parent!,
          element: original,
          reference: original.nextSibling
        }
      })
    );

    await this.updateComplete;

    this.dispatchEvent(
      newActionEvent({
        new: {
          parent: parent!,
          element: clone,
          reference: clone.nextSibling
        }
      })
    );
  }
```

I want this to be done with 1 newActionEvent instead of 2, but I can't get it to work.
When I use a `new` element and an `old` element as well at the same time, it's just adding the `new` element into the `old` element.

It's working as it is now, but it's not a classy solution. You're free to fix it :)

JakobVogelsang (approved):
LGTM

PR #570, opened 2022-03-02
feat(plugins/IED): Add icon set to IED editor containers

### LINKED ISSUE #439: Add icon set to IED editor containers
As a CoMPAS user
I want to see better distinguished IED sections by using icons in the separate IED containers
So I can have a quicker recognition of different IED sections.

Flurb:
Resolves #439


dlabordus:
Strange that the Snapshots haven't changed?

JakobVogelsang:
Hi @dlabordus you can set a SVG by using the slot `icon` instead of the property `icon` with `action-pane` likeso:

```
<action-pane label="myLabel">
    <mwc-icon label="icon">${someSVG}</mwc-icon>
</action-pane>
```

The thing with the label property is correct. That was my mistake. I triggered rob to use &mdash and can not be done as a string. This could be refactored using `—` directly in the string.

src/editors/ied/access-point-container.ts
```diff
@@ -33,13 +34,13 @@ export class AccessPointContainer extends LitElement {
...
 
   render(): TemplateResult {
-    return html`<action-pane .label="${this.header()}">
-    ${Array.from(this.element.querySelectorAll(':scope > Server')).map(
-      server => html`<server-container
-        .element=${server}
-        .nsdoc=${this.nsdoc}
-        .ancestors=${[this.element, ...this.ancestors]}
-      ></server-container>`)}
+    return html`<action-pane .icon=${accessPointIcon} .label="${this.header()}">
```
JakobVogelsang, line 37: The same for the other cases. The reason your snapshot won't show any difference is that you have set the property, not the attribute. The snapshot test do test attributes. `lit-element` does write the attribute based on the property, but only for defined types. One such kind is a String, Boolean, Array and others. 

The way we intended to use `svg` within the `action-pane` header is liek so

```
<action-pane label="myLabel">
    <mwc-icon label="icon">${someSVG}</mwc-icon>
</action-pane>
```

Sorry for the missing documentation for that!!!

JakobVogelsang (changes requested):

Flurb:
> Hi @dlabordus you can set a SVG by using the slot `icon` instead of the property `icon` with `action-pane` likeso:
> 
> ```
> <action-pane label="myLabel">
>     <mwc-icon label="icon">${someSVG}</mwc-icon>
> </action-pane>
> ```
> 
> The thing with the label property is correct. That was my mistake. I triggered rob to use &mdash and can not be done as a string. This could be refactored using `—` directly in the string.

I implemented the solution for the IED container, but I don't think the icon alignment is as nice as the others. What do you guys think?

Schermafbeelding 2022-03-05 om 00 03 15

Sander3003:
@Flurb it looks good to me!

Flurb:
Let's fully compare them:

Old implementation:
Schermafbeelding 2022-03-07 om 09 36 54

Suggested implementation from Jakob:
Schermafbeelding 2022-03-07 om 09 36 25

It's not completely aligned, @JakobVogelsang do I need to do something in order to align it? I hope I don't need some custom CSS for this :)

danyill:
I'm not sure whose suggestion we're on now, but visually the alignment on the current commit looks excellent to me with my :eye: and my :eyeglasses: 

image

JakobVogelsang:
Hi @Flurb, this is a CSS issue in the `action-pane`. We should not resolve it here. You can open an extra issue for this that can be resolved in an extra step. @danyill the changed have not been made in this PR. I think @Flurb has tested it on his machine and found the issue.

Flurb:
Implemented it, and created an issue: #584

JakobVogelsang:
Thanks!!

JakobVogelsang (approved):
LGTM

PR #576, opened 2022-03-06
refactor: use complex actions to subscribe and unsubscribe

JakobVogelsang:
Most of the changed lines are due to code linting. We have a hook on pushing from git bash. You can use a prettier in VSCode instead.

- I changed the actions to complex action. Please add the title and its translation, as this might be something I want to discuss with @randy993 
- I moved the action definition to the plugins foundation as they are only used within this plugin
- I fixed the import statements. Please make sure that all web-component definitions are imported also the @materail one

### DIFF
--- a/src/editors/subscription/elements/goose-message.ts
+++ b/src/editors/subscription/elements/goose-message.ts
@@ -6,8 +6,8 @@ import {
   property,
   TemplateResult,
 } from 'lit-element';
-import { newGOOSESelectEvent } from '../../../foundation.js';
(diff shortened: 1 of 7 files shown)


Flurb (approved):

Flurb (changes requested):
The `SubscribeStatus` import in `test/unit/editors/subscription/elements/ied-element.test.ts` should be updated :)

Flurb (approved):

PR #582, opened 2022-03-08
fix(editors/template): properly update xxxType list after ad/editing

JakobVogelsang:


src/editors/Templates.ts
```diff
@@ -66,7 +66,10 @@ export default class TemplatesPlugin extends LitElement {
 
   openLNodeTypeWizard(identity: string): void {
     const wizard = lNodeTypeWizard(identity, this.doc);
-    if (wizard) this.dispatchEvent(newWizardEvent(wizard));
+    if (wizard)
```
ca-d, line 69: Could we refactor out the `if (!lnodetype) return undefined;` check into its own function so we don't have to generate a whole wizard we never display just in order to run it? If that's the only condition we're checking here, I'd prefer to do things that way around.

ca-d (changes requested):

src/editors/Templates.ts
JakobVogelsang: You are making a good point here. I saw that too and know that I build this foofoo into the code some year ago. When working on this issue, I though that this should be dealt with in an extra PR. It also makes sense since these wizard, basically all of them, need a refactor anyway. However, we can do it here too. What is your preference?

ca-d (approved):
LGTM

PR #583, opened 2022-03-08
feat(plugins/SampledValues): Create Sampled Values tab in Subscription plugin

Flurb:


public/js/plugins.js
```diff
@@ -20,6 +20,13 @@ export const officialPlugins = [
     default: false,
     kind: 'editor',
   },
+  {
+    name: 'Sampled Values',
```
JakobVogelsang, line 24: The name must be a bit more precise. We need to talk about naming for the different kind of subscriber plugins but for now I would suggest something like `Sampled Values Subscriber`

src/translations/de.ts
```diff
@@ -271,6 +271,19 @@ export const de: Translations = {
     missing: 'DataTypeTemplates fehlen',
     add: 'DataTypeTemplates hinzufügen',
   },
+  sampledvalues: {
+    none: 'Keine Verbindung vorhanden',
+    sampledValuesList: {
+      title: '???'
```
JakobVogelsang, line 277: Sampled Values

src/translations/de.ts
```diff
@@ -271,6 +271,19 @@ export const de: Translations = {
...
+    },
+    subscriberIed: {
+      title: 'Verbunden mit {{ selected }}',
+      subscribed: 'Verbunden',
+      availableToSubscribe: 'Kann verbunden werden',
+      partiallySubscribed: 'Teilweise verbunden',
+      noSampledValuesSelected: 'Keine Sampled Values ausgewählt'
```
JakobVogelsang, line 284: Keinen Kontrollblock ausgewählt

src/translations/en.ts
```diff
@@ -268,6 +268,19 @@ export const en = {
...
+    none: 'none',
+    sampledValuesList: {
+      title: 'Sampled Values'
+    },
+    subscriberIed: {
+      title: 'Subscriber of {{ selected }}',
+      subscribed: 'Subscribed',
+      availableToSubscribe: 'Available to subscribe',
+      partiallySubscribed: 'Partially subscribed',
+      noSampledValuesSelected: 'No Sampled Values message selected'
```
JakobVogelsang, line 281: `No control block selected` is enough. We are in the sampled value subscriber plugin. I think the rest is obvious

JakobVogelsang (commented):
LGTM in general. Some translations could be changed

JakobVogelsang (approved):
LGTM

PR #588, opened 2022-03-09
fix(plugins/Subscription): Empty 'Inputs' element after unsubscribing GOOSE should be deleted

Flurb:


JakobVogelsang (changes requested):

src/editors/subscription/subscriber-ied-list.ts
```diff
@@ -299,6 +303,47 @@ export class SubscriberIEDList extends LitElement {
...
+    const inputsMap: Record<string, Element> = {};
+  
+    // save clone Inputs with identity to map, remove extRef from cloned inputs
+    for (const extRefDeleteAction of extRefDeleteActions) {
+      const extRef = <Element>extRefDeleteAction.old.element;
+      const inputs = <Element>extRefDeleteAction.old.parent;
+
+      const id = identity(inputs);
+  
+      if (!inputsMap[id]) inputsMap[id] = inputs;
```
JakobVogelsang, line 327: The inputs used here are no copies. Don't we remove the ExtRef forever and short circuit the action events? Or in other works undo would not work, isn't it?

JakobVogelsang (commented):
I in my opinion still not working is should. Or do I miss something?

Flurb:
@JakobVogelsang 

With the current solution, Inputs element is removed when no ExtRefs are available anymore.
When there are ExtRefs left, the Inputs element is not removed.

I do see that undo / redo doesn't work, but the same applies to the subscribing action.
Does using cloneNode fix this?

Flurb:
Now works with `cloneNode()` :)

JakobVogelsang (approved):
LGTM

PR #590, opened 2022-03-09
fix(editors/template): on id attribute update update references as well

JakobVogelsang:


src/translations/en.ts
```diff
@@ -306,6 +309,9 @@ export const en = {
         edit: 'Edit DAType',
       },
     },
+    action: {
+      edit: 'Edited DAType ID "{{oldId}}" and its DA references to {{newId}} ',
```
ca-d, line 313: Maybe use something like "changed ... to ..." instead of "Edited ... to ..." in these action translations?

ca-d (approved):
LGTM apart from one translation suggestion.

src/translations/en.ts
JakobVogelsang: Thanks that is a good hint

PR #594, opened 2022-03-11
fix(action-pane): adjust css rules for icon slot

### LINKED ISSUE #584: Icons should be aligned correctly in action-pane
When using the icon slot in action panes like this:

```
<action-pane>
      <mwc-icon slot="icon">developer_board</mwc-icon>
</action-pane>
```

The icon is not correctly aligned with the header text, see the following screenshot:

Schermafbeelding 2022-03-07 om 09 36 25

JakobVogelsang:
Close #584


ca-d (approved):
LGTM

PR #596, opened 2022-03-14
feat(wizard-dialog): wizard-dialog content definition through `WizardInput` objects

JakobVogelsang:
Hi all, finally the PR that will extend the `Wizard` API to allow content definition through `WizardInput` objects definition instead of `TemplateResult`. 

The advantage is that a plug-in author in JavaScript can trigger a wizard-dialog in without loading lit-element. 
The disadvantage is that with `WizardInput` you can only define `Select`, `TextField` and `Checkbox` kind inputs. With `TemplateResult` we can add list, buttons .... basically all kind of web-components. This is why we added the WizardInput content definition on top of the old API.


test/unit/wizard-dialog.test.ts
```diff
@@ -258,4 +262,371 @@ describe('wizard-dialog', () => {
...
+
+      it('makes sure to not render nullSwitch with missing/false nullable', () =>
+        expect(select.nullSwitch).to.not.exist);
+
+      it('makes sure that wizard-select has correct label', () =>
+        expect(select.label).to.be.equal('myLabel'));
+
+      it('makes sure that wizard-select has correct maybeValue', async () =>
+        expect(select.maybeValue).to.equal('multi1'));
+    });
```
dlabordus, line 389: Should there be a test to see that there are 2 options available for the "values" passed. (There are 2 'mwc-list-item' under 'wizard-select')

src/foundation.ts
```diff
@@ -171,7 +171,7 @@ export function newActionEvent<T extends EditorAction>(
 
 export const wizardInputSelector =
   'wizard-textfield, mwc-textfield, ace-editor, mwc-select,wizard-select, wizard-checkbox';
-export type WizardInput =
+export type WizardInputElement =
```
dlabordus, line 174: Isn't the WizardCheckbox missing here?

dlabordus (changes requested):

test/unit/wizard-dialog.test.ts
JakobVogelsang: I think this is covered with the snapshot test above. Here the list-items reflect the selection.

src/foundation.ts
JakobVogelsang: True it is missing I will create an extra PR that points to this on. For now I think for others easier to review, if not mixed with the changes here :) We can keep it open until the PR is merged.

dlabordus (approved):

test/unit/wizard-dialog.test.ts
```diff
@@ -258,4 +262,371 @@ describe('wizard-dialog', () => {
       after(() => localStorage.removeItem('mode'));
     });
   });
+
+  describe('with content definition throught WizardInput s', () => {
```
danyill, line 266:
```suggestion
  describe('with content definition throught WizardInputs', () => {
```

I'm better with spelling than code.

test/unit/wizard-dialog.test.ts
```diff
@@ -258,4 +262,371 @@ describe('wizard-dialog', () => {
...
+    describe('for a specific WizardInputs of the kind Checkbox', () => {
```
danyill, line 267:
```suggestion
    describe('for a specific WizardInput of the kind Checkbox', () => {
```

danyill (commented):
Christopher Robbin once patted my head and said "You are only a :bear: with a very small :brain:, so let's imagine, for now, that the review process is more for your education than feedback".

On such a basis I can say I like this idea.

PR #597, opened 2022-03-14
feat(foundation): add schema attribute definition parser

JakobVogelsang:
This is the second of three PRs that is aiming to simplify wizard-dialog definition for co-developers of OpenSCD as well as plug-in contributors. The basic idea behind the function `schemaAttributeDefinition` is to get the attributes' definition for a specific element. Say you want to manipulate the `name` attribute of the element `GSEControl` in a standard way this function will tell you if there are restrictions for this attribute like pattern, minLength, maxLength its base type and or enumerations.

### DIFF
--- a/test/unit/foundation/schema/attribute.test.ts
+++ b/test/unit/foundation/schema/attribute.test.ts
@@ -1,0 +1,386 @@
+import { expect } from '@open-wc/testing';
+
+import { getSchema } from '../../../../src/schemas.js';
+import { schemaAttributeDefinition } from '../../../../src/foundation.js';
+
+const schema2007B4: XMLDocument = <XMLDocument>(
+  new DOMParser().parseFromString(
+    getSchema('2007', 'B', '4'),
+    'application/xml'
+  )
+);
+
+const schema2007B: XMLDocument = <XMLDocument>(
+  new DOMParser().parseFromString(getSchema('2007', 'B', ''), 'application/xml')
+);
+
+const schema2003: XMLDocument = <XMLDocument>(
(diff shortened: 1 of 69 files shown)


test/unit/foundation/schema/attribute.test.ts
```diff
@@ -0,0 +1,386 @@
...
+      expect(schemaAttributeDefinition('nonsence', schema2007B4)).to.be.empty);
+
+    describe('for element SampledValueControl', () => {
+      describe('and version 2003', () => {
+        const attributes = schemaAttributeDefinition(
+          'SampledValueControl',
+          schema2003
+        );
+
+        it('return the correct number of attributes ', () =>
```
danyill, line 33: There is a few of these with a space at the end you might like to remove.

```suggestion
        it('return the correct number of attributes', () =>
```

test/unit/foundation/schema/attribute.test.ts
```diff
@@ -0,0 +1,386 @@
...
+              .pattern
+          ).to.be.undefined);
+
+        it('return an optional minLength', () =>
+          expect(
+            attributes.find(attribute => attribute.name === 'name')?.restriction
+              .minLength
+          ).to.equal('1'));
+
+        it('return an otional maxLength', () =>
```
danyill, line 54:
```suggestion
        it('return an optional maxLength', () =>
```

And a few with this typo

danyill (approved):
I did not do more than read the code but I can see that this would be helpful and is required. Thankyou :+1:

JakobVogelsang:
Will be tackled within `open-scd-core` reporitory directly :)

PR #602, opened 2022-03-15
feat(wizard-dialog): allow used-defined actions in menu

JakobVogelsang:
Triggered by @danyill I was thinking how to get rid of the buttons in the wizard content. @danyill rightly so pointed out that those buttons/actions should be allocated like the primary/secondary action not in the content of the wizard. 

This mock puts the in a menu that can be open on icon-button click in the right upper corner of the wizard. You can test it working with `EnumType`s in the template editor: 

1. go to Template editor
2. go to EnumType list
3. click on any of the list items
4. see the more vert icon button on the right upper corner
5. click and try the behaviour
6. compare to e.g. DAType list -> wizard

@Sander3003 @Flurb @dlabordus @ca-d @danyill I would be curious about your take on it. If this is a viable option, I would get rid of the buttons in wizard content

### DIFF
--- a/src/foundation.ts
+++ b/src/foundation.ts
@@ -225,6 +225,16 @@ export function getMultiplier(input: WizardInput): string | null {
   else return null;
 }
 
+/** @returns [[`WizardAction`]]s to dispatch on [[`WizardDialog`]] menu action. */
+export type WizardMenuActor = () => WizardAction[];
+
+/** User interactions rendered in the wizard-dialog menu */
+interface MenuAction {
+  label: string;
+  icon?: string;
+  action: WizardMenuActor;
+}
+
 /** Represents a page of a wizard dialog */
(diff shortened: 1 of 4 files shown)


Sander3003:
As discussed today, the main concern is that users need to click an extra time and/or that new users don't know what can be changed. These are UI related things. However, form an architecture point of view, this gives more flexibility. I'm in favour, if needed we can change the UI later and/or build seperate plug-ins for functions are used a lot.

JakobVogelsang:
This is the cleaned up implementation for the `menuActions` in `wizard-dialog`. This is the feature branch where step by step all wizard-dialog using buttons in the content will be refactored. If you want to see an example, see ...

dlabordus (approved):

Flurb (approved):

PR #603, opened 2022-03-15
feat(editors/template/lnodetype-wizards): make use of -7-420 NSD

### LINKED ISSUE #542: Update 7-420 NSD file
As an engineer
I want to use the lastest DER (7-420) NSD files to configure substations
So I can use the improvements

Background:
https://blog.nettedautomation.com/2021/11/four-additional-light-namespace.html

JakobVogelsang:
Closes #542 

I did ad the NSD for -7-420 and used it in the template editor to make a preselection. I did adopt the template validator. That is potentially a more time-consuming one :). @Sander3003 could you have a lot if this is what you wanted already?


Flurb:
@JakobVogelsang I think the issue also covers the new updates of the existing NSD files we are using (7-2, 7-3 and 7-4. 8-1 is not updated). According to the Supported Documents site, new versions are available for them as well (2007B4).

See https://www.iec.ch/dyn/www/f?p=103:227:0::::FSP_ORG_ID,FSP_LANG_ID:1273,25

JakobVogelsang:
Do you need those in parallel, or is the newest version fine? Just out of curiosity, what is the difference between the two versions?

Flurb (approved):

Sander3003:
@JakobVogelsang the latest version is fine: IEC 61850-7-420:2021 and IEC 61850-7-4:2010+A1:2020

I'm not shure, what I understand is that some LN's are being removed.

Sander3003:
I created a seperate issue for the template validator: https://github.com/openscd/open-scd/issues/617

PR #632, opened 2022-03-22
feat(wizards/reportcontrol): add copy to other IEDs 

JakobVogelsang:


src/wizards/reportcontrol.ts
```diff
@@ -400,6 +405,91 @@ function getRptEnabledAction(
...
+        )
+      )
+        return [];
+
+      // clone DataSet and make sure that FCDA is valid in ied
+      const sinkDataSet = <Element>sourceDataSet?.cloneNode(true);
+      Array.from(sinkDataSet.querySelectorAll('FCDA')).forEach(fcda => {
+        if (!existFcdaReference(fcda, ied)) sinkDataSet.removeChild(fcda);
+      });
+      if (sinkDataSet.children.length === 0) return []; // when no data left no copy needed
```
danyill, line 451: Should we convey to the user if the dataset could not be transferred in its entirety?

I know it's off-topic but I keep having this reaction: I feel like most actions should be able to indicate success ("Successfully removed x items", "Copied control to Y IEDs in full", "Copied Report Control to 5/6 IEDs successfully, check the log for details" etc.) using a snackbar without being overridden by the validator. I'd like to see the validator get a little number in red for the validation errors or something like this and possibly see a little animation on it rather than validation errors taking up such a prominent position in the bottom center of the screen.

src/wizards/reportcontrol.ts
```diff
@@ -400,6 +405,91 @@ function getRptEnabledAction(
...
+
+      const sinkReportControl = <Element>element.cloneNode(true);
+
+      const source = element.closest('IED')?.getAttribute('name');
+      const sink = ied.getAttribute('name');
+
+      complexActions.push({
+        title: `ReportControl copied from ${source} to ${sink}`,
```
danyill, line 459: I can multiselect, so I thought I was copying to multiple IEDs but perhaps I wasn't and it was just the last in the selection? Multiple IEDs seems like a nice feature to have.

test/unit/wizards/reportcontrol.test.ts
```diff
@@ -687,4 +688,54 @@ describe('Wizards for SCL ReportControl element', () => {
...
+
+    describe('with a sink IED not meeting partially the data references', () => {
+      beforeEach(async () => {
+        iedsPicker.paths = [['IED: IED5']];
+        primaryAction.click();
+
+        await element.requestUpdate();
+      });
+
+      it('does copy the control block ', () =>
```
danyill, line 737:
```suggestion
      it('does copy the control block', () =>
```

test/unit/wizards/reportcontrol.test.ts
```diff
@@ -687,4 +688,54 @@ describe('Wizards for SCL ReportControl element', () => {
...
+
+    describe('with missing sink IED', () => {
+      beforeEach(async () => {
+        iedsPicker.paths = [['IED: IED20']];
+        primaryAction.click();
+
+        await element.requestUpdate();
+      });
+
+      it('does not copy the control block ', () =>
```
danyill, line 725:
```suggestion
      it('does not copy the control block', () =>
```

Linting is my greatest skill :stuck_out_tongue_winking_eye:

danyill (commented):
I had a play with this and think it is quite a valuable feature.
I have suggested a few changes.
Thank you for the opportunity to review this.

Here is a few more comments from playing with the feature:

One of the difficulties in engineering using SCL files is "how to fix mistakes which imply you need to redo something and rescale across a substation". 

I haven't checked, but I'm hoping this feature would allow me to "correct" some other devices (imagine a line bay with 5 bays, I found an error in one report, now I want to scale it to all five).

It would be helpful if you could see in the IED list other IEDs which already had a report with an identical name to facilitate this.

When I try to copy to an IED where it wouldn't make sense, I am a little frustrated (not being a very clever user) by the button just refusing to do something. If it's multiple IEDs. In the attached file this can be done if one starts with `SEL_411L_1_Test`and ReportControl `BRepo01` and attempts to copy to all three IEDs in this file (or any of them). This very likely a rightly invalid thing but the UI is not providing adequate feedback to the user IMHO.

Also, I note that the originating IED is included in the list which doesn't seem correct.

Finally, the title:

image

I think should be "Select Destination IEDs" or similar.

Oh I can't resist. Almost always when OpenSCD calls a button "Cancel" it would be better named "Back", whether or not there are nested dialogs because while in the Wizard architecture is cancellation, from the UI point of view it is returning to the base Editor or application. Cancel sounds more severe (I cancel an ongoing operation, my fiancee doesn't like me to cancel my wedding, but we don't mind going back to a happier place in time). WDYT?

src/wizards/reportcontrol.ts
```diff
@@ -400,6 +405,91 @@ function getRptEnabledAction(
...
+
+export function reportControlCopyToIedSelector(element: Element): Wizard {
+  const doc = element.ownerDocument;
+
+  return [
+    {
+      title: get('report.wizard.location'),
+      primary: {
+        icon: 'save',
+        label: get('save'),
```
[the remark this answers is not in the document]
JakobVogelsang: I do agree. I think this should be somethink like `proceed` and the `cancel` should be `close`. This is the most correct explanation for what is happenning.

src/wizards/foundation/finder.ts
```diff
@@ -48,6 +48,16 @@ export function iEDPicker(doc: XMLDocument): TemplateResult {
   ></finder-list>`;
 }
 
+export function iEDsPicker(doc: XMLDocument): TemplateResult {
+  return html`<finder-list
+    multi
+    path="${JSON.stringify(['SCL: '])}"
+    .read=${getReader(doc.querySelector('SCL')!, getIED)}
```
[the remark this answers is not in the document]
JakobVogelsang: This is the parent tag of the `IED`s elements. You can give `finder-list` a reader function where you can define how to navigate between parent and children. The function we have defined is expecting path element defined as this:

`tagName: identity string`

src/wizards/reportcontrol.ts
JakobVogelsang: You are right, we wanted to animate the log button on a long run and indicate the number of errors and warning separately. Would that be something to go with?

JakobVogelsang: This is out of scope with this issue, however. We should start a discussion how we want to inform the user that something happened he has to have an eye on. The way we are doing it now works fine so long in projects with little errors, which ultimately I hope we as a community we come to. Loading files with a lot of issues, you are blind to new issues...that is an issue. I would suggest to start an issue or a discussion regarding this topic.

src/wizards/reportcontrol.ts
JakobVogelsang: You do copy to multiple IED. Each copy should result in an extra log message. Is this not the case for you?

test/unit/wizards/reportcontrol.test.ts
JakobVogelsang: :)

src/wizards/reportcontrol.ts
danyill: Hmmm, perhaps I have erred. The most recent time I definitely got this.

danyill:
What do you think of having the IEDs we copy to as a filtered list so we can choose the manufacturer/type/name as filtering options?

JakobVogelsang:
This is what I am working on. Additionally, this allows to disable item IEDs where a copy to cannot be done as well as to disable the source IED.

JakobVogelsang:
@danyill thanks you for all the good hints. The solution now is way more user-friendly, mainly because to some extent self explaining.

src/translations/de.ts
```diff
@@ -293,28 +293,28 @@ export const de: Translations = {
...
+      title: 'GOOSE-Publizierer',
     },
     subscriberIed: {
       title: 'Verbunden mit {{ selected }}',
       subscribed: 'Verbunden',
       availableToSubscribe: 'Kann verbunden werden',
       partiallySubscribed: 'Teilweise verbunden',
-      noGooseMessageSelected: 'Keine GOOSE ausgewählt'
-    }
+      noGooseMessageSelected: 'Keine GOOSE ausgewählt',
```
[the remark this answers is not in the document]
JakobVogelsang: It was my linter. I do not think we need it from JavaScript perspective, rather a code guide.

test/integration/wizards/reportcontrol-wizarding-editing.test.ts
```diff
@@ -333,6 +335,25 @@ describe('Wizards for SCL element ReportControl', () => {
...
+      const copyMenuAction = <HTMLElement>(
+        Array.from(
+          element.wizardUI.dialog!.querySelectorAll<ListItemBase>(
+            'mwc-menu > mwc-list-item'
+          )
+        ).find(item => item.innerHTML.includes(`[controlblock.label.copy]`))
+      );
+      await element.wizardUI.dialog?.requestUpdate();
+      copyMenuAction.click();
+      await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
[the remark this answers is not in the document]
JakobVogelsang: Wait, sorry this is necessary. We need to await an animation frame as the menuAction click opens a new wizard

Flurb (approved):

PR #638, opened 2022-03-25
feat(editors/subscriber): use filtered lists

### LINKED ISSUE #559: Use filtered-list in Subscription plugin instead of mwc-list
If we have a lot of IEDs, filtering would be nice!

### LINKED ISSUE #629: No message when file has no GOOSE / SMV
It's better to show a message when there are no messages available.

Right now it's like this:

Schermafbeelding 2022-03-21 om 13 38 36

JakobVogelsang:
Close #559 Close #629

@Flurb I would especially appreciate a focused look at the tests. I did only adopt your but did not add any.


Flurb:
@JakobVogelsang The IED names are persistent, they don't disappear when searching for a specific GOOSE message.
Also, searching for a specific IED is not possible.

Is this what we want? I think it's also handy to search for a specific IED :)

JakobVogelsang:
@Flurb I had a hard time thinking about it as well. I left it like that because I was thinking that elements potentially carrying buttons - e.g. non-interactive IEDs having add GSEControl button - should not be filtered. I tried it with files having 40 IEDs and it did not bother me so far :)

A perfect solution to this would be to adapt the filtered-list and add a property to it. Something like filterNoninteractive that per default is false. This way we can filter for both the GOOSE and IEDs. Would that be a feasible solution to you?

Flurb:
@JakobVogelsang And do we want the same functionality for the Sampled Values plugin?

Sander3003:
> @JakobVogelsang And do we want the same functionality for the Sampled Values plugin? (Probably separate issue)

I would say yes, indeed a seperate issue

JakobVogelsang:
Hi @Sander3003 @Flurb,

I would say that this refactors should be done for the other plugin. Even more so when - from my perspective - both should be part of one plugin. I opened an issue #653 covering the missing review comment you had @Flurb

Flurb (approved):

Sander3003:
@Flurb can you file an seperate issue of the sampled value plug-in?

Flurb:
#656 has been created!

PR #654, opened 2022-03-31
feat(wizards/gsecontrol): add create wizards

JakobVogelsang:
Create wizards for the `GSEControl` elements. Some implementation details:

When adding a `GSEControl` a `GSE` element must be added to the `Communication` section as well. This is true only if there is a `ConnectedAP` for the `AccessPoint` where the `Server` resides in. If there is no `ConnectedAP` no `GSE` element must be added. The way it is implemented now: when `ConnectedAP` is present the second page of the create wizard allows to specify the `GSE` element. Otherwise this page is not rendered and the create wizard consists of only two wizard pages. To test delete all `SubNetwork` elements in the Communication plugin.

### DIFF
--- a/src/wizards/foundation/scl.ts
+++ b/src/wizards/foundation/scl.ts
@@ -1,0 +1,54 @@
+function incrementMac(oldMac: string): string {
+  const mac = oldMac.split('-').join('');
+  //destination MAC in IEC61850 always starts with 01:0C:CD:...
+  const newMac = '0' + (parseInt(mac, 16) + 1).toString(16).toUpperCase();
+  return newMac.match(/.{1,2}/g)!.join('-');
+}
+
+export function uniqueMacAddress(
+  doc: XMLDocument,
+  serviceType: 'SMV' | 'GOOSE'
+): string | null {
+  const maxMac =
+    serviceType === 'GOOSE' ? '01-0C-CD-01-01-FF' : '01-0C-CD-04-01-FF';
+  const startMac =
+    serviceType === 'GOOSE' ? '01-0C-CD-01-00-00' : '01-0C-CD-04-00-00';
+
+  const givenMacs = Array.from(doc.querySelectorAll('Address > P'))
+    .filter(pElement => pElement.getAttribute('type') === 'MAC-Address')
+    .map(mac => mac.innerHTML.trim());
+
+  let mac = startMac;
+  while (mac !== maxMac) {
+    if (!givenMacs.includes(mac)) return mac;
+    mac = incrementMac(mac);
+  }
+
+  return givenMacs.includes(mac) ? null : mac;
+}
+
+function incrementAppId(oldAppId: string): string {
+  return (parseInt(oldAppId, 16) + 1)
+    .toString(16)
+    .toUpperCase()
+    .padStart(4, '0');
+}
+
+export function uniqueAppId(doc: XMLDocument): string | null {
+  const maxAppId = 'FFFF';
+  const startAppId = '0001';
+
+  const givenAppIds = Array.from(doc.querySelectorAll('Address > P'))
+    .filter(pElement => pElement.getAttribute('type') === 'APPID')
+    .map(mac => mac.innerHTML.trim());
+
+  if (givenAppIds.length === 0) return null;
+
+  let appId = startAppId;
+  while (appId !== maxAppId) {
+    if (!givenAppIds.includes(appId)) return appId;
+    appId = incrementAppId(appId);
+  }
+
+  return givenAppIds.includes(appId) ? null : appId;
+}
--- a/src/wizards/foundation/scl.ts
+++ b/src/wizards/foundation/scl.ts
@@ -1,3 +1,17 @@
+export function getConnectedAP(element: Element): Element | null {
+  return element.ownerDocument.querySelector(
+    `:root > Communication > SubNetwork > ConnectedAP[iedName="${element
+      .closest('IED')
+      ?.getAttribute('name')}"][apName="${element
+      .closest('AccessPoint')
+      ?.getAttribute('name')}"]`
+  );
+}
+
+export function isAccessPointConnected(element: Element): boolean {
+  return getConnectedAP(element) ? true : false;
+}
+
 function incrementMac(oldMac: string): string {
   const mac = oldMac.split('-').join('');
   //destination MAC in IEC61850 always starts with 01:0C:CD:...
--- a/test/unit/wizards/foundation/scl.test.ts
+++ b/test/unit/wizards/foundation/scl.test.ts
@@ -1,0 +1,204 @@
+import { expect } from '@open-wc/testing';
+import {
+  uniqueAppId,
+  uniqueMacAddress,
+} from '../../../../src/wizards/foundation/scl.js';
+
+function incrementMac(oldMac: string): string {
+  const mac = oldMac.split('-').join('');
+  //destination MAC in IEC61850 always starts with 01:0C:CD:...
+  const newMac = '0' + (parseInt(mac, 16) + 1).toString(16).toUpperCase();
+  return newMac.match(/.{1,2}/g)!.join('-');
+}
+
+export function createMacs(serviceType: 'SMV' | 'GOOSE'): string[] {
+  const maxMac =
+    serviceType === 'GOOSE' ? '01-0C-CD-01-01-FF' : '01-0C-CD-04-01-FF';
+  const startMac =
+    serviceType === 'GOOSE' ? '01-0C-CD-01-00-00' : '01-0C-CD-04-00-00';
+
+  const macs: string[] = [];
+
+  let mac = startMac;
+  while (mac !== maxMac) {
+    macs.push(mac);
+    mac = incrementMac(mac);
+  }
+
+  macs.push(maxMac);
+
+  return macs;
+}
+
+function incrementAppId(oldAppId: string): string {
+  return (parseInt(oldAppId, 16) + 1)
+    .toString(16)
+    .toUpperCase()
+    .padStart(4, '0');
+}
+
+export function createAppIds(): string[] {
+  const maxAppId = 'FFFF';
+  const startAppId = '0001';
+
+  const appIds: string[] = [];
+
+  let appId = startAppId;
+  while (appId !== maxAppId) {
+    appIds.push(appId);
+    appId = incrementAppId(appId);
+  }
+
+  appIds.push(maxAppId);
+
+  return appIds;
+}
+
+describe('SCL specific functions', () => {
+  describe('define a function to get unique GSE MAC addres', () => {
+    describe('with all MAC address in use', () => {
+      const gseElementString = `<SubNetwork>${createMacs('GOOSE').map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNEtwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return null with no unique MAC', () =>
+        expect(uniqueMacAddress(doc, 'GOOSE')).to.be.null);
+    });
+
+    describe('with available MAC address', () => {
+      const macs = createMacs('GOOSE');
+      macs.splice(4, 5);
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNEtwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique MAC', () =>
+        expect(uniqueMacAddress(doc, 'GOOSE')).to.equal('01-0C-CD-01-00-04'));
+    });
+
+    describe('with one available MAC address', () => {
+      const macs = createMacs('GOOSE');
+      macs.pop();
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNetwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique MAC', () =>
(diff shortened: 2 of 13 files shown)


test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+} from '../../../../src/wizards/foundation/scl.js';
+
+function incrementMac(oldMac: string): string {
+  const mac = oldMac.split('-').join('');
+  //destination MAC in IEC61850 always starts with 01:0C:CD:...
+  const newMac = '0' + (parseInt(mac, 16) + 1).toString(16).toUpperCase();
+  return newMac.match(/.{1,2}/g)!.join('-');
+}
+
+export function createMacs(serviceType: 'SMV' | 'GOOSE'): string[] {
```
Flurb, line 14: `export`s are not needed here, because all the functions are only used within this file.

src/wizards/foundation/scl.ts
```diff
@@ -0,0 +1,68 @@
...
+  const newMac = '0' + (parseInt(mac, 16) + 1).toString(16).toUpperCase();
+  return newMac.match(/.{1,2}/g)!.join('-');
+}
+
+export function uniqueMacAddress(
+  doc: XMLDocument,
+  serviceType: 'SMV' | 'GOOSE'
+): string | null {
+  const maxMac =
+    serviceType === 'GOOSE' ? '01-0C-CD-01-01-FF' : '01-0C-CD-04-01-FF';
```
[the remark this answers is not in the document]
JakobVogelsang: Yes, those restrictions are defined in the IEC 61850:
- GSE destination MAC address: 01-0C-CD-01-00-00 until 01-0C-CD-01-01-FF
- SMV destination MAC address: 01-0C-CD-04-00-00 until 01-0C-CD-04-01-FF

test/unit/wizards/foundation/scl.test.ts
JakobVogelsang: True. Thanks for the hint!

src/wizards/foundation/scl.ts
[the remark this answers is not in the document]
danyill: Just to note that these are the fixed ranges which are correct according to the standard. Some vendor tools are configurable outside this range (the standard only allows 512 values for SMV/GOOSE but some utilities use this as a mechanism for traffic segmentation which is difficult with some a small address range).

danyill:
Thank you for this. It is a joy for me to see a new world unfold in front of me :ringed_planet: 

I'm just playing with the UI and (surprise :tada:) have not looked at any code.

In no particular order, just as I play.

The VLAN-ID appears to be incorrectly validated I expect to be able to fit a 12-bit value in.

image

Investigating this further I think the validation criteria are in hexadecimal but we have not indicated this is a hexadecimal field. If the validation criteria are decimal then they require modification.

FWIW, I recommend we use decimal throughout OpenSCD.

When I hover over an illegal entry, I get "Please match the format requested". How can the user achieve this without a tooltip or more information? Same for VLAN-PRIORITY.

On at least the MAC-address I think we should disable spell checking on the input field:

image

If I delete the value, it's not easy to figure out how to get it back unless your memory of the prefix is quite good. If it is fixed prefix then I think the user should only be able to edit the suffix.

If we do it the way we are doing it now, then we don't appear to be validating that the MAC address range is correct for GOOSE. Perhaps we need a button for a "default" or a tooltip with the lower bound value.

I note that the standard says a number of vague things, including (IEC 61850-8-2 Ed 2.1 Annex B) that:

* "The value of `00-00-00-00-00-00` shall be used to indicate that the multicast address has
not been configured"
* "It is recommended that the second and third octets have a value of 0C-CD." Is it recommended or mandatory?? Srsly ! I think it is mandatory and this is the approach we should take. I think this is why some tools allow modification of this. Perhaps this means any validation criteria should be quite wide.

The APPID is also not letting me away with "Shall be 4 characters. Characters shall be limited to 0 to 9 and A to F."

image

Generally there seem to be some problems with the implementation of IEC 61850-8-1 Ed 2.1, section 25.3.2. We need to ensure the criteria are adjusted for decimal or hexadecimal entry.

Perhaps on screens like this, we need an overall reset button or a way to reset individual fields. I went back by clicking "ADD GSECONTROL", and then went forward to "ADD GSE" but it had my entries on the screen (which to be fair I would normally complain if it did otherwise :wink:). But that means at present with a validation error if I am not a clever user I can get stuck (unless I cancel and start again).

While I'm on this topic, is it possible for us to have a pluggable helper for items like VLAN-ID, VLAN-Priority and MAC-Address? I'd like to heuristically assign these and have a plugin that would either automate it or ask the user some questions and allocate it correctly.

The two different entry points to the screen behave differently.

* It looks correct if I click on an IED and then selected "GOOSE +".
* But if I select the "G" icon and click on "GOOSE +" then I get a "Select Report Control Location". To make this appropriately generic it should be a "Select IED Location" or "Choose IED" or similar.

When I add a GSEControl, the appId is empty by default. I think we should choose a default value so the dialog is not blocking if I am just "clicking my way through" (this is a bit pernickety). 

I have no idea, what "Add XMLSchema-instance type" means:

image

Is this something the user should configure with each GOOSE (I have not seen this in other applications) or should it be a global setting or configuration option? I think many end-users would be confused by this (you should not need to know a great deal about XML schemas to configure a GOOSE message). If it is needed should it be the first entry field? I was stymied.

danyill (commented):
I have made a few comments... thank you for this opportunity to provide feedback. :wink:

danyill:
> If there is no ConnectedAP no GSE element must be added. The way it is implemented now: when ConnectedAP is present the second page of the create wizard allows to specify the GSE element. Otherwise this page is not rendered and the create wizard consists of only two wizard pages. To test delete all SubNetwork elements in the Communication plugin.

Hmmm, "As a user if I want to make a GSE I want to make a GSE and I do not expect to have my options curtailed on my path to greatness". Should we be warning the user and/or opening an AccessPoint creation wizard? Or creating a base access point and informing the user at the end of the process?

Otherwise we force the user to adopt a different workflow with many more button clicks and back and forth. Creating a GSE should not be fundamentally different depending on whether an Access Point is present (some IED tools don't make this even a visible part of the configuration and it would confuse end users).

JakobVogelsang:
@danyill thank you as always for being the usability guard. This is crucially important in making this whole thing a success. For me, it is difficult to add all your comments into this PR. I have therefore filtered some points that I think are valid and must be improved before merging:
1. GSE wizard page should indicate that the server has no connection to a subnetwork instead of just dropping it under the table. I myself was not so sure about this decision. We could go about this in two ways: show the page, but show a warning message. Decide that GSE creation is the Communication editor and remove that from GSEControl edit as well. I am for the first one
2. Pattern validation. This part is clearer to me. We must allow whatever the schema allows. At least so long, we are working towards the goal to become a SCL editor. For me it is important to serve as many users as possible, including those adding some nonsense information. I am sure however we will further restrict this later, working in the non pro mode :). For now, I would suggest to add a validation message that appears always when the input does violate the pattern/limits telling what those are. Would that be ok for you?

Homework to do/separate issues
- Start discussion: Where do we want to have the setting that decides, if XMLSchema type extension shall be added to the `P` elements
- Extend wizard-textfield to deal with hex type inputs. Discussion on that already started.

danyill:
> GSE wizard page should indicate that the server has no connection to a subnetwork instead of just dropping it under the table. I myself was not so sure about this decision. We could go about this in two ways: show the page, but show a warning message. Decide that GSE creation is the Communication editor and remove that from GSEControl edit as well. I am for the first one

I guess it depends on what users are doing. 

* Today, most users start with some IED tool or file so usually there is a definition of the Communication section which is appropriate. 
* If OpenSCD allowed us to build our whole substation from scratch as a configuration tool from a blank file where we just imported the SCL from other providers then I would be more concerned (but we are some distance away from that ambitious target).

I don't think we want to require a two step process: first GSEControl then GSE. I have not seen any tool require this and it seems klunky and inefficient. 
So I'm for the first one as well.

> Pattern validation. This part is clearer to me. We must allow whatever the schema allows. At least so long, we are working towards the goal to become a SCL editor. For me it is important to serve as many users as possible, including those adding some nonsense information. I am sure however we will further restrict this later, working in the non pro mode :). For now, I would suggest to add a validation message that appears always when the input does violate the pattern/limits telling what those are. Would that be ok for you?

Agree. For now we could just make sure the UI was clear about hex or decimal inputs.

JakobVogelsang:
Alight then, I will rework the create wizard and will add a wizard page with missing warning the user about a missing Communication section connection.

src/wizards/foundation/scl.ts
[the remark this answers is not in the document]
JakobVogelsang: This code here makes a suggestion. I therefore would argue that this does fit most users. If others want to do something else they can so long it is valid acc. to the schema.

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNEtwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique MAC', () =>
```
danyill, line 85:
```suggestion
      it('return the first available first unique MAC', () =>
```

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        expect(uniqueMacAddress(doc, 'GOOSE')).to.be.null);
+    });
+
+    describe('with available MAC address', () => {
+      const macs = createMacs('GOOSE');
+      macs.splice(4, 5);
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNEtwork>`;
```
danyill, line 79:
```suggestion
      )}</SubNetwork>`;
```

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  return appIds;
+}
+
+describe('SCL specific functions', () => {
+  describe('define a function to get unique GSE MAC addres', () => {
+    describe('with all MAC address in use', () => {
+      const gseElementString = `<SubNetwork>${createMacs('GOOSE').map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNEtwork>`;
```
danyill, line 63:
```suggestion
      )}</SubNetwork>`;
```

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNetwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique MAC', () =>
```
danyill, line 149:
```suggestion
      it('return the first available unique MAC', () =>
```

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const gseElementString = `<SubNetwork>${appIds.map(
+        appId =>
+          `<ConnectedAP><Address><P type="APPID">${appId}</P></Address></ConnectedAP>`
+      )}</SubNetwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique APPID', () => {
```
danyill, line 167:
```suggestion
      it('return the first available unique APPID', () => {
```

test/unit/wizards/foundation/scl.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      const gseElementString = `<SubNetwork>${macs.map(
+        mac =>
+          `<ConnectedAP><Address><P type="MAC-Address">${mac}</P></Address></ConnectedAP>`
+      )}</SubNetwork>`;
+      const doc = new DOMParser().parseFromString(
+        gseElementString,
+        'application/xml'
+      );
+
+      it('return the first availablefirst unique MAC', () =>
```
danyill, line 133:
```suggestion
      it('return the first available unique MAC', () =>
```

danyill (commented):
Thank you for this, this is a super important feature for me. :tada: :rocket: 
I would like to be able to completely programmatically define GOOSE messages as a templating mechanism within my own extension and I think this fine addition will allow me to do it.

Some usual English language pedantry in suggestions.

I played around with this:


* When editing an already created GOOSE, when I get to the "Edit DataSet" screen and click the three dots in the top right and select, "Add Data Attributes", I can add the data attributes. When I exit this screen, I can't click "Save" (because I haven't edited anything within that dialog). This seems to be generally true for "dialogs accessed via the three dots".

* I have a small range of usability concerns like this but perhaps they will be addressed with the upcoming publisher interface?

* Often on larger datasets the Dialog extends to the bottom of the screen which means that validation errors from the snackbar are on top of the "Cancel" and "Save" buttons if there are any existing validation errors in the file (usual gripe, just saying! :wink:)

* When creating a new GOOSE message we still have a reference to Report Control:

    image

* Arguably if VLAN-ID in the GSE is enabled, the one thing it should not be set to is 0 (usually used to indicate a priority tag only) and neither should it be 1 (see [here](https://networkengineering.stackexchange.com/questions/24404/vlan-0-1-and-4095-are-reserved-what-are-they-reserved-for)). Although it seems a bit odd, may I suggest a default of 2?

    It is probably an issue for another day but it would be nice to be able to set a reference to an existing dataset.

JakobVogelsang:
@danyill what do you think about disabling `VLAN-ID` and `VLAN-PRIORITY` by default? From my experience, rarely this is set or kind of disable through using VLAN-ID 0.

danyill:
> @danyill what do you think about disabling `VLAN-ID` and `VLAN-PRIORITY` by default? From my experience, rarely this is set or kind of disable through using VLAN-ID 0.

I don't mind disabling by default. I think these are very often set but I don't think we can choose a sane, obvious default for all users.

danyill (approved):
LGTM, I have skimmed the code and mainly focussed on the user interface.

PR #660, opened 2022-04-01
fix(filtered-list): allow filter nested mwc-...-list-item s

### LINKED ISSUE #646: Filtering in Diagnostic Panel not working
When a SCL File has some validation errors these are shown in the Diagnostic Panel.
This panel has a filter to filter the list of message, but the filter isn't working anymore.

image

It looks like with some refactoring the list was changed. The lines were changed to elements "abbr".

Maybe also check something else directly. When there are no errors the entry "abbr" contains "undefined".
It's not displayed on the screen, but can cause strange behavior in the future.

image

JakobVogelsang:
Close #646 

This PR is in between fix and feat. The `filtered-list` was not designed to filter nested list items. However, in dialogs the space is so limited that we do use `abbr` like in the diagnosis dialog. 

With this PR I propose to change the filter strategy of the `filtered-list` a bit an allow to filter for any nested `mwc-...-list-item`.

### DIFF
--- a/src/filtered-list.ts
+++ b/src/filtered-list.ts
@@ -16,7 +16,28 @@ import '@material/mwc-textfield';
 import { CheckListItem } from '@material/mwc-list/mwc-check-list-item';
 import { List } from '@material/mwc-list';
 import { ListBase } from '@material/mwc-list/mwc-list-base';
+import { ListItemBase } from '@material/mwc-list/mwc-list-item-base';
 import { TextField } from '@material/mwc-textfield';
+
+function slotItem(item: Element): Element {
+  if (!item.closest('filtered-list') || !item.parentElement) return item;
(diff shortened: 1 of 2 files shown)


Flurb:
When I'm filtering, I see the following grey bar appearing:
Schermafbeelding 2022-04-03 om 21 26 26

After checking the developer console, I think it's a bunch of `li` elements used as a divider.
They are not deleted when filtering, I think this should be fixed :)

JakobVogelsang:
Sorry for that one!! I will fix it

JakobVogelsang:
@Flurb I did break the filtered-list refactor into two PRs. This one is dealing with filtering on nested elements only. The other one will deal with filtering on noninteractive items

dlabordus (approved):

Issue #658, opened 2022-04-01
Styling issues Subscription Editor Plugin


ca-d:
There are several styling issues in the subscription editor:

- The gap between the list containers widens at higher resolutions.
- At lower resolutions the gap becomes negative and the lists overlap (see below).
- At an even lower resolution the right list container is moved below the left one without the containers becoming wider, leading to an empty right half of the screen (see below).
- The list containers have a fixed height that is larger than the available content space. This makes the entire window scroll unnecessarily, adding a scroll bar to the right.
- The list containers additionally have their own scroll bars, even if they're short enough that no scrolling is needed.
- The click targets for focusing the list containers (green border) are somewhat inconsistent, only clicking on elements inside the filtered-list will give focus to the container, not clicking the header or empty space below the filtered-list (which is there due to the fixed height)
- The "bold" font weight on the header list items looks slightly janky, since in other places we only use the three officially recommended material design font weights "Light", "Regular" and "Medium" (css values 300, 400, and 500, respectively) and therefore only include these three weights in https://openscd.github.io/public/google/fonts/roboto-v27.css . The "bold" typeface used here is therefore a version of one of these fonts artificially embiggened by the browser.

> *overlapping list containers*
> ![Screenshot from 2022-04-01 12-22-23](https://user-images.githubusercontent.com/20446628/161253076-5f613704-ecd5-4b6c-8712-ef37b2a6feb9.png)
>
> *missing list container*
> ![Screenshot from 2022-04-01 12-24-14](https://user-images.githubusercontent.com/20446628/161253192-2c2d02e3-880b-418e-92a3-18cfc70a1307.png)

PR #661, opened 2022-04-01
fix(editors/subscription): several style issues

JakobVogelsang:
This is a suggestion to deal with some styling issues #658 reported by @ca-d. To be more precise

**Fixing**:
 - The gap between the list containers widens at higher resolutions.
 - At lower resolutions, the gap becomes negative, and the lists overlap (see below).
 - At an even lower resolution, the right list container is moved below the left one without the containers becoming wider, leading to an empty right half of the screen (see below).
 - The list containers have a fixed height that is larger than the available content space. This makes the entire window scroll unnecessarily, adding a scroll bar to the right.
 - The click targets for focusing the list containers (green border) are somewhat inconsistent, only clicking on elements inside the filtered-list will give focus to the container, not clicking the header or empty space below the filtered-list (which is there due to the fixed height)

**Maybe this**:
 - The list containers additionally have their own scroll bars, even if they're short enough that no scrolling is needed.

**Not this**(I did not have an idea how this could be solved the best)
- The "bold" font weight on the header list items looks slightly janky, since in other places we only use the three officially recommended material design font weights "Light", "Regular" and "Medium" (css values 300, 400, and 500, respectively) and therefore only include these three weights in https://openscd.github.io/public/google/fonts/roboto-v27.css . The "bold" typeface used here is therefore a version of one of these fonts artificially embiggened by the browser.

### DIFF
--- a/src/editors/Subscription.ts
+++ b/src/editors/Subscription.ts
@@ -4,6 +4,7 @@ import '@material/mwc-fab';
 
 import './subscription/subscriber-ied-list-goose.js';
 import './subscription/publisher-goose-list.js';
+import { styles } from './subscription/foundation.js';
 
 /** An editor [[`plugin`]] for subscribing IEDs to GOOSE messages using the ABB subscription method. */
 export default class SubscriptionABBPlugin extends LitElement {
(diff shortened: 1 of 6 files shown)


Flurb:
Maybe (if it's not too much work), make the header text of both lists floated.
This way it's always visible, even if you're on the bottom of the list and you always now which GOOSE is selected for example (for the subscriber list at least)

ca-d:
The bold font fix is just `font-weight: 500` instead of `font-weight: bold` in the `span`'s css. Sorry for phrasing it in such a roundabout way above :sweat_smile:

JakobVogelsang:
@ca-d Thanks!
@Flurb I would separate this into an extra issue. This here should stay a pure style fix. The other one would be a feat or refactor

JakobVogelsang:
@ca-d @Flurb I have added a commit font-weight for `noninteractive` list-items. For me it is good enough to be merged.

Flurb (approved):

PR #679, opened 2022-04-06
Filter noninteractives

JakobVogelsang:
This PR is not a draft but a pure prototype. I was plying around with the `filtered-list` and was thinking about a possibility to filter on something like sections. The following implementation is a suggestion as there are still a tone of open questions: 
1. What means checked filter an element or keep an element
2. Do we need everything checked/everything unchecked by default
 
The implementation has some technical limits:
- the menu does allow multiple checked items, but does not allow checking multiple items before closing. This means that one have to open the menu quite often to have said half of the element selected. 

I pushed this PR to get a discussion started: What must be implemented in a general filter function and does it make sense to add it to the `filtered-list`. I have the feeling that the code would become very messy and using such a component would become quite complex. In opposition to that the logic behind `filtered-list` is very simple and thus also limited.

@ca-d @Sander3003 @danyill @dlabordus @Flurb what are your thoughts on that?

### DIFF
--- a/src/editors/subscription/publisher-goose-list.ts
+++ b/src/editors/subscription/publisher-goose-list.ts
@@ -64,11 +64,11 @@ export class PublisherGOOSEList extends LitElement {
   render(): TemplateResult {
     return html` <section tabindex="0">
       <h1>${translate('subscription.publisherGoose.title')}</h1>
-      <filtered-list>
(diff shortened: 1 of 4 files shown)


ca-d:
I'm not sure I quite understand the use case for the particular filters you implemented.

PR #682, opened 2022-04-08
feat(expandable-list-item): container adding expand/collapse to list item

JakobVogelsang:
The web-component proposed here is a CSS container with expanding/collapsing function. It is working with slots and therefore allows adding all kind of list items. Theoretically, It can be used to build a tree. 

If you want to test exchange the `renderGoose` and `render` methods in `editors/subscriber/publisher-goose-list` with:

```
renderGoose(element: Element): TemplateResult {
    return html`<mwc-list-item
        twoline
        slot="child"
        @click=${() => this.onGooseSelect(element)}
        graphic="large"
      >
        <span>${element.getAttribute('name')}</span>
        <span slot="secondary">${identity(element)}</span>
        <mwc-icon slot="graphic">${gooseIcon}</mwc-icon>
      </mwc-list-item></extendable-list-item
    >`;
  }

  render(): TemplateResult {
    return html` <section tabindex="0">
      <h1>${translate('subscription.publisherGoose.title')}</h1>
      <filtered-list>
        ${this.ieds.map(
          ied =>
            html`
              <expandable-list-item ?defaultExtended=${true}>
                <mwc-list-item slot="parent" graphic="icon">
                  <span>${getNameAttribute(ied)}</span>
                  <mwc-icon slot="graphic">developer_board</mwc-icon>
                </mwc-list-item>
                ${this.getGSEControls(ied).map(control =>
                  this.renderGoose(control)
                )}
              </expandable-list-item>
            `
        )}
      </filtered-list>
    </section>`;
  }
```

There is only one thing I do not like about the implementation as is: For proper styling, I needed a `.parent` div that align the parent list item with the toggle button. This however requires reflecting the toggle buttons `on` attribute to the `.parent` div that then does this 

```
.parent:not([on]) ~ .content mwc-list {
   display: none;
}
```

The method `onExpandToggle`  that does the reflection, is ugly although it does what it has to do. 

I am curious about your take on it :)

### DIFF
--- a/test/unit/__snapshots__/expandable-list-item.test.snap.js
+++ b/test/unit/__snapshots__/expandable-list-item.test.snap.js
@@ -4,7 +4,7 @@ export const snapshots = {};
 snapshots["expandable/collapsable list item container with missing parent and children slotted looks like the latest snapshot"] = 
 `<div class="container">
   <div class="parent">
-    <div class="extendContainer">
+    <div class="expandIconContainer">
       <mwc-icon-button-toggle
         }}=""
         class="hidden"
@@ -29,7 +29,7 @@ snapshots["expandable/collapsable list item container with missing parent and ch
 snapshots["expandable/collapsable list item container with only parent item slotted looks like the latest snapshot"] = 
 `<div class="container">
   <div class="parent">
-    <div class="extendContainer">
+    <div class="expandIconContainer">
       <mwc-icon-button-toggle
         }}=""
         class="hidden"
@@ -54,7 +54,7 @@ snapshots["expandable/collapsable list item container with only parent item slot
 snapshots["expandable/collapsable list item container with both parent and childen slotted and defaultExpanded not set looks like the latest snapshot"] = 
 `<div class="container">
   <div class="parent">
-    <div class="extendContainer">
+    <div class="expandIconContainer">
       <mwc-icon-button-toggle
         }}=""
         officon="keyboard_arrow_down"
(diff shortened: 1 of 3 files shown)


test/unit/__snapshots__/expandable-list-item.test.snap.js
```diff
@@ -0,0 +1,105 @@
...
+      <slot name="child">
+      </slot>
+    </mwc-list>
+  </div>
+</div>
+`;
+/* end snapshot expandable/collapsable list item container with missing parent and children slotted looks like the latest snapshot */
+
+snapshots["expandable/collapsable list item container with only parent item slotted looks like the latest snapshot"] = 
+`<div class="container">
```
dlabordus, line 30: Is it correct that this snapshot looks the same as the previous snapshot?

Flurb: @JakobVogelsang seems like something is not correct here

JakobVogelsang:
What do you think about this component theoretically? Do you think it is useful? Or are there better ways to make an expandable list-item?

test/unit/__snapshots__/expandable-list-item.test.snap.js
JakobVogelsang: Yeah, it is not easy to see. I had to look very close too :)
- 1 and 2 look the same by design
- 3 mwc-icon-toggle-button has no class definition 
- 4 div class parent has additional attribute on

PR #687, opened 2022-04-14
feat(wizards/dai): Change DAI values

### LINKED ISSUE #639: Change DAI values
As an engineer
I want to be able to change DAI values
So I can change the values according to specifications

Button location:
image

```
							<DAI name="sboTimeout">
								<Val>
									30000
								</Val>
							</DAI>
```

Once you click on the button, the values can be changed inline. The value becomes editable in the same window (inline editing). It should be able to edit multiple DAI at the same time. A change will be stored if the checkmark/Oke (green) is clicked or reverted if the users clicks on a X (red). 

XSD values according to the IEC61850 XSD:

```
<xs:simpleType name="tPredefinedBasicTypeEnum">
        <xs:restriction base="xs:Name">

Implementation for now:
            <xs:enumeration value="BOOLEAN"/>
            <xs:enumeration value="INT8"/>
            <xs:enumeration value="INT16"/>
            <xs:enumeration value="INT24"/>
            <xs:enumeration value="INT32"/>
            <xs:enumeration value="INT64"/>
            <xs:enumeration value="INT128"/>
            <xs:enumeration value="INT8U"/>
            <xs:enumeration value="INT16U"/>
            <xs:enumeration value="INT24U"/>
            <xs:enumeration value="INT32U"/>
            <xs:enumeration value="FLOAT32"/>
            <xs:enumeration value="FLOAT64"/>
            <xs:enumeration value="VisString32"/>
            <xs:enumeration value="VisString64"/>
            <xs:enumeration value="VisString65"/>
            <xs:enumeration value="VisString129"/>
            <xs:enumeration value="VisString255"/>

For later (new stories):
            <xs:enumeration value="Enum"/>
            <xs:enumeration value="Dbpos"/>
            <xs:enumeration value="Tcmd"/>
            <xs:enumeration value="Quality"/>
            <xs:enumeration value="Timestamp"/>
            <xs:enumeration value="Struct"/>
            <xs:enumeration value="EntryTime"/>
            <xs:enumeration value="Check"/>
            <xs:enumeration value="ObjRef"/>
            <xs:enumeration value="Currency"/>
            <xs:enumeration value="PhyComAddr"/>
            <xs:enumeration value="TrgOps"/>
            <xs:enumeration value="OptFlds"/>
            <xs:enumeration value="SvOptFlds"/>
            <xs:enumeration value="LogOptFlds"/>
            <xs:enumeration value="EntryID"/>
            <xs:enumeration value="Octet6"/>
            <xs:enumeration value="Octet16"/>
            <xs:enumeration value="Octet64"/>
            <xs:enumeration value="Unicode255"/>
            <!-- for 61850-8-1 Edition 2.1 -->
        </xs:restriction>
    </xs:simpleType>

```
Regular expressions can be used to check if the user input is.

Examples:
Validation of: <xs:enumeration value="INT8"/>
The value can be +/- 2^7

Validtaion of:  <xs:enumeration value="INT8U"/>
The value can be 2^8

All values can contrains can be found in IEC 61850-7-2 6.1 and 6.2 ([link for NEN connect users](https://connect.nen.nl/Family/Detail/38669))

Only editable fields to have an edit icon.

Flurb:
Closes #639


src/editors/ied/da-container.ts
```diff
@@ -110,7 +118,19 @@ export class DAContainer extends LitElement {
           @click=${() => this.requestUpdate()}
         ></mwc-icon-button-toggle>
       </abbr>` : nothing}
-      <h6>${this.renderValue()}</h6>
+      ${this.instanceElement && getCustomField()[<DaiValidationTypes>bType] ?
+        html`<div style="display: flex; flex-direction: row;">
+          <div id="value">
```
JakobVogelsang, line 123: `id` should be used with care. I know this is used  a lot in the code base, due to pending refactors. We learned that on the way as well. `id` should be used for something unique in the dom. In this case, due to the encapsulated style within the web-compoentent `da-container` this does not have an drastic effect here. Theoretically, you could imagine a scenario where someone style an element in the light DOM and by accident uses the same `id`. This could affect another element in the light DOM using the same `id`. There are a lot of discussion online about this topic. I personally like this rule of thumb: 

```
1. IDs reference unique elements. There should only be 1 instance of any 1 ID. Classes can be applied to many elements and an element can have multiple classes
2. A CSS rule using an ID will override a CSS rule with a class if they both try to assign the same properties.
3. IDs are used for all sorts of native applications: anchor references, forms, iframes, whereas classes are used purely for styling with CSS.
```

test/testfiles/wizards/ied.scd
```diff
@@ -158,6 +158,11 @@
 						</DOI>
 					</LN>
 					<LN prefix="CB" lnClass="CSWI" inst="2" lnType="Dummy.CSWIwithoutCtlModel">
+						<DOI name="Beh">
+								<DAI name="integer">
```
JakobVogelsang, line 162: I would not use `Beh` as CDC INS. This has been replaced by `ENC` in Ed.2 and therefore is only part of Ed.1 projects. I would not recommend to use Ed.1 at all,  Ed.2.1 being published. I would certainly not use it in your test files as we say on the first page README we concentrate in Ed.2 and upwards

JakobVogelsang: Sorry, I just see you are using it with some randow DA.

test/unit/editors/ied/foundation/foundation.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+      expect(element).shadowDom.to.equalSnapshot();
+      expect(element.shadowRoot?.querySelector('input')?.value).to.eql('true');
+    });
+
+    it('renders a INT8 field correctly', async () => {
+      const value = getValue("int8test");
+      const element = await fixture(html`${getCustomField()[<DaiValidationTypes>'INT8']?.render(value!)}`);
+
+      expect(element).shadowDom.to.equalSnapshot();
```
JakobVogelsang, line 27: Missing `await` an all ...equalSnapshot()

test/unit/editors/ied/foundation/foundation.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    it('renders a VisString255 field correctly', async () => {
+      const value = getValue("visstring255test");
+      const element = await fixture(html`${getCustomField()[<DaiValidationTypes>'VisString255']?.render(value!)}`);
+
+      expect(element).shadowDom.to.equalSnapshot();
+      expect(element.shadowRoot?.querySelector('input')?.value).to.eql('deadlift');
+    });
+  });
+
+  function getValue(daiName: string) {
```
JakobVogelsang, line 160: Why not putting the function before the tests?

JakobVogelsang (changes requested):
This review is code based. I would like to play around with some files as well tomorrow

src/editors/ied/foundation/foundation.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface CustomField {
+  render(value: string): TemplateResult;
+}
+
+const daiValidationTypes = ['BOOLEAN', 'INT8', 'INT16', 'INT24', 'INT32', 'INT64',
+  'INT128', 'INT8U', 'INT16U', 'INT24U', 'INT32U', 'FLOAT32', 'FLOAT64', 'VisString32', 'VisString64',
+  'VisString65', 'VisString129', 'VisString255'] as const;
+export type DaiValidationTypes = typeof daiValidationTypes[number];
+
+export function getCustomField(): Record<DaiValidationTypes, CustomField> {
```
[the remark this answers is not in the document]
Flurb: Thank you!

test/unit/editors/ied/foundation/foundation.test.ts
Flurb: @JakobVogelsang 
Why is this? `element` is already fully loaded, and equalSnapshot is not a `Promise` or something.
Using `await` here has no effect.

Flurb: I already see why it's needed!
[Testing: Semantic Dom Diff](https://open-wc.org/docs/testing/semantic-dom-diff/)

JakobVogelsang (approved):
LGTM

PR #689, opened 2022-04-19
feat(editors/Subscription): When undo / redo actions Subscription / SMV plugin, lists are not refreshed

### LINKED ISSUE #633: When undo / redo actions Subscription / SMV plugin, lists are not refreshed
When undo / redo an action in the GOOSE or SMV plugin, the action is done but the lists are not being refreshed so it looks like the undo / redo didn't work.

Flurb:
Closes #633


src/editors/sampledvalues/foundation.ts
```diff
@@ -10,13 +10,13 @@ export enum SubscribeStatus {
...
-  dataset: Element;
+  sampledValuesControl: Element | undefined;
+  dataset: Element | undefined | null;
 }
 export type SampledValuesSelectEvent = CustomEvent<SampledValuesSelectDetail>;
 export function newSampledValuesSelectEvent(
-  sampledValuesControl: Element,
-  dataset: Element,
+  sampledValuesControl: Element | undefined,
+  dataset: Element | undefined | null,
```
dlabordus, line 19: Why need to have both undefined and null?
Is it not nicer to select one of the two and work with that, prefer undefined just like 'sampledValuesControl'.

src/editors/subscription/foundation.ts
```diff
@@ -10,13 +10,13 @@ export enum SubscribeStatus {
 }
 
 export interface GOOSESelectDetail {
-  gseControl: Element;
-  dataset: Element;
+  gseControl: Element | undefined;
+  dataset: Element | undefined | null;
```
dlabordus, line 14: Same remark as with SampledValues, do we need both?

dlabordus (approved):

Issue #485, opened 2022-01-14
Vacuum definitions function for DataTypeTemplates


danyill:
This is a feature request.

It would be useful to have a "vacuum definitions" function to remove unused `LNodeType`s or `DOType`s, `DAType`s or unused `Enum`s in the `DataTypeTemplates` section. This is the kind of thing that other XML editing software (e.g. Inkscape) has.

I think this could usefully be part of the Templates section or it could be an additional menu item (although this is growing quickly and I guess we can't just keep adding menu items for forever...).

Even just for working on open-scd constructing a test case is more laborious than necessary even when I have the right starting material because I wish to keep a "simple" test file with as little extraneous data as possible. Currently, I manually edit and modify the types.

I like the name "Vacuum definitions" but "Remove unlinked types" would be more correct :wink:

JakobVogelsang:
Thanks for this issue. This is an interesting one that I did not have on screen. It fits very nicely with another feature wish #290  posted by @BurTah00. I would love to combine those in a menu type plug-in that I would call something like clean up. When clicking on the menu button, a dialog opens where the user selects the type of clean-up. Your wish would be one of those. I don't think we need to be complete from the beginning as it can grow but start with the two issues , this one and #290.

JakobVogelsang:
@danyill would you say we can close this?

danyill:
Yes. Soon I will have a PR to improve the current state.

PR #701, opened 2022-04-24
feat(editors/cleanup) Support DataTypeTemplates

danyill:
Remove unused types. Part of #485


danyill:
I would be delighted to receive a review.

* There is not a method to assist in deletion of an IED but I could look into that (or in another PR which might be preferable).
* I have updated the test file a bit to make it more conforming to the SCL schema because the validation errors were causing irritation (and I worried they might mask other problems or misunderstandings).

I think the only tricky thing here is whether my removal of subtypes logic is correct for `SDO/BDA[bType=Struct]`

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+
+const iconMapping = {
+  EnumType: <iconType>'enumIcon',
+  DAType: <iconType>'dAIcon',
+  DOType: <iconType>'dOIcon',
+  LNodeType: <iconType>'lNIcon',
+};
+
+/** An editor component for cleaning SCL DataType templates. */
+@customElement('cleanup-datatypes')
```
JakobVogelsang, line 57: A small thing but as we had that recently this hint: `CleanupDataTypes` would translate `cleanup-data-types`. This is not consisted due to my not strict review in the past, but since this is new, I would suggest changing that.

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+  @property({ type: Boolean })
+  disableControlClean = false;
+
+  @property({ type: Array })
+  unreferencedDataTypes: Element[] = [];
+
+  @property()
+  selectedDataTypeItems: MWCListIndex | [] = [];
+
+  @query('.deleteButton')
```
JakobVogelsang, line 72: Class names are case-insesitive and it is suggested to use lower case. See for example the discussion here https://stackoverflow.com/questions/1547986/can-i-use-camelcase-in-css-class-names

test/integration/editors/cleanup/datatypes-container.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      // the correct number of DOTypes should remain (we actually remove 3 here)
+      expect(doc.querySelectorAll('DOType')).to.have.lengthOf(23);
+      expect(
+        doc.querySelectorAll(
+          'DOType[id="NotUsedDummy.SPS"], DOType[name="WYE_2_3"], DOType[id="CMV_1"]'
+        )
+      ).to.have.lengthOf(0);
+    });
+
+    // it('correctly removes two DOTypes and one referenced via an SDO from the SCL', async () => {
```
danyill, line 90: I don't recall why this is commented out. It should perhaps be removed.

test/integration/editors/cleanup/datatypes-container.test.ts
```diff
@@ -0,0 +1,289 @@
...
+        await element.cleanButton.click();
+
+        // the correct number of DOTypes should remain (we remove 2 here)
+        expect(doc.querySelectorAll('DOType')).to.have.lengthOf(24);
+        expect(
+          doc.querySelectorAll(
+            'DOType[id="NotUsedDummy.SPS"], DOType[id="WYE_2_3"]'
+          )
+        ).to.have.lengthOf(0);
+        expect(doc.querySelectorAll('DOType[id="CMV_1"]')).to.have.lengthOf(1);
```
danyill, line 187: In #814, ca-d made many changes in the form of `doc.querySelectorAll` to `element.doc.querySelectorAll`.
I didn't understand the purpose of them there but should these be changed to better scope the query?

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+   * @param usedSelector - CSS selector for SCL type's instantiated name, e.g. LN, LN0.
+   * @param keyAttributeName - attribute name for SCL types uniqueness guarantee, e.g. lnType.
+   * @param templateSelector - CSS selector for SCL template element in DataTypeTemplate section.
+   * @returns an array of unreferenced elements sorted by their identity string.
+   */
+  private getUnusedType(
+    usedSelector: string,
+    keyAttributeName: string,
+    templateSelector: string
+  ) {
```
danyill, line 324: I expect Jakob will suggest this method is too complex :wink: 
I don't mind refactoring it but the code will look very similar for LN/DO/DA.

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+          this.doc.querySelector(`DataTypeTemplates DAType[id="${bdaType}"]`)!
+        );
+      }
+    });
+    return removableItems;
+  }
+
+  /**
+   * Get items from selection list and and any subtypes.
+   * @returns An array of SCL elements representing selected items and subtypes as required .
```
danyill, line 270:
```suggestion
   * @returns An array of SCL elements representing selected items and subtypes as required.
```

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+
+    if (this.cleanSubTypesCheckbox!.checked === true) {
+      cleanItems = cleanItems.concat(
+        this.getUnusedSDOReferencedTypes(cleanItems)
+      );
+      cleanItems = cleanItems.concat(
+        this.getUnusedBDAReferencedTypes(cleanItems)
+      );
+    }
+    console.log(cleanItems)
```
danyill, line 285: Leftovers!

```suggestion
```

src/editors/Cleanup.ts
```diff
@@ -18,6 +19,7 @@ export default class Cleanup extends LitElement {
       <div class="cleanup">
         <cleanup-datasets .doc=${this.doc}></cleanup-datasets>
         <cleanup-control-blocks .doc=${this.doc}></cleanup-control-blocks>
+        <cleanup-datatypes .doc=${this.doc}></cleanup-datatypes>
```
danyill, line 22: Is passing `.doc=${this.doc}` sufficient to get an update after the refactoring?

danyill (commented):
A few suggested changes for me.

@JakobVogelsang suggested he had found some bugs, if he has some example scd files for this I will be happy to look into them prior to a review.

I think there is a need for some convenience methods -- today probably just "remove an IED and any templates only used by this IED" (or perhaps remove a DO/DA/LN and any templates only used by this element?). Some of this code doesn't conveniently meet this use case at the moment and I might need some suggestions on how best to refactor with this goal in mind (and @JakobVogelsang has been clever in the past in showing me how my initial approach often goes about this the hard way).

(Incidentally I really feel that Github should allow me to Request changes on my own PR! One can certainly be in conversation with one's self... :rofl: )

src/editors/Cleanup.ts
JakobVogelsang: Yes. As we exchange the doc the reference changes and the `hasChanged` method of `LitElement` triggers an update.

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+      'DA[bType="Struct"], BDA[bType="Struct"]',
+      'type',
+      'DAType'
+    );
+    const unreferencedEnumTypes = this.getUnusedType(
+      'DA[bType="Enum"], BDA[bType="Enum"]',
+      'type',
+      'EnumType'
+    );
+    this.unreferencedDataTypes = unreferencedLNTypes.concat(
```
JakobVogelsang, line 375: Why not directly have a function `unreferencedType()` that return a concatenated `Element[]`. It seems you are using it only here. That would save some not very readable lines :)

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+            </mwc-icon-button>
+          </abbr>
+        </h1>
+        ${this.renderFilterIconButton('LNodeType')}
+        ${this.renderFilterIconButton('DOType')}
+        ${this.renderFilterIconButton('DAType')}
+        ${this.renderFilterIconButton('EnumType')}
+        <filtered-list multi class="cleanupList"
+          >${Array.from(
+            this.unreferencedDataTypes.map(uType => this.renderListItem(uType))
```
JakobVogelsang, line 400: `uType` reads a bit bumpy. If you think the array upfront does give enough information on the element within map, use element or type. I would suggest to use descriptive names were ever possible. `unreferencedDataType` for example here.

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+            'cleanup.unreferencedDataTypes.alsoRemoveSubTypes'
+          )}"
+        >
+          <mwc-checkbox
+            checked
+            class="cleanSubTypesCheckbox"
+            ?disabled=${(<Set<number>>this.selectedDataTypeItems).size === 0 ||
+            !(
+              (<Set<number>>this.selectedDataTypeItems).size !== 0 &&
+              this.selectionContainsDOOrDAType()
```
JakobVogelsang, line 418: Why do you want to remove sub types only if `DOType` element and `DAType` element is selected.
This button can also be tricked by selecting `DOType` do whatever to the checkbox and deselecting it again.

I would always leave it enabled and checked by default.

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+      label="${translate('cleanup.unreferencedDataTypes.deleteButton')} (${(<
+        Set<number>
+      >this.selectedDataTypeItems).size || '0'})"
+      ?disabled=${(<Set<number>>this.selectedDataTypeItems).size === 0 ||
+      (Array.isArray(this.selectedDataTypeItems) &&
+        !this.selectedDataTypeItems.length)}
+      @click=${(e: MouseEvent) => {
+        const dataTypeItemsDeleteActions = cleanSCLItems(this.getCleanItems());
+        dataTypeItemsDeleteActions.forEach(deleteAction =>
+          e.target?.dispatchEvent(newActionEvent(deleteAction))
```
JakobVogelsang, line 307: Also, a minor thing: you can also dispatch on the host using `this.dispatchEvent(...)`. This pattern we use in wizards sometimes as it is a function and there is nothing else than to dispatch in the wizards items itsself.

Then you can also get rid of (e: MouseEvent)

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+      ><span class="unreferencedControl">${dType.getAttribute('id')!} </span>
+      <span>
+        <mwc-icon-button
+          label="Edit"
+          icon="edit"
+          class="editItem"
+          @click="${(e: MouseEvent) => {
+            e.stopPropagation();
+            if (dType.tagName === 'LNodeType') {
+              e.target?.dispatchEvent(
```
JakobVogelsang, line 159: Same here you can dispatch on host

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+                newSubWizardEvent(
+                  lNodeTypeWizard(<string>identity(dType), this.doc)
+                )
+              );
+            } else if (dType.tagName === 'DAType') {
+              e.target?.dispatchEvent(
```
JakobVogelsang, line 165: Same here you can dispatch on host

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+                newSubWizardEvent(
+                  editDaTypeWizard(<string>identity(dType), this.doc)
+                )
+              );
+            } else if (dType.tagName === 'DOType') {
+              e.target?.dispatchEvent(
```
JakobVogelsang, line 171: Same here you can dispatch on host

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+                newSubWizardEvent(
+                  dOTypeWizard(<string>identity(dType), this.doc)
+                )
+              );
+            } else if (dType.tagName === 'EnumType') {
+              e.target?.dispatchEvent(
```
JakobVogelsang, line 177: Same here you can dispatch on host

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+                newSubWizardEvent(
+                  eNumTypeEditWizard(<string>identity(dType), this.doc)
+                )
+              );
+            }
+          }}}"
+        ></mwc-icon-button>
+      </span>
+      <span slot="secondary"
+        >${dType.getAttribute('lnClass') ?? 'Unknown'}
```
JakobVogelsang, line 187: That I think is a bug. If the type is anything other than `LNodeType` you see an unknown. I would suggest to show `cdc` for `DOType`. For `DAType` there is nothing to show other than the `desc`

src/editors/cleanup/datatypes-container.ts
```diff
@@ -0,0 +1,529 @@
...
+    initialState = true
+  ): TemplateResult {
+    return html`<mwc-icon-button-toggle
+      slot="graphic"
+      label="filter"
+      ?on=${initialState}
+      class="t${dataType}Filter filter"
+      @click="${(e: MouseEvent) => {
+        e.stopPropagation();
+        this.toggleHiddenClass(`t${dataType}`);
```
JakobVogelsang, line 132: You do not need a toggle method here, although it is perfectly fine. The `mwc-toggle-icon-button` exposes is state as an attribute, and it therefore can be used for a CSS only solution. Like so:
```css
.list-items {
    display: none;
}

mwc-icon-button-toggle[on]~filtered-list>.whatever-list-item {
     display: inherit (don't know from scratch)
}

**Review by JakobVogelsang — CHANGES_REQUESTED** [review]

File: src/editors/cleanup/datatypes-container.ts (hunk 17 above)
Comment (danyill) [reply src/editors/cleanup/datatypes-container.ts:187]: You are quite correct.

File: src/editors/cleanup/datatypes-container.ts (hunk 18 above)
Comment (danyill) [reply src/editors/cleanup/datatypes-container.ts:132]: I think this was unnecessary cruft. I think the CSS-only solution was also implemented. I have removed it.
Good catch. Thanks! :+1:

Comment (danyill) [reply src/editors/cleanup/datatypes-container.ts:132]: Wait up, no, that's not correct. This code is adding a class to the children who don't appear as part of the list so they are not included in the filter selection (e.g. when I click on Select All in the filter).

**Comment by danyill** [comment]
Thank you for this review, I appreciate it. Hopefully, the :bug: have been reduced and I have made some improvements to the readability and fixed the style suggestions. 

Eventually, I hope to make these improvements to the other cleanup containers.

I am mainly bothered by not providing a convenience method for IED or Logical Node removal but perhaps that can be a separate PR for when the need is great.

**Comment by danyill** [comment]
It's unclear to me why CI failed so severely. Locally seems to work fine on both Node 14 and Node 16.

**Comment by danyill** [comment]
This PR has the same issue as the Control Blocks container in #873 where the filtering of the list doesn't work because the filter is changing the `display` property and so is the `filtered-list`.

**Comment by JakobVogelsang** [comment]
I had time on my hand to have a closer look at the inner function. I think that what we have here will lead to bug reports. 
Let's assume you have a `DOType` selected that has a child `DA` this is the only one referencing a `DAType`. Then after deleting the `DOType` the `DAType` will pop up in the list. Even if we check back recursively down to the last element, this can go wrong. Let's assume two `DOType` element are selected  and the algorithm checks back for the first one, we would first need to remove those to check for the second element. 

It appears to me that this algorithm is not very easy to make it real clean in any cases. We could also think of just removing those and show if new unused will appear. 

Do I miss something?

**Comment by danyill** [comment]
> Do I miss something?

No, you are quite correct :smile_cat:.

> It appears to me that this algorithm is not very easy to make it real clean in any cases. We could also think of just removing those and show if new unused will appear.

I know... it is tempting, but I cannot give in so easily, as, from an end-user perspective, that's making the user do the work the machine should.
Additionally we will want to add functions to remove IEDs and be able to safely and cleanly remove associated `DataTypeTemplates` without user interaction.

I find this a little confusing :confused: and time has passed, so I thought I should refresh and write my understanding before I try to implement it.

My current understanding for which I would be grateful to receive comment/correction:

* A `DOType` can reference an `SDO`
* An `SDO` can then reference another `DOType`
* Recursive references are not allowed, the standard says "Warning: recursive references are not allowed, but cannot be checked at syntax level!" I think this means there should not be references to `DOType[id='x'] > SDO[type='x'] > DOType[id='x']` as this would never end
* There can be an unspecified depth on `DOType > SDO > DOType > SDO > DOType` as long as there are no references that are looped
* `DOType`s eventually _must_ terminate with potentially a large number of leaves which become `DA` or `DAType`s

* A `DOType` can reference a `DA`
* A `DA` can reference an enum if `bType` = `Enum`
* A `DA` can reference a `DAType` if `bType` = `Struct` 
* Otherwise a `DA` reference a basic type
* There can be an unspecified depth on `DA[bType = 'Struct', type='thing1'] > DAType[id='thing1'] > BDA[bType='Struct', type="thing3"]> DAType[id='thing3']> BDA...` between `DAType` and `BDA`s where `bType=Struct` on a `BDA`
* Eventually `DAType`s _must_ terminate in basic data type (e.g. `BOOLEAN` or `FLOAT32`) via a `BDA`
* DATypes can also potentially have many leaves of `BDA`s

The code as written doesn't support the recursion for an unspecified depth, so that is a bug when the 'Remove subtypes' button is checked.

My understanding comes in diagram form if the above made no sense:

datatypetemplates

What we need to do is the following:

* `LNodeType`s are easy. If no `LN` has a `LNType` for the `id` of an `LNodeType`, it doesn't belong.
* A `DOType` is not needed if there is no matching `DO` _and_ there is no matching `SDO`
* A `DAType` is not needed if there is no matching `DA` _and_ there is no matching `BDA`
* I ignore enums because I think they are done fine for now.

I think this is done correctly at the moment.

* To "remove subtypes" we must step through each `LNodeType` reference to each `DO` and trace all the `DOType`s and `DAType`s that are referenced. We must then do the same for each selected `DOType` and `DAType`

  * Let's imagine we do this with an array of "referenced subtypes".
We store the `id` and object tagname (e.g. `DOType` or `DAType`) for object which is either a `DOType` or `DAType` as we do a search through them.
  * We must spot recursion by tracking the `id` in the `DA[id='x' =>  DAType[type='x'] and BDA[id='x'] => DAType[type='x']`. If there is more than one mention of the same `id` we must terminate and not loop endlessly (and provide an error in the log)

  * At the end, we have a list for all the `DataTypeTemplate`s of:
    * The unused ones (which we know we can remove "safely") -- I think this is working OK
    * A list of dependent objects catalogued by tagname and id. These _may_ be able to be removed if they are not used by other objects outside the selection if "remove subtypes" is checked.
      * We then need to go through all objects which are _not_ part of the selection. The moment we hit an id which is in our catalogue we drop this from the list of objects which could be removed (or we could do a selector and compare the total number in the selection vs. the total number in the document as we do presently)
      * At the end of this, we should be able to remove just the types which are genuinely not used.

>Even if we check back recursively down to the last element, this can go wrong. Let's assume two DOType element are selected and the algorithm checks back for the first one, we would first need to remove those to check for the second element.

WDYT? 
Does this address your concern above?
If it seems mostly right, I'll try to write some code soon.

**Comment by JakobVogelsang** [comment]
I think your analysis is quite correct. I think we can break it down to an even simpler rule that also allow a simple implementation, that is:

We must delete each data type template `LNodeType`, `DOType`, `DAType` and `EnumType` always then there is no reference to it from `lnType`, `type`. 

Based on this, I would therefore propose something like this:
1. Create a map of all `id`s and the number of their usages/references 
```ts
const dataTypeUsageCounter = new Map<string, number>();
Array.from(dataTypeTemplates.querySelectorAll('DO, SDO, DA, BDA')).forEach(
    typeElement => {
      const type = typeElement.getAttribute('type');
      if (!type) return;

      if (dataTypeUsageCounter.has(type)) dataTypeUsageCounter.set(type, dataTypeUsageCounter.get(type)! + 1);
      else dataTypeUsageCounter.set(type, 1);
    }
  );
```
2. when removing a data type template, check if its children have a reference
3. update the dataTypeUsageCounter
4. check if new data type templates are now unused
5. repeat...

danyill:
I, like a :snail: contemplate this.

>    Create a map of all ids and the number of their usages/references

We examine the code sample. The idea is good, but I think this must also be done recursively to get the correct count :wink:

danyill:
Thank you :clap: for reviewing the approach and for your helpful comments above.

I have done a little more work on this:

* I've rebased this to the latest `main` to deal with the passing of time.
* I've done my best to implement a proper search function, but it is possible there is a better way to write it.
* We must avoid a looped DataType reference, which never occurs except in a malformed file. Nevertheless, we must not let an infinite loop occur, so I've simply chosen an arbitrary limit for stack depth. Otherwise, we must carefully track a depth-first search through the templates, which seemed more complex than worthwhile.
* I've applied the fix for filtering as per #910.
 
I shall be delighted to receive further feedback.

danyill:
That is true, it is awful with larger files. I get a "Page Unresponsive" with Chrome 3-4 times with your example file.

I think indexing all usages of all templates is quite costly because it requires the full data model to be traversed and on large files this is causing significant slowdown.

Perhaps a web worker is required? 

I think EnumTypes were not being indexed (was that your "left behind"?). I forgot because I was focussed on the DO/DA issues, sorry :facepalm:. 

I have pushed a commit for this (and can add a test later if you'd like).

JakobVogelsang (commented):
One missing test and a little CSS thing :)

danyill:
Thank you for your review, I am grateful, :clap: I have addressed the Enum issue through snapshots and other tests and I think this is adequate.
Having taken the longest PR review medal, :1st_place_medal: we must not prolong the pain :drop_of_blood: and the joy :grin: 
I guess we'll worry about performance issues at some future point as we have decided to for other plugins or if it makes people unhappy.

I have not been brave enough to try and rebase this. There have been force pushes on main since this started, so if possible I would like to leave this to Jakob :pray:

JakobVogelsang (approved):
LGTM

PR #703, opened 2022-04-25
fix(wizard-dialog): avoid header overlap with extra action buttons

JakobVogelsang:
Unfortunately we do not have direct access to the header of `mwc-dialog` nor there are special CSS properties defined for that. The only somewhat responsive solution I found is this. The solution is not perfect as I feel!!

Alternatively, we could just set a min width, which does not work optimally on all wizard-dialog. They seem to be stretched.


dlabordus (approved):

dlabordus:
It's not perfect indeed, but if it works for now let's use it and maybe we find a better solution in the future.

ca-d (approved):
LGTM

PR #706, opened 2022-04-27
feat(zeroline): add read-only subfuction-editor

JakobVogelsang:
The second of the two to render read-only `Function` and `SubFunction` editors within the Substation editor.


dlabordus (approved):

ca-d (approved):
The code LGTM, but I couldn't do local exploratory testing because I lack a file containing functional substation elements and this editor is readonly.

JakobVogelsang:
You can use the file provides in test/testifies/zeroline/functions

PR #712, opened 2022-04-28
bug(wizards): Update attributes of Terminal when updating Voltage Level/Bay name.

### LINKED ISSUE #674: Updating Terminal Attributes after updating name of VoltageLevel or Bay
**Describe the bug**
When updating the name of a Bay or VoltageLevel these names are also registered as BayName/VoltageLevelName by Terminals.
These attributes of the Terminal aren't updated.
In PR #671 this is solved for the name of the Substation, the strange thing is that this cause problems when validating the XML, for Bay and VoltageLevel this isn't cause any validation problems.

**To Reproduce**
1. Open a SCD with Substation, VoltageLevel, Bay and Terminal(s), where there is a reference in Terminal to the different levels.
2. Update the name of the VoltageLevel and Bay.
3. Save the file and check if the attributes in the terminal are updated.

**Expected behavior**
The attributes in the Terminal should be updated

**Additional context**
Example part where the attributes aren't updated: 
```
<Substation name="AA1-UPD" desc="Substation" sxy:x="1" sxy:y="3">
	<VoltageLevel sxy:x="1" sxy:y="3" name="J1-UPD" desc="Voltage Level">
		<Voltage unit="V" multiplier="k">20</Voltage>
		<Bay sxy:x="1" sxy:y="6" name="Bay A-UPD" desc="Feld A">
			<ConductingEquipment type="DIS" sxy:x="1" sxy:y="1" name="QB1">
				<Terminal name="T3" connectivityNode="AA1/J1/Bay A/L1" substationName="AA1" voltageLevelName="J1" bayName="Bay A" cNodeName="L1" txy:actTerm="2"/>
			</ConductingEquipment>
```

dlabordus:
closes #674 

- Updated generic reference util to make it possible to search in the document or only in the parent of the element.
For Voltage Level the name is unique inside the Substation and for Bay this is in the Voltage Level.
- Moved some test from zeroline to wizards, because these test are only testing these wizards.
- Renamed "foundation.ts" in test directory to "test-support.ts". There is a more clear differences now between foundation.ts and foundation.test.ts in the test directory.


JakobVogelsang:
@dlabordus I have to apologize. I have moved files that are related to this fix and not seeing you are working on the same files. I will fix the missing tests.

However, I did the first review and the implementation looks good to me. 
There is one thing that I have to double-check, though. You are assuming that the `Terminal` element is only within the parent element of the name element you want to change. I am not sure if this always true :)

dlabordus:
@JakobVogelsang, already fixed the merge. You just moved some files, but I also moved these files. 
The Test Files where not related to the Editor, but to the Wizards, so I moved these to the Wizards directory.
This way we don't create double test, like what happened with Substation Wizard.

dlabordus:
@JakobVogelsang, I updated the reference function to include parent names if needed. This should work for all situation we currently know (, I think :-) )

src/wizards/foundation/references.ts
```diff
@@ -1,54 +1,132 @@
...
+ * More complex function to search for elements for which the value of an attribute needs to be updated.
+ * To find the correct element the name of a parent element also needs to be included in the search.
+ *
+ * For instance when the name of a Bay is updated only the terminals need to be updated where of course
+ * the old name of the bay is the value of the attribute 'bayName', but also the voltage level and substation
+ * name need to be included, because the name of the bay is only unique within the voltage level.
+ * The query will then become
+ * `Terminal[substationName="<substationName>"][voltageLevelName="<voltageLevelName>"][bayName="<oldName>"]`
+ *
+ * @param elementName - The tagName of the elements to search for.
```
JakobVogelsang, line 102: The function is working perfectly :) Thanks for that. There is only one remark about naming
`elementName` looking at our code could imply the attribute name of the element

JakobVogelsang, line 102: The function is working perfectly :) Thanks for that. There is only one remark about naming
`elementName` looking at our code could imply the attribute name of the element

JakobVogelsang (approved):
LGTM aside of one naming issue as a suggestion

src/wizards/foundation/references.ts
dlabordus: Fixed.

dlabordus: Fixed.

PR #718, opened 2022-05-01
feat(Subscription): Add edit button or hyperlink to GSEControl dialog in the Subscription Editor

### LINKED ISSUE #577: Add edit button or hyperlink to GSEControl dialog in the Subscription Editor
In the upcoming subscription editor, it would help the user's context switching to be able to directly see and edit a GSEControl from the subscription window.

I suggest either adding an edit button or being able to hyperlink the dataset to open the editor.

Perhaps something like the following would be sufficient:

image

Flurb:
Closes #577


src/editors/subscription/subscriber-list.ts
```diff
@@ -279,6 +281,9 @@ export class SubscriberList extends LitElement {
   private async onViewChange(event: ViewEvent) {
     view = event.detail.view;
 
+    this.currentSelectedIed = undefined;
```
JakobVogelsang, line 284: Should this not be part of the upstream branch?

Flurb: It depends on the select-by-subscriber branch, but it's a feature on it's own.

Ideally, merge the select-by-subscriber first and this one after.

JakobVogelsang:
I have tested this PR with some files and have some very basic remarks:
1. the button is allocated to the subscriber list. I do not think this is very intuitive. I would have expected those in the list item on the left-hand side. 
2. I do not think we need to edit the IED here. I cannot think of any case where it makes sense to change the IED name halfway through the process of subscription.

Flurb:
> I have tested this PR with some files and have some very basic remarks:
> 
> 1. the button is allocated to the subscriber list. I do not think this is very intuitive. I would have expected those in the list item on the left-hand side.
> 2. I do not think we need to edit the IED here. I cannot think of any case where it makes sense to change the IED name halfway through the process of subscription.

Removing the IED edit here is fine for me, but the allocation of the edit button is as refined in the issue. Why is this not correct anymore?

JakobVogelsang:
@Flurb I is my fault, I am sorry. If you look into the issue @danyill 's proposal was putting it next to the GOOSE element. That would be my preferred solution as well. I however, missed the comment of @Sander3003 putting it in the place you have putted it in. During the refinement we were talking about different thing therefore. Sorry for that!

How about making it a topic for next Monday. The implementation is good so far. It is just about the location.

Flurb:
> @Flurb I is my fault, I am sorry. If you look into the issue @danyill 's proposal was putting it next to the GOOSE element. That would be my preferred solution as well. I however, missed the comment of @Sander3003 putting it in the place you have putted it in. During the refinement we were talking about different thing therefore. Sorry for that!
> 
> How about making it a topic for next Monday. The implementation is good so far. It is just about the location.

Good idea!

danyill:
Thanks for this feature, I am very grateful. Some brief feedback:

1. I agree that changing the IED name within a GOOSE subscription is confusing. I think we should just remove this. I'm sorry I also didn't comment on this earlier (this occurs when I have the radio button on "Show subscribed GOOSE publisher for selected IED").
2. When I have "Show subscriber IED(s) per selected GOOSE" the edit button allows the GSEControl to be edited. The aspect that I find a little confusing about this is that I would expect editing to be on the Publisher side.

JakobVogelsang (changes requested):
Thanks for moving it. I really do think it is more intuitive like that. 
I still would request one change 
1. Use `mwc-icon-button` instead of `mwc-icon`. To make sure it is still centered, there is a nice CSS rule: `--mdc-list-item-meta-size` that must be set to the same size as the mwc-icon-button default 48px. The reason here is we have something indicating around the icon that it is clickable.

...and suggest one other:
2. Use CSS to hide the non-selected meta. 
```ts
<mwc-icon-button
     class="${classMap({
          hidden: gseControl !== selectedGseControl,
        })}"
        slot="meta"
        icon="edit"
        @click=${() => this.openEditWizard(gseControl)}
     ></mwc-icon-button>

...

static styles = css`
    ${styles}
   ...
    mwc-icon-button.hidden {
      display: none;
    }
  `;
```

Flurb:
> Thanks for moving it. I really do think it is more intuitive like that. I still would request one change
> 
> 1. Use `mwc-icon-button` instead of `mwc-icon`. To make sure it is still centered, there is a nice CSS rule: `--mdc-list-item-meta-size` that must be set to the same size as the mwc-icon-button default 48px. The reason here is we have something indicating around the icon that it is clickable.
> 
> ...and suggest one other: 2. Use CSS to hide the non-selected meta.
> 
> ```ts
> <mwc-icon-button
>      class="${classMap({
>           hidden: gseControl !== selectedGseControl,
>         })}"
>         slot="meta"
>         icon="edit"
>         @click=${() => this.openEditWizard(gseControl)}
>      ></mwc-icon-button>
> 
> ...
> 
> static styles = css`
>     ${styles}
>    ...
>     mwc-icon-button.hidden {
>       display: none;
>     }
>   `;
> ```

What is the advantage of using CSS here instead of the proposed solution?

dlabordus (approved):

dlabordus (approved):

JakobVogelsang (approved):

Issue #707, opened 2022-04-27
Show `EqFunction` and `SubEqFunction` in Substation editor


JakobVogelsang:
As a user of OpenSCD I want to see defined `EqFunction` and `SubEqFunction` element in the Substation editor.

** Requirements **
- on toggle `showfunction` in the Substation editor convert action-icon type editors to action-pane type ones
- render `Eqfunction` and `SubEqFunction` as child elements within `ConductingEquipment`, `PowerTransformer`

PR #720, opened 2022-05-02
feat(editors/substation): show read-only `EqFunction` and `SubEqFunction`

JakobVogelsang:
The first of three requests that cover the issue #707 
1. ~~transform CondunctingEquipment and PowerTransformer to action-pane on function filter disabled~~
2. ~~show `EqFunction` element within ConductingEquipment and PowerTransformer~~
3. ~~show `EqSubFunction` within `EqFunction`~~

### DIFF
--- a/src/editors/substation/bay-editor.ts
+++ b/src/editors/substation/bay-editor.ts
@@ -45,7 +45,7 @@ export class BayEditor extends LitElement {
   element!: Element;
   @property({ type: Boolean })
   readonly = false;
-  /** Wheter `Function` and `SubFunction` are rendered */
+  /** Whether `Function` and `SubFunction` are rendered */
   @property({ type: Boolean })
   showfunctions = false;
 
--- a/src/editors/substation/bay-editor.ts
+++ b/src/editors/substation/bay-editor.ts
@@ -7,6 +7,7 @@ import {
   query,
   TemplateResult,
 } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map';
 import { translate } from 'lit-translate';
 
 import '@material/mwc-icon-button';
@@ -177,13 +178,19 @@ export class BayEditor extends LitElement {
         >
       </abbr>
       ${this.renderIedContainer()} ${this.renderFunctions()}
-      <div id="ceContainer">
+      <div
+        class="${classMap({
+          content: true,
+          actionicon: !this.showfunctions,
+        })}"
+      >
         ${Array.from(
           getChildElementsByTagName(this.element, 'PowerTransformer')
         ).map(
           pwt =>
             html`<powertransformer-editor
               .element=${pwt}
+              ?showfunctions=${this.showfunctions}
             ></powertransformer-editor>`
         )}
         ${Array.from(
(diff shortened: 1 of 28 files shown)


src/editors/substation/conducting-equipment-editor.ts
```diff
@@ -6,10 +6,12 @@ import {
   property,
   TemplateResult,
 } from 'lit-element';
+import { translate } from 'lit-translate';
 
 import '@material/mwc-fab';
```
dlabordus, line 12: Shouldn't there be some extra imports for MWC Components?

This also applies to others editors where mwc-icon and mwc-icon-button are added.

src/editors/substation/bay-editor.ts
```diff
@@ -202,12 +210,16 @@ export class BayEditor extends LitElement {
   static styles = css`
     ${styles}
 
-    #ceContainer {
+    .content.actionicon {
```
dlabordus, line 213: Is it not possible to use the prtContent way, like doen in Substation and VoltageLevel?
Bay is the only one handling powertransformer-editor differently.

src/editors/substation/powertransformer-editor.ts
```diff
@@ -31,6 +33,9 @@ export class PowerTransformerEditor extends LitElement {
   get name(): string {
     return this.element.getAttribute('name') ?? 'UNDEFINED';
   }
+  /** Wheter `EqFunction` and `SubEqFunction` are rendered */
```
dlabordus, line 36: Small typo, wheter, should be Whether.

dlabordus (changes requested):

dlabordus:
More a general question,
How does this PR relates to EqFunction and SubEqfunction? 
It looks more like there was a PowerTransformer problem when showing functions....

ca-d:
I really like the user interface overall. Just a few comments:

- `h6` looks too small on my screen to be readable. I'd consider stopping the size reduction at `h5` or even `h4` for deeply nested editors.
- Now that the action icons transform into action panes, I distinctly notice that Equipment doesn't have the "Duplicate" and "Add children" buttons yet. Which of these might make sense to add?
- If we find the dotted outline without focus to be a disturbance, we could think about making the unfocused heading and/or the secondary colour.

Otherwise, LGTM!

src/editors/substation/conducting-equipment-editor.ts
JakobVogelsang: True!! Thanks I missed it!

src/editors/substation/powertransformer-editor.ts
JakobVogelsang: Thanks :)

src/editors/substation/bay-editor.ts
JakobVogelsang: I was wondering a bit too, but at the end did like it. The `PowerTransformer` otherwise does take up too much space. Meaning, I kind of like that `PowerTransformer` within a Bay is handled like a `...Equipment`. Is this ok with you or does it bother you this way?

dlabordus: Fine for me.

JakobVogelsang:
> More a general question, How does this PR relates to EqFunction and SubEqfunction? It looks more like there was a PowerTransformer problem when showing functions....

Within the `Substation` editor we did on purose hide the children of `ConductingEquipment`, `PowerTransformer` and in the future `GeneralEquipment` to reduce the complexity a user sees. For function naming not usable as the elements `EqFunction` is a direct child of `ConductingEquipment`, `PowerTransformer` and others ... thus we need to be able to show those children and that is why I started to do this :)

PR #722, opened 2022-05-04
feat(editors/substation): add eq-function-editor

JakobVogelsang:
The second of the three


src/editors/substation/eqfunction-editor.ts
```diff
@@ -0,0 +1,35 @@
...
+  property,
+  customElement,
+  state,
+} from 'lit-element';
+
+import '../../action-pane.js';
+
+/** Pane rendering `Function` element with its children */
+@customElement('eqfunction-editor')
+export class EqFunctionEditor extends LitElement {
```
ca-d, line 14: Very minor naming question: I think so far we mostly transcribed camel case class names like `MyNewExampleEditor` into kebab case tag names like `my-new-example-editor`. Am I right about this? If so, do we want to continue that trend or is there a good reason to leave out the first dash here?

src/editors/substation/eqfunction-editor.ts
```diff
@@ -0,0 +1,35 @@
...
+  /** The edited `EqFunction` element */
+  @property({ attribute: false })
```
ca-d, line 16: Is this the same as `@state` or the deprecated `@internalProperty`?

src/editors/substation/eqfunction-editor.ts
```diff
@@ -0,0 +1,35 @@
...
+  get header(): string {
+    const name = this.element.getAttribute('name');
+    const desc = this.element.getAttribute('desc');
+    const type = this.element.getAttribute('type');
+
+    return `${name}${desc ? ` - ${desc}` : ''}${type ? ` (${type})` : ''}`;
+  }
+
+  render(): TemplateResult {
+    return html`<action-pane
```
ca-d, line 28: Have you tried replacing this placeholder pane with an `action-icon`?

ca-d (approved):
LGTM, all questions are suggestions only. Merge at your pleasure.

src/editors/substation/eqfunction-editor.ts
JakobVogelsang: `@state()` is a private/protected property, whereas `@property({attribute: false})` is a public one that is not reflected as attribute.

src/editors/substation/eqfunction-editor.ts
JakobVogelsang: This is empty just because the `EqSubFunction` is not implemented yet :). Next PR

src/editors/substation/eqfunction-editor.ts
JakobVogelsang: It reads a bit strange with short word e.g. `eq-sub-function-editor` :). Still I would stick to the clear rule we used in most parts of the code :)

PR #726, opened 2022-05-05
feat(editors/substation): add read-only eq-sub-function-editor

JakobVogelsang:
The last to have all the function types visible as read-only panes

### DIFF
--- a/src/editors/substation/eq-sub-function-editor.ts
+++ b/src/editors/substation/eq-sub-function-editor.ts
@@ -8,13 +8,12 @@ import {
 } from 'lit-element';
 
 import '../../action-pane.js';
-import './eq-sub-function-editor.js';
 import { getChildElementsByTagName } from '../../foundation.js';
 
-/** Pane rendering `EqFunction` element with its children */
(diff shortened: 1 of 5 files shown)


src/editors/substation/eq-sub-function-editor.ts
```diff
@@ -0,0 +1,46 @@
...
+import '../../action-pane.js';
+import { getChildElementsByTagName } from '../../foundation.js';
+
+/** Pane rendering `EqSubFunction` element with its children */
+@customElement('eq-sub-function-editor')
+export class EqSubFunctionEditor extends LitElement {
+  /** The edited `EqSubFunction` element */
+  @property({ attribute: false })
+  element!: Element;
+  @state()
```
ca-d, line 19: Sorry if it's a stupid question: what's the difference between this and the `attribute: false` above?

JakobVogelsang: It is not. I was thinking the `@state()` decorator will set the property to private. My bet!

PR #730, opened 2022-05-09
feat(editors/substation): add read-only l-node-editor

JakobVogelsang:


ca-d (changes requested):
I feel confused regarding two aspects of styling:

- The `l-node-editor` element, when clicked, becomes small and a label hovers far away. (screenshot 1)
  > Do you think there's a way we could adapt the `action-icon` to only do these things when there are actual actions slotted into it? I'd love to have that feature in general but am not sure where in the code or when in the lifecycle to detect slotted elements. If not, maybe you could add a `hideActions` boolean property analogous to `mwc-dialog`'s?
- I'm confused by the difference between different types of functions (EqSubFunc and EqFunc in the example in screenshot 2) in terms of how they graphically distribute their children. The lower option (stacked and touching) just looks off without any distance between the elements, as a user it makes me think there must be a bug.

## Screenshots

### Focus issue
l-node-editor-styling

### Grid issue
l-node-editor-eqfunc-eqsubfunc-difference

JakobVogelsang:
The second only was a bug. Sorry for that very obvious issue. The first one is resolved in an extra PR #752 adding an extra property to `action-icon`.

ca-d (approved):
LGTM

PR #732, opened 2022-05-10
feat(editor/ied): Add wizard/action to remove IED including references

### LINKED ISSUE #616: Remove IED button
As an OpenSCD engineer user
I want to a button to remove IED's from a SCL file
So I can remove IED's that are not needed any longer

Accpetance criteria:
The IED and related information should be removed
Including the related reporting blocks 

Background:
When editing a substation, it could be usefull to remove a IED of an exsiting configuration.

dlabordus:
closes #616


src/translations/de.ts
```diff
@@ -228,10 +228,13 @@ export const de: Translations = {
       descHelper: 'Beschreibung des IED',
       title: {
         edit: 'IED bearbeiten',
+        delete: '???',
```
dlabordus, line 231: @JakobVogelsang, do you have the German translations for these two.

English versions:
- delete: 'Remove IED with references'
- references: 'References to be removed'

JakobVogelsang, line 231:
```suggestion
        delete: 'Gelöschte Abhängigkeiten',
```

JakobVogelsang, line 231:
```suggestion
        delete: 'IED mit Abhängigkeiten entfernen',
```

src/translations/de.ts
```diff
@@ -228,10 +228,13 @@ export const de: Translations = {
...
+        references: '???',
```
JakobVogelsang, line 232:
```suggestion
        references: 'Gelöschte Abhängikeiten',
```

JakobVogelsang:
Before doing the code review, I tested the functionality and have some comments:
1. I think that having a MenuAction in the edit wizard for the IED is a nice to have feature. However, as we have an extra action-pane for each IED there should be a delete/remove icon button next to the edit icon button. To prevent unintentional removal, I would love to have a confirmation dialog: `Are you really sure!!`
2. The `Remove IED with references` wizard does show the code toggle button in pro mode. I think this is misleading and should be removed here.
3. When removing `ExtRef` elements, we need to check that the parent `Inputs` should not be empty after removal. We had the same issue in the subscription. Otherwise, removing IED potentially leads to schema invalid files.
4. When removing, the `Select IED` is not updated properly.

JakobVogelsang:
Missing also on the first glimpse is the cleaning up of the `DataTypeTemplates` section is missing. My test: I opened an SCD with some IEDs. Removed all of them and there were still entries in the `DataTypeTemplates` section. When removing, we should make sure that:
1. A `LNodeType` is removed if it is not referenced by any `LN0`/`LN` lnType
2. A `DOType` is removed if it is not referenced by `DO`/`SDO` type
3. A `DAType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Struct"`
4. A `EnumType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Enum"`

dlabordus:
> Missing also on the first glimpse is the cleaning up of the `DataTypeTemplates` section is missing. My test: I opened an SCD with some IEDs. Removed all of them and there were still entries in the `DataTypeTemplates` section. When removing, we should make sure that:
> 
> 1. A `LNodeType` is removed if it is not referenced by any `LN0`/`LN` lnType
> 2. A `DOType` is removed if it is not referenced by `DO`/`SDO` type
> 3. A `DAType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Struct"`
> 4. A `EnumType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Enum"`

Is this maybe not something that should be done by the Cleanup Plugin? Is this not a bit to much cleaning up?

dlabordus:
> Before doing the code review, I tested the functionality and have some comments:
> 
> 1. I think that having a MenuAction in the edit wizard for the IED is a nice to have feature. However, as we have an extra action-pane for each IED there should be a delete/remove icon button next to the edit icon button. To prevent unintentional removal, I would love to have a confirmation dialog: `Are you really sure!!`

Don't we have the Redo option for this? 
You can now also remove a Substation without this question....

> 2. The `Remove IED with references` wizard does show the code toggle button in pro mode. I think this is misleading and should be removed here.

Is this easy to do? Or do we need to draw our own MWC Dialog for this?

> 3. When removing `ExtRef` elements, we need to check that the parent `Inputs` should not be empty after removal. We had the same issue in the subscription. Otherwise, removing IED potentially leads to schema invalid files.

Need to look at this.

> 4. When removing, the `Select IED` is not updated properly.

What do you exactly mean by this? On the IED Editor screen...

JakobVogelsang:
Thanks for the quick reply @dlabordus. 
1.

JakobVogelsang:
> > Missing also on the first glimpse is the cleaning up of the `DataTypeTemplates` section is missing. My test: I opened an SCD with some IEDs. Removed all of them and there were still entries in the `DataTypeTemplates` section. When removing, we should make sure that:
> > 
> > 1. A `LNodeType` is removed if it is not referenced by any `LN0`/`LN` lnType
> > 2. A `DOType` is removed if it is not referenced by `DO`/`SDO` type
> > 3. A `DAType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Struct"`
> > 4. A `EnumType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Enum"`
> 
> Is this maybe not something that should be done by the Cleanup Plugin? Is this not a bit to much cleaning up?

I would argue now this cannot be done in the CleanUp editor. Although, they can share the same logic, but whenever we are firing a complex action I would argue that the result should be at least schema valid. Or in other words, how does a user know he/she should after removing an IED go to a non-default editor and do something here that for most is probably magic.

JakobVogelsang:
> > Before doing the code review, I tested the functionality and have some comments:
> > 
> > 1. I think that having a MenuAction in the edit wizard for the IED is a nice to have feature. However, as we have an extra action-pane for each IED there should be a delete/remove icon button next to the edit icon button. To prevent unintentional removal, I would love to have a confirmation dialog: `Are you really sure!!`
> 
> Don't we have the Redo option for this? You can now also remove a Substation without this question....
I would agree. I just suggested this because I was thinking you did not want to have the button due to easy accessibility :). Sorry for the misunderstanding. This would be a nice to have, not a must-have as we can always redo :)
> 
> > 2. The `Remove IED with references` wizard does show the code toggle button in pro mode. I think this is misleading and should be removed here.
> 
> Is this easy to do? Or do we need to draw our own MWC Dialog for this?
No the Wizard has an optional `element?: Element`. Just remove this :)
> 
> > 3. When removing `ExtRef` elements, we need to check that the parent `Inputs` should not be empty after removal. We had the same issue in the subscription. Otherwise, removing IED potentially leads to schema invalid files.
> 
> Need to look at this.
> 
> > 4. When removing, the `Select IED` is not updated properly.
> 
> What do you exactly mean by this? On the IED Editor screen...

Yes, there is a select field within the IED editor. When I remove an IED this field is not being updated nor does the editor itself. It is still showing the removed IED. At least this is happening on my mac with firefox.

dlabordus:
> > > Missing also on the first glimpse is the cleaning up of the `DataTypeTemplates` section is missing. My test: I opened an SCD with some IEDs. Removed all of them and there were still entries in the `DataTypeTemplates` section. When removing, we should make sure that:
> > > 
> > > 1. A `LNodeType` is removed if it is not referenced by any `LN0`/`LN` lnType
> > > 2. A `DOType` is removed if it is not referenced by `DO`/`SDO` type
> > > 3. A `DAType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Struct"`
> > > 4. A `EnumType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Enum"`
> > 
> > 
> > Is this maybe not something that should be done by the Cleanup Plugin? Is this not a bit to much cleaning up?
> 
> I would argue now this cannot be done in the CleanUp editor. Although, they can share the same logic, but whenever we are firing a complex action I would argue that the result should be at least schema valid. Or in other words, how does a user know he/she should after removing an IED go to a non-default editor and do something here that for most is probably magic.

But not removing these items, doesn't make the schema invalid, correct?
After a IED is removed these are unused, like that is also the case when you first add these as templates.

JakobVogelsang:
> > > > Missing also on the first glimpse is the cleaning up of the `DataTypeTemplates` section is missing. My test: I opened an SCD with some IEDs. Removed all of them and there were still entries in the `DataTypeTemplates` section. When removing, we should make sure that:
> > > > 
> > > > 1. A `LNodeType` is removed if it is not referenced by any `LN0`/`LN` lnType
> > > > 2. A `DOType` is removed if it is not referenced by `DO`/`SDO` type
> > > > 3. A `DAType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Struct"`
> > > > 4. A `EnumType` is removed if it is not referenced by `DA`/`BDA` type with `bType="Enum"`
> > > 
> > > 
> > > Is this maybe not something that should be done by the Cleanup Plugin? Is this not a bit to much cleaning up?
> > 
> > 
> > I would argue now this cannot be done in the CleanUp editor. Although, they can share the same logic, but whenever we are firing a complex action I would argue that the result should be at least schema valid. Or in other words, how does a user know he/she should after removing an IED go to a non-default editor and do something here that for most is probably magic.
> 
> But not removing these items, doesn't make the schema invalid, correct? After a IED is removed these are unused, like that is also the case when you first add these as templates.

That is true, you are right. During specification phase, this is normal and even necessary, as you design the Templates before having any IED. 

I still would say we have to do it - potentially in two steps. I cannot think of any use case where removing this element is wrong. On the downside, I saw a lot of SCD files that were wasted with unused `...Types`. 

Is there a particular reason not to do it that I do not see?

src/translations/de.ts
dlabordus: Thanks.

src/translations/de.ts
dlabordus: Thanks.

src/editors/IED.ts
```diff
@@ -49,7 +49,9 @@ export default class IedPlugin extends LitElement {
   }
 
   private get selectedIed(): Element | undefined {
-    if (iedEditorSelectedIed === undefined) {
+    // When there is IED selected, or the selected IED has no parent (IED has been removed)
```
JakobVogelsang, line 52:
```suggestion
    // When there is no IED selected, or the selected IED has no parent (IED has been removed)
```

JakobVogelsang (approved):
LGTM

PR #733, opened 2022-05-11
feat(wizards/subfunction): add create wizard

JakobVogelsang:
The second of the three, allowing to add `SubFunction` elements within `Function` elements.


src/editors/substation/function-editor.ts
```diff
@@ -26,6 +45,19 @@ export class FunctionEditor extends LitElement {
...
+  @query('mwc-icon-button[icon="playlist_add"]') addButton!: IconButton;
+
+  private openCreateWizard(tagName: string): void {
+    const wizard = wizards[<SCLTag>tagName].create(this.element!);
+
+    if (wizard) this.dispatchEvent(newWizardEvent(wizard));
+  }
+
+  firstUpdated(): void {
+    this.addMenu.anchor = <HTMLElement>this.addButton;
```
dlabordus, line 58: @JakobVogelsang, why is this needed? This is new for me.

JakobVogelsang: The web-component `mwc-menu` has a property anchor that decides where the menu is rendered. As an example the attributed `corner` and `menuCorner` are relative to this property. If not set than the component just takes the next logical and in this specific case it does not suite us. You can test by removing this line and then testing with the testfile `.../zeroline/functions`. 

The reason why you do not see it on every menu is that the anchor can be set indirect by wrapping it into a `div` with a very specific styling: 
```ts
<div style="position: relative;">
  <mwc-button id="button" raised label="Open Menu"></mwc-button>
  <mwc-menu id="menu">
    <mwc-list-item>Item 0</mwc-list-item>
    <mwc-list-item>Item 1</mwc-list-item>
    <mwc-list-item>Item 2</mwc-list-item>
    <mwc-list-item>Item 3</mwc-list-item>
  </mwc-menu>
</div>
```
 This does not fit however to what we are doing here and that is why we need to have this line :)

ca-d (approved):
LGTM :+1:

PR #744, opened 2022-05-13
feat(wizards/sampledvaluecontrol): add create wizard

JakobVogelsang:
This is finally the PR allowing to create `SampledValueControl` element using wizards. All the functionality is implemented, but tests are missing. @danyill you can try it out, but the code review is not forth doing yet


danyill:
Thank you for this opportunity :-)

It looks really good although I have only played a little so far and my SMV experience will grow soon but at the moment it is a bit rudimentary. I will plan to take another look at this in a day or two.

I am not sure why this is not accepted but I am fairly sure 22 is a valid VLAN ID:

image

However, 222 is accepted.

I do think some of the field limits (e.g. APPID) at some point need to have a human-readable tooltip (otherwise the casual user will never guess!). Particularly because it has the min/max length of 4 and no default. I'd almost say it should be input like a pin code with a `_ _ _ _` or similar (same applies to GOOSE creation wizard).

Surely when I select data for an SMV message I should be constrained in the finder screen to a particular class?

I found this in IEC 61850-7-2:

> The model applies to the exchange of values of a DataSet. The DataSetMembers typically belong to the Common Data Class SAV (sampled analogue value), but can belong to any other Common Data Classes as defined in IEC 61850-7-3, if sampled with the specific rate. A buffer structure shall be defined for the transmission of the sampled values.

Similar statements in 9-2.

I think for starters we should restrict the CDC to `SAV` as I think most (all?) tools do this at present. Otherwise, it's a little too confusing for the user. Perhaps a toggle button or a setting would allow overriding this. WDYT? Is there a current use case for having this completely open? If there is then I think we may need some CDC toggle buttons... or something to help the user out.

I have no particular idea what this tickbox does:

image

I thought all SMV were IEC 61850-9-2 by definition? :wink:

danyill:
I note in passing that I seem to be able to duplicate DA values in a dataset by taking an existing SV dataset but then I can't save them (however this is checked the algorithm probably discards duplicates). Either users shouldn't be able to add duplicates or having been able to successfully add them they should be able to save. This may be a bit tangential to this create wizard and need a separate issue?

JakobVogelsang:
Regarding the first issue:
IEC 61850 has both the multicast version and the well other that is defined in the 9-1. Those have never been used, but according to the schema you can choose. We had the same discussion with the `GSEControl` `type` attribute which can be selected to `GSSE` although it is deprecated. Do you think this is so bad? So long we are working to become a SCL editor, I would like to not be too restrictive and rather work with good default values :)

Regarding the second one:
Indeed, misleading but as is dealing more with editing data set should be an extra issue :) Would you open one?

JakobVogelsang:
**Allow only specific data type in DataSets used within SampledValue streams.** 
I first was struggling with it. As the -6 does not restrict anything here. But I think in this case it really will do more harm, not restricting it. I even more so do think we should not use a finder here but on a list with those 8 filterers
 1. `TCTR.AmpSv.instMag.i`
 2. `TCTR.Amp.instMag.i`
 3. `TCTR.AmpSv.instMag.f`
 4. `TCTR.Amp.instMag.f`
 5. `TCTR.VolSv.instMag.i`
 6. `TCTR.Vol.instMag.i`
 7. `TCTR.VolSv.instMag.f`
 8. `TCTR.Vol.instMag.i`

Or should we even be more restictive here. Acc to 9-2LE only the int32 type for instMag is allowed?

JakobVogelsang:
**Regarding the pattern definition in the input fields:**
Here I clearly would like to work with the schema for various reasons. The most important is that it will allow us to use automated methods to create standard wizard in the future. Your comment is very valid, though. What would you think about having good validation messages?

ca-d:
> Or should we even be more restictive here. Acc to 9-2LE only the int32 type for instMag is allowed?

For what it's worth: I personally think we should consider compatibility with all parts of `IEC 61850` , not just `61850-6`, so be as restrictive as possible in general. In the case of `9-2LE` I think currently a warning would be appropriate, but we should allow for values that would be valid in `9-2` "fat edition" as well, just in case anyone ever implements that.

Just my two cents as a software developer who has never been and likely will never be in a position to actually use any of these standards.

danyill:
> Regarding the first issue: IEC 61850 has both the multicast version and the well other that is defined in the 9-1. Those have never been used, but according to the schema you can choose. We had the same discussion with the `GSEControl` `type` attribute which can be selected to `GSSE` although it is deprecated. Do you think this is so bad? So long we are working to become a SCL editor, I would like to not be too restrictive and rather work with good default values :)
> 

I guess we're referring to "IEC 61850-9-1 Sampled values over serial unidirectional multidrop point to point link" which is a part of the standard which is withdrawn and replaced with 9-2 a long time ago... I think we should only make reference and support that if we think this is a "real thing" or that there is a use case for it. But it does go to the heart of whether OpenSCD is trying to support all possible variants of the standard over time or a reasonable set of use cases. I think taking a narrower focus is likely to lead to a better outcome.

If OpenSCD is too generic an XML editor for SCL files without taking into account the historical development of the standards and what is in the marketplace then "accumulated cruft" of the standard will detract from usability IMHO.

danyill:
> Allow only specific data type in DataSets used within SampledValue streams.
> I first was struggling with it. As the -6 does not restrict anything here. But I think in this case it really will do more harm, not restricting it. I even more so do think we should not use a finder here but on a list with those 8 filterers

Seems fair. I think for the last four you mean `TVTR`.
We also need the daName `q`.

If we restricted to `cdc=SAV` and `fc=MX` would that do? 

It's probably not adequate to produce an acceptable dataset without the user having to "know more" to meet either 9-2LE or 61869-9 requirements.

IEC 61869-9 also imposes constraints on the dataset name `PhsMeasx` where x=1 has a definition.
It also imposes constraints on the SampledValueControl name to `MSVCBxx` where `xx` has a definition.
It also imposes constraints on the order of the values in the dataset, e.g.

>Dataset members shall consist of AmpSv.instMag.i (current sampled value) or VolSv.instMag.i (voltage sampled value) attributes, each followed immediately by the corresponding AmpSv.q or VolSv.q (quality) attribute. The number of current sampled values and the number of voltage sampled values shall match the number of each specified by the variant code for the dataset.

Is compliance with IEC 61869-9 something we should aim for? It is probably the future... 
Or is this a separate plugin?

danyill:
I guess I think DataSet creation for SV traffic needs to be able to be done to meet the requirements of the different standards, possibly in a "pluggable" fashion so that:

* DataSet and SampledValueControl name are consistent with the requirements
* DataSet item order is correct
* Required fields are present

There is probably a "pro mode" required for the SV dataset and SV control as well which "lets you do anything" according to IEC 61850-9-2 (because someone will want to do something different) but at least from a "user point of view" I expect the minimum viable product doesn't require them to read the standard to build a compliant DataSet and SVControl to either 9-2LE, or IEC 61869-9

If we take the "OpenSCD" is strictly an XML editor then I guess generic dataset creation is fine (just not immediately useful with the level of automation I'd expect, I don't want to read international standards before I can build a dataset...).

JakobVogelsang:
I have given this whole topic a bit more, though. Some of your requests can be done very easily. Some need a bit more time to be implemented properly:
1. The attribute `multicast` will be rendered only if the value for it is `false`. This will allow correcting for wrong setting by tools including maybe future plugins within OpenSCD. If it is true, however, we do not render it.
3. The DataSet selection regarding the different standards and guidelines around the SMV need to be addressed with different wizards. I therefore opened another two issues describing those wizards. 
4. We need however something that is generic enough to cope with the 9-2 as is. I therefore would propose to adapt the data selection and restrict it to `SAV` type value as @danyill proposed.

JakobVogelsang:
@danyill  I have added some changes
1. Hide `multicast` when `true`
2. adapt the data picker to `lnClass` `TVTR` and `TCTR`, data object `cdc` to `SAV`, data attribute name `instMag` and `q`

The other issue we will tackle in other PRs including having a separate create wizard to 9-2LE and 61968-9


I in the meantime will add unit and integration tests so, that the code review can be started

danyill:
Thanks for your consideration.

I did what I thought was a typical example:

image

And found I was getting an error in the console:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'value')
    at getValue (foundation.js:99:18)
    at sampledvaluecontrol.js:163:39
    at Array.forEach (<anonymous>)
    at sampledvaluecontrol.js:162:29
    at WizardDialog.act (wizard-dialog.js:170:27)
    at WizardDialog.<anonymous> (wizard-dialog.js:286:33)
    at EventPart.handleEvent (lit-html-c8d21c9b.js:1059:24)
    at Button.EventPart.__boundHandleEvent (lit-html-c8d21c9b.js:1025:47)
```

I think this is related to the multicast attribute missing.

JakobVogelsang:
That was my bet, I did fix it for the update action but not for the create action. :)

ca-d (commented):
I'm having some difficulty validating the behaviour of this plugin against the provided `sampledvaluecontrol.scd` test file:

Upon initially opening the file, both the Template section (`nsd`) and the position of the `ReportControl` element in line 178 seem to be invalid:

initial-errors

When trying to add an SMVControl to `IED1`, no DAs can be selected on the last wizad page:

no-das

The SMVControl can still be added successfully. However, after adding it, the `SCL`'s `release` attribute is invalid, we have an invalid `DataSet` (no `DA`s), and I get complaints that `securityEnabled` and `synchSourceId` are not allowed. I'm especially distraught by the "fixed value constraint" apparently coming about on the `SCL` element due to the change to a single `IED`, I can't explain it to myself.

JakobVogelsang:
Ola Senor,

I can comment on some of the issues you are seeing maybe giving you a direction.

1.) The file is invalid. I did not focus to make it schema valid but to make a snippet to test SampleValueControl. This has been a mistake I see now ad those files are a good source for others to review the functionality. 

2.) You are not seeing any data as the IED you are testing it with does not contain any data objects SAV. The IED3 will show you some. This shows me however that we need so kind of message here exposing what kind of data is accepted, we do restrict.

3.) Allowing to add a SampledValueControl with a DataSet but without FCDA is debatable. There are other options. Not adding a DataSet which at the moment collides with your functionality as we cannot add a DataSet without any Control blocks and also cannot assign/reassign DataSet to control blocks. The other option is to not allow adding a SampledValueControl block without selected data. 

3.) The issue regarding the fixes release value is indeed strange. My guess is that the attributes version, revision and release get corrupted somehow. I think the best way is to save the SCL after editing and have a look at those. Due to those broken attributes the validator thinks this is an Ed1 file and that then causes all the issues you are seeing.

Everything starts in Linz they say 😆

ca-d (approved):
LGTM, I'll create a new issue for the schema version bug since I couldn't find its root quickly.

PR #745, opened 2022-05-13
feat(editor/ied): Show the technical path to a DO/DA

### LINKED ISSUE #554: Show the technical path to a DO/DA
It could be useful to show the technical path to the user. For example:
image

Will be:
VI3P1_OperationalValues/PPRE_MMXU1.PPV.PhsAB.instCVal.mag.f

Flurb:
Closes #554

### DIFF
--- a/src/editors/IED.ts
+++ b/src/editors/IED.ts
@@ -69,9 +69,9 @@ export default class IedPlugin extends LitElement {
     if (iedList.length > 0) {
       return html `
         <section>
-          <div>
+          <div class="header">
             <mwc-select
-              id="iedSelect"
+              class="iedSelect"
               label="${translate("iededitor.searchHelper")}"
               @selected=${this.onSelect}>
               ${iedList.map(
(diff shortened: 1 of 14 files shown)


danyill:
I like it! But I do wish I could click on it to copy and paste the breadcrumbs without them disappearing (when the item loses focus) :wink: 

I also wonder if the selected IED and the breadcrumb should be fixed to the top of the screen (as part of the editor header). Recently I learnt about [scroll-margin-top](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) which might be useful here (if I understand correctly) if this was even a good idea.

Flurb:
> I like it! But I do wish I could click on it to copy and paste the breadcrumbs without them disappearing (when the item loses focus) :wink: 
> 
> 
> 
> I also wonder if the selected IED and the breadcrumb should be fixed to the top of the screen (as part of the editor header). Recently I learnt about [scroll-margin-top](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) which might be useful here (if I understand correctly) if this was even a good idea.
> 
> 
> 
> 

Having a copy function wasn't part of the story, but will be added in the future 😄

dlabordus (approved):

dlabordus (approved):

Issue #746, opened 2022-05-14
Cleanup plugin suffers from an excess of buttons


danyill:
**Is your feature request related to a problem? Please describe.**

As a user of OpenSCD, I don't wish to be overwhelmed by buttons. The Cleanup plugin has an edit button against each list item. That's way too many buttons.

**Describe the solution you'd like**

I'd like a thing so that if an item is selected or I hover over an item I get an edit button as the cool kids did in #718

**Requirements**

It should be simple.

**Suggestions for a specific user interaction**

If I hover over the list item I should see the button or - if I select the list item I should see the button. But I wonder which is better?
A couple of PRs momentarily.

PR #747, opened 2022-05-14
feat(editors/cleanup) Provide appearing/disappearing edit button on hover

danyill:
This is the CSS solution to #746. I quite like it. But I would like some feedback.
I haven't applied for the dataset cleanup yet until we can decide between this and the other PR.


danyill:
I also rearranged the icons here so the edit button is the last icon. I have not updated snapshots or anything.

PR #748, opened 2022-05-14
 Provide appearing/disappearing edit button on select

### LINKED ISSUE #746: Cleanup plugin suffers from an excess of buttons

danyill:
This is the alternative to #747 which I have basically copied without fully understanding #718.
Is this better? It certainly requires a much more sophisticated mechanism...
This would close #746

### DIFF
--- a/src/editors/cleanup/control-blocks-container.ts
+++ b/src/editors/cleanup/control-blocks-container.ts
@@ -45,6 +45,12 @@ import {
 } from '../../wizards/sampledvaluecontrol.js';
 import { cleanSCLItems, identitySort } from './foundation.js';
 
+interface ControlBlockSelectDetail {
+  controlBlock: Element | undefined;
+}
+
+type ControlBlockSelectEvent = CustomEvent<ControlBlockSelectDetail>;
+
 type controlType =
   | 'GSEControl'
   | 'LogControl'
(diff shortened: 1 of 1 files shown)


JakobVogelsang:
You can use the same solution as you did with `hover` using pure CSS that would look something like this:

```
.checkListItem mwc-icon-button {
    display: none;
}

.checkListItem:focus mwc-icon-button {
   display: inline-block;
}
```

Btw I think it is worth trying to put the edit button into the meta slot as with the subscriber editor. That would require to put the checkbox of the list item to the left which is possible. I think we need to add this to the filtered-list but is a quick fix. 

To be honerst with you I think the hover version is better. More intuitive and working better with check-list-item's

danyill:
>To be honerst with you I think the hover version is better. More intuitive and working better with check-list-item's

Thanks. Deal!

> You can use the same solution as you did with hover

Ha. I think I used focussed rather than focus. That was my mistake.

> Btw I think it is worth trying to put the edit button into the meta slot as with the subscriber editor. That would require to put the checkbox of the list item to the left which is possible. I think we need to add this to the filtered-list but is a quick fix.

I'm not sure this is possible with a graphic on a checklist. I think when `left` is set, that overrides the graphic.

PR #749, opened 2022-05-16
fix(menu/subscriberinfo): fix lnInst attribute for lnClass LLN0

JakobVogelsang:
I have found this bug preparing my presentation and just fixed it right away:

> When creating a `IEDName` element referencing `ExtRef` within `LN0` parent element in other IEDs the standard says the `lnInst` must be missing. At the moment we however do set it to a empty string.


ca-d (approved):
LGTM, thanks for the test name clarifications!

PR #752, opened 2022-05-16
feat(action-icon): add hideActions property

JakobVogelsang:
With read-only `l-node-editor` it became obvious that we need to disable transform CSS rules designed to adapt to floating action buttons. This PR is adding this feature to the `action-icon` component

### DIFF
--- a/src/action-icon.ts
+++ b/src/action-icon.ts
@@ -33,7 +33,7 @@ export class ActionIcon extends LitElement {
   highlighted = false;
   /** disables CSS adoption to action buttons */
   @property({ type: Boolean })
-  noaction = false;
+  hideActions = false;
 
   async firstUpdated(): Promise<void> {
     this.tabIndex = 0;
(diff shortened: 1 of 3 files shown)


src/action-icon.ts
```diff
@@ -31,6 +31,9 @@ export class ActionIcon extends LitElement {
   /** highlight pane with dotted outline */
   @property({ type: Boolean })
   highlighted = false;
+  /** disables CSS adoption to action buttons */
+  @property({ type: Boolean })
+  noaction = false;
```
ca-d, line 36: I prefer calling this `hideActions` because that's what `mwc-dialog` does for exactly the same purpose.

If you prefer this property name, why not use the camelCase `noAction`?

ca-d (changes requested):
LGTM apart from the action property name, where I have a suggestion.

In general I still don't quite like the styling of highlighted action icons (where the border touches the label text without any distance between them) and also I still find it weird that the click targets for widely spaced action icons extend so far beyond the actual icon. I'd prefer if we could have smaller click targets. But these two should probably be separate bug fix issues.

src/action-icon.ts
ca-d: PS: in the comment you probably mean adaption rather than adoption.

ca-d (approved):

ca-d:
Please merge as soon as you wish whether you still adapt the property name or not.

JakobVogelsang:
Regarding the other issue that you have, mentioned. Would you be so kind and create issues for those?

PR #753, opened 2022-05-16
fix(wizards/conductingequipment): on create earth switch add missing ground

JakobVogelsang:
When OpenSCD add a new earth switch, it does add a `Terminal` pointing to a `ConnectivityNode` with name `grounded`. It does not check, however, if such a `ConnectivityNode` does exist. This PR extends this with the following logic:

1. check if there is a `ConnecitvityNode` with name `grounded` in the same `VoltageLevel`
2. if is there use the path in the `Terminal`
3. if not add such a `ConnectivityNode`

### DIFF
--- a/src/wizards/conductingequipment.ts
+++ b/src/wizards/conductingequipment.ts
@@ -248,13 +248,36 @@ export function createAction(parent: Element): WizardActor {
       desc,
     });
 
-    if (proxyType === 'ERS')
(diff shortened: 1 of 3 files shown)


ca-d (approved):
LGTM, thanks for so clearly describing the bug, made it quite easy for me to review without any standard related questions.

PR #757, opened 2022-05-18
feat(wizards/eqsubfunction): add create wizard

JakobVogelsang:
The last of the four completing create wizards for function related editors


src/wizards/eqsubfunction.ts
```diff
@@ -0,0 +1,56 @@
...
+      primary: {
+        icon: 'save',
+        label: get('save'),
+        action: createEqSubFunctionAction(parent),
+      },
+      content: [
+        ...contentFunctionWizard({
+          name,
+          desc,
+          type,
```
ca-d, line 50: Is `type` here really just an arbitrary string or does it need any restrictions? Please excuse the stupid question, I was too lazy to look it up in the `pdf` .

ca-d (approved):
LGTM!

However, on testing I found out that the `functions.scd` test file is slightly invalid:

```
functions.scd:9 Function  (Element '{http://www.iec.ch/61850/2003/SCL}Function')
```

is an unexpected element:

```
This element is not expected. Expected is one of ( ##other{http://www.iec.ch/61850/2003/SCL}*, {http://www.iec.ch/61850/2003/SCL}Text, {http://www.iec.ch/61850/2003/SCL}Private, {http://www.iec.ch/61850/2003/SCL}LNode, {http://www.iec.ch/61850/2003/SCL}PowerTransformer, {http://www.iec.ch/61850/2003/SCL}GeneralEquipment, {http://www.iec.ch/61850/2003/SCL}VoltageLevel ).
```

src/wizards/eqsubfunction.ts
JakobVogelsang: It is indeed just a string. I think this just helps to define some kind of groups/types to functions. I honestly don't know if this is used at the moment by someone :)

PR #759, opened 2022-05-19
ci: delete travis.yml

ca-d:
Since we're no longer running tests on Travis CI we should no longer need this file IMO.


JakobVogelsang (approved):
This is very true!

Issue #332, opened 2021-10-11
Allow editing `Function`, `SubFucntion`, `EqFunction` and `EqSubFunction` elements attributes


JakobVogelsang:
As a system specification engineer, I want to edit the attributes of SCL elements `Function`, `SubFunction`, `EqFunction` and `EqSubFunction`. 

**Requirements**
- add wizard for editing attribute
- add edit button to the header of the respective function type editors

PR #762, opened 2022-05-20
feat(wizards/function): add edit wizard

JakobVogelsang:
This is one of four PRs covering #332


JakobVogelsang:
> LGTM, but when testing all of these branches I suddenly started noticing that dots appear next to the buttons on the functional level, as if the underline for the `<abbr>` elements would be displayed.
> 
> Can you replicate that? abbr

I do! Sorry for that I missed to add 

```css
abbr {
  text-decoration: none;
  border-bottom: none;
 }
```

PR #766, opened 2022-05-22
feat(editors/subtation): allow instatiation of LNode from LNodeTypes

### LINKED ISSUE #740: Instantiate `LNode` from LNodeType's
As a system specification engineer I want to add `LNode` references based on the `LNodeType` into any of the substation section elements

** Requirements **
- set `LNode`s `iedName` attribute to **None**!!
- do not set `ldInst` attribute
- by default `prefix` shall be `null`
- `lnClass` shall be the one of the selected `LNodeType` `lnClass` attribute
- `lnType` shall be the `id` of the `LNodeType` element
- `lnInst` shall be the minimal unique number for a `lnClass`. E.g. if there are already 5 `CSWI` within the same element, and they have 1,2,4,5,10 the new one shall be set to 3
- for `lnClass` `LLN0` the `lnInst` shall be empty string
 
** Mockup **
740_LNodeTypeSelector

JakobVogelsang:
Closes #740 

This wizard is straight forward, as are its test. 

At the moment the other `LNode` type wizard is used both as edit and as create wizard. With this PR it will be an edit wizard only. I do think it is logically correct as for a top-down procedure usually engineers will instantiate from LNodeType first, which create the `LNode` initially, and in the process of creating an instantiated virtual IED will complete the `LNode` pointing to both the LNodeType and the LN/LN0. 

However, I am very open for discussion here :)

UPDATE: After some discussion, the existing LNode type wizard and the new wizard implemented here are combined. The entry point LNode wizard chooses based in the parent tag name which one fits best. In addition to that the user can navigate between the two with using the menu action.


src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+function getUniqueLnInst(parent: Element, lnClass: string): number {
+  const lnInsts = Array.from(
+    parent.querySelectorAll(`LNode[lnClass="${lnClass}"]`)
+  )
+    .map(lNode => Number.parseInt(lNode.getAttribute('lnInst')!))
+    .sort((a, b) => a - b);
+
+  if (lnInsts.length === 0) return 1;
+
+  for (let i = 1; i < 99; i++) {
```
ca-d, line 38: Why use `<= 98` here? How many iterations do we need? This and the next line seem buggy.

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+    if (lnInsts[i - 1] !== i) return i;
```
ca-d, line 39: I don't understand this escape condition.

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+  }
+
+  return NaN;
```
ca-d, line 42: Is this how we signal failure here? Why not throw?

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+      .filter(item => item.selected)
+      .map(item => item.value)
+      .map(identity => {
+        return parent.ownerDocument.querySelector(
+          selector('LNodeType', identity)
+        );
+      })
+      .filter(item => item !== null);
+
+    const clonedParent = <Element>parent.cloneNode(true); //for multiple selection of same lnClass
```
ca-d, line 61: Thanks for the comment! Somehow I still don't quite understand the line. Could you explain in a bit more detail?

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+
+    const createActions: Create[] = <Create[]>selectedLNodeTypes
+      .map(selectedLNodeType => {
+        const lnClass = selectedLNodeType.getAttribute('lnClass');
+        if (!lnClass) return null;
+
+        const uniqueLnInst = getUniqueLnInst(clonedParent, lnClass);
+        if (isNaN(uniqueLnInst) || uniqueLnInst > 99) return null;
```
ca-d, line 69: Again: why use `99` here? If it's an SCL significant constant guaranteed to stay the same forever regardless of edition, it belongs in foundation, otherwise at least at the top of the file. Scattering literal constants throughout source code is considered a code smell.

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+
+        const existLLN0 =
```
ca-d, line 71: Could we call this `hasLLN0` or `lLN0Exists` to read more like English?

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+          clonedParent.querySelector('LNode[lnClass="LLN0"]') !== null;
+        if (lnClass === 'LLN0' && existLLN0) return null;
+
+        const lnInst = lnClass === 'LLN0' ? '' : `${uniqueLnInst}`;
+
+        const element = createElement(parent.ownerDocument, 'LNode', {
+          iedName: 'None',
+          ldInst: null,
+          prefix: null,
```
ca-d, line 80: Why do these need to be set to `null` explicitly here?

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+      ],
+      primary: {
+        icon: 'save',
+        label: get('save'),
+        action: createLNodeAction(parent),
+      },
+      content: [
+        html`<filtered-list multi
+          >${lNodeTypes.map(lNodeType => {
+            const isDesabled =
```
ca-d, line 126: Should be `isDisabled`

src/wizards/lnode.ts
```diff
@@ -11,18 +11,146 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
...
+}
+
+function openLNodeReferenceWizard(parent: Element): WizardMenuActor {
+  return (wizard: Element): void => {
+    wizard.dispatchEvent(newWizardEvent());
+    wizard.dispatchEvent(newWizardEvent(lNodeReferenceWizard(parent)));
+  };
+}
+
+function lNodeInstanceWizard(parent: Element): Wizard {
```
ca-d, line 103: As discussed, let's change Instance and Reference to InstRef and TypeRef, respectively.

ca-d (changes requested):

src/translations/de.ts
```diff
@@ -460,8 +460,12 @@ export const de: Translations = {
         selectIEDs: 'Auswahl IEDs',
         selectLDs: 'Auswahl logische Geräte',
         selectLNs: 'Auswahl logische Knoten',
+        selectLNodeTypes: 'Auswahl logische Knoten Type',
       },
       placeholder: 'Bitte laden Sie eine SCL-Datei, die IED-Elemente enthält.',
+      uniquewarning: 'Logische Knoten Klasse existiert bereits',
+      reference: 'Reference auf bestehenden logischen Knoten erstellen',
```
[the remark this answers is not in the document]
JakobVogelsang: It was a mistake :) Thanks for spotting

src/translations/de.ts
```diff
@@ -338,9 +338,9 @@ export const de: Translations = {
       noIedSelected: 'Keine IED ausgewählt',
     },
     view: {
-      publisherView: "Zeigt verbundene IED(s) der ausgewählten GOOSE",
-      subscriberView: "Zeigt verbundene GOOSE(s) des ausgewählten IED"
-    }
+      publisherView: 'Zeigt verbundene IED(s) der ausgewählten GOOSE',
+      subscriberView: 'Zeigt verbundene GOOSE(s) des ausgewählten IED',
```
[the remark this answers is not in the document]
JakobVogelsang: I just opened a new issue and a new PR just to have it right there. #776

src/translations/en.ts
```diff
@@ -457,8 +457,12 @@ export const en = {
         selectIEDs: 'Select IEDs',
         selectLDs: 'Select logical devices',
         selectLNs: 'Select logical nodes',
+        selectLNodeTypes: 'Select logical node types',
       },
       placeholder: 'Please load an SCL file that contains IED elements.',
+      uniquewarning: 'Logical node class already exist',
```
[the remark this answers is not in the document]
JakobVogelsang: True

src/wizards/lnode.ts
JakobVogelsang: The array `lnInsts` is a sorted array containing integers of existing `lnInst`'s.  This excape function makes sure to take the first not used integer between 1 and 99. Say `lnInsts` is `[1, 2, 3, 5, 10, 13, 15]` then the return value would be `4`. 

Sorry, I did not find a more compact and elegant way to implement this.

src/wizards/lnode.ts
JakobVogelsang: This is how we indicate that no unique `lnInst` can be found in the parent element for the specific `lnClass`. But you are right, we should dispatch a log event.

src/wizards/lnode.ts
JakobVogelsang: When importing, say three `CSWI` type logical nodes to a parent that has already two included with `lnInst`'s say 1 and 3. What we need is the algorithm to give the first to be imported `lnInst` 2 the second 4 and the first 5. 

If the do the check on the parent element itsself the algorithm would give all the three `lnInst` 2. As we can add LNode element to the cloned parent we can do the correct check for multiple imports of the same `lnClass`

src/wizards/lnode.ts
JakobVogelsang: Ahhh, I see you would like this to be a constant and not hard coded with the function itself. This makes sense, I will change it

src/wizards/lnode.ts
JakobVogelsang: `ldInst` must be `null` as not pointing to any IED that has logical device as child element. 
`prefix` can be `null` and will be in lots of cases. I think it is a good default. Users can change with the to be developed edit wizard.

JakobVogelsang: This is not really necessary here. It was more or less for me to see it directly in the code. But we do not do that in other parts of the create wizard I will kick it out.

src/wizards/lnode.ts
JakobVogelsang: True

JakobVogelsang:
The last commit LGTM. Thanks for your effort :)

ca-d (approved):
LGTM

PR #769, opened 2022-05-23
refactor(foundation): change menu action API

JakobVogelsang:
This is a small refactor as an output of a discussion between @ca-d and me. We were trying to trigger a wizard, not a sub-wizard from menu actions and thought the API as it is not flexible enough. This is therefore our new proposal :)


ca-d (approved):
LVGTM!

That no tests had to be touched is a great sign for a pure refactor branch 😁👍

PR #778, opened 2022-05-24
feat(wizards/lnode): add edit wizard

### LINKED ISSUE #777: Allow edit prefix and LNode instance
As a system specification engineer, I want to edit `prefix` and `lnInst` when defining instance of logical nodes in the substation section.

** requirements **
- only edit `prefix` and `lnInst` and only when `iedName` is None!

JakobVogelsang:
Closes #777

### DIFF
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -60,6 +60,10 @@ export const de: Translations = {
     sampleRate: 'Abtastrate zu Telegram hinzufügen',
     security: 'Potentiel in Zukunft für z.B. digitale Signature',
     synchSourceId: 'Identität der Zeitquelle zu Telegram hinzufügen',
+    iedName: 'Referenziertes IED',
+    ldInst: 'Referenziertes logisches Gerät',
+    prefix: 'Präfix des logischen Knoten',
+    lnInst: 'Instanz des logischen Knoten',
   },
   settings: {
     title: 'Einstellungen',
--- a/src/translations/de.ts
+++ b/src/translations/de.ts
@@ -62,8 +62,8 @@ export const de: Translations = {
     synchSourceId: 'Identität der Zeitquelle zu Telegram hinzufügen',
     iedName: 'Referenziertes IED',
     ldInst: 'Referenziertes logisches Gerät',
-    prefix: 'Präfix des logischen Knoten',
-    lnInst: 'Instanz des logischen Knoten',
+    prefix: 'Präfix des logischen Knotens',
+    lnInst: 'Instanz des logischen Knotens',
   },
   settings: {
     title: 'Einstellungen',
--- a/src/wizards/lnode.ts
+++ b/src/wizards/lnode.ts
@@ -12,9 +12,11 @@ import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation';
 import '../filtered-list.js';
 import {
   Create,
+  cloneElement,
   createElement,
   EditorAction,
   getChildElementsByTagName,
+  getValue,
   identity,
   isPublic,
   newLogEvent,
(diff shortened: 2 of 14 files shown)


src/wizards/lnode.ts
```diff
@@ -489,3 +492,130 @@ export function lNodeWizard(parent: Element): Wizard {
...
+      helperPersistent
+      disabled
+    ></wizard-textfield>`,
+    html`<wizard-textfield
+      label="ldInst"
+      .maybeValue=${options.ldInst}
+      helper="${translate('scl.ldInst')}"
+      helperPersistent
+      nullable
+      disabled
```
ca-d, line 522: The combination of `nullable` and `disabled` is currently dangerous due to a bug in `wizard-textfield`:

Even when it's `disabled`, as long as it's `nullable` the null switch can be toggled, and by toggling it back and forth we even enable the TextField itself, allowing the user to enter arbitrary values.

src/translations/de.ts
```diff
@@ -60,6 +60,10 @@ export const de: Translations = {
     sampleRate: 'Abtastrate zu Telegram hinzufügen',
     security: 'Potentiel in Zukunft für z.B. digitale Signature',
     synchSourceId: 'Identität der Zeitquelle zu Telegram hinzufügen',
+    iedName: 'Referenziertes IED',
+    ldInst: 'Referenziertes logisches Gerät',
+    prefix: 'Präfix des logischen Knoten',
+    lnInst: 'Instanz des logischen Knoten',
```
ca-d, line 66: I'd use "des Knotens" here instead of "des Knoten", I think that's the more common form.

ca-d (changes requested):
LGTM apart from the 'wizard-textfield` bug.

src/translations/de.ts
JakobVogelsang: True! Thanks for the hint

src/wizards/lnode.ts
JakobVogelsang: Thanks for the thorough review. I did not see that...
Me, I'd not really want to have attributes like `ldInst` be editable and therefore would like to have the fix upfront and rebase on this one rather than removing the `disabled` attribute from the `wizard-textfield`s at this point.

What would you say be the best procedure?

JakobVogelsang: Is fixed with #781

src/wizards/lnode.ts
```diff
@@ -489,3 +492,130 @@ export function lNodeWizard(parent: Element): Wizard {
...
+
+function contentLNodeWizard(options: ContentOptions): TemplateResult[] {
+  const isIedRef = options.iedName !== 'None';
+
+  return [
+    html`<wizard-textfield
+      label="iedName"
+      .maybeValue=${options.iedName}
+      helper="${translate('scl.iedName')}"
+      helperPersistent
```
Flurb, line 513: When do we want to have a persistent helper text, and when do we not want to have it?
I'm not using the persistent helper text at all, is there a common rule for this?

JakobVogelsang: This is needed here as we have pure read only text field. The helper text usually only appears when text fields are focused, and this for a reason. In this particular case they cannot be focused and that is why I decided to show the helper text permanently.

Flurb: Good point, thanks for the explanation!

Flurb (approved):

src/wizards/lnode.ts
JakobVogelsang: I think we do not need a common rule here. I depend on the importance of the helper text to understand the input. If so and it is read-only - set to `disabled` - we show permanently show the helper. Otherwise, the component is doing its job just fine :)

ca-d (approved):
LGTM

PR #779, opened 2022-05-25
feat(104Protocol): Refactor list of 104 Address and add wizard to edit existing Address Element

dlabordus:


test/unit/editors/protocol104/doi-container.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  beforeEach(async () => {
+    document = await fetch('/test/testfiles/104/valid-addresses-case1.scd')
+      .then(response => response.text())
+      .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+  });
+
+  describe('DOI with simple Address Element', () => {
+    beforeEach(async () => {
+      doiElement = document.querySelector('IED[name="B1"] DOI[name="CmdBlk"]')!;
```
Flurb, line 21: Good to know I don't have to use the `>` character every time here.

dlabordus: You only use the > character when the element is directly below the parent. 
So in this case it wouldn't work, because DOI isn't directly below IED.

src/editors/protocol104/doi-container.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private openEditTiWizard(): void {
+    this.dispatchEvent(newWizardEvent(showDOIInfoWizard(this.element)));
+  }
+
+  private static get104DetailsLine(address: Element): string {
+    const values = [];
+
+    if (address.hasAttribute('casdu')) values.push(`casdu: ${address.getAttribute('casdu')}`);
```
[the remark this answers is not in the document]
dlabordus: With FCDA is seems to be more usable to do it like that, but I like the solution, so in the next PR it will be included.

PR #781, opened 2022-05-25
fix(wizard-diagram, wizard-select, wizard-checkbox):  disabled attribute

### LINKED ISSUE #783: Disabled nullable wizard inputs can be enabled by toggling the null switch
Disabled nullable wizard inputs can be enabled by toggling the null switch.

**To Reproduce**
1. Open any wizard with a disabled nullable wizard input.
2. Toggle the null switch off and on once.
3. The wizard input is now enabled.

**Expected behavior**
The null switch is disabled as well so that it can't be toggled to begin with.

JakobVogelsang:
The components we are providing `wizard-textfield`, `wizard-select` and `wizard-checkbox` had an incorrect behaviour when `disabled`.

Closes #783 .

### DIFF
--- a/src/wizard-checkbox.ts
+++ b/src/wizard-checkbox.ts
@@ -40,6 +40,9 @@ export class WizardCheckbox extends LitElement {
       this.checked = check === 'true' ? true : false;
     }
   }
+  /** Disables component including null switch */
+  @property({ type: Boolean })
+  disabled = false;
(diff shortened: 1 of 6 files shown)


ca-d (approved):
LGTM, thanks for the FIXME comments reminding us to solve this more radically upon refactor.

PR #782, opened 2022-05-26
feat(wizards/substation/l-node-editor): add duplicate button

### LINKED ISSUE #741: Duplicate `LNode` instance
As a system specification engineer, I want to have a quick process to add multiple instances of the same `LNodeType` when working on a SSD file.

** Requirements **
- add a duplicate button to the l-node-editor
- use icon `content_copy`
- add `LNode` with the same attributes except `lnInst`. This must be incremented to the next free instance number within the same parent

JakobVogelsang:
Closes #741 

This is a useful tool when working on functions with multiple logical node instances of the same `LNodeType` e.g. a distance function that consists of 5 to 7 zones where each zone is represented with a logical node `PDIS` and usually is the same structure.

### DIFF
--- a/src/editors/substation/l-node-editor.ts
+++ b/src/editors/substation/l-node-editor.ts
@@ -8,7 +8,13 @@ import {
 } from 'lit-element';
 
 import '../../action-icon.js';
-import { identity, newActionEvent, newWizardEvent } from '../../foundation.js';
+import {
+  cloneElement,
+  identity,
+  newActionEvent,
+  newLnInstGenerator,
+  newWizardEvent,
+} from '../../foundation.js';
 import {
   automationLogicalNode,
   controlLogicalNode,
@@ -75,6 +81,27 @@ export class LNodeEditor extends LitElement {
   private get missingIedReference(): boolean {
     return this.element.getAttribute('iedName') === 'None' ?? false;
   }
+  @state()
+  private get isIedRef(): boolean {
+    return this.element.getAttribute('iedName') !== 'None';
+  }
+
+  private cloneLNodeElement(): void {
+    const lnClass = this.element.getAttribute('lnClass');
+    if (!lnClass) return;
+
+    const uniqueLnInst = newLnInstGenerator(this.element.parentElement!)(
+      lnClass
+    );
+    if (!uniqueLnInst) return;
+
+    const newElement = cloneElement(this.element, { lnInst: uniqueLnInst });
+    this.dispatchEvent(
+      newActionEvent({
+        new: { parent: this.element.parentElement!, element: newElement },
+      })
+    );
+  }
 
   private openEditWizard(): void {
     const wizard = wizards['LNode'].edit(this.element);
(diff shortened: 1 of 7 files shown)


src/editors/substation/l-node-editor.ts
```diff
@@ -8,7 +8,13 @@ import {
 } from 'lit-element';
 
 import '../../action-icon.js';
-import { identity, newActionEvent, newWizardEvent } from '../../foundation.js';
+import {
+  cloneElement,
+  identity,
```
ca-d, line 13: Why do we import identity here?

src/editors/substation/l-node-editor.ts
```diff
@@ -75,6 +81,27 @@ export class LNodeEditor extends LitElement {
   private get missingIedReference(): boolean {
     return this.element.getAttribute('iedName') === 'None' ?? false;
   }
+  @state()
+  private get isIedRef(): boolean {
```
ca-d, line 85: For private internal state only used in a single place I believe we can spare a couple more characters :wink: .

For my taste something like `isInstanceReference` or `isIEDReference` or `referencesLNInstance` or something like that would be preferable. To be clear: I only say this because as a property this has not-completely-local scope, meaning it's not defined exactly where it is used (unlike e.g. a loop index, which I prefer to be as succinct as possible).

src/foundation.ts
```diff
@@ -2756,6 +2758,37 @@ export function getChildElementsByTagName(
   );
 }
 
+/** maxumum allow number for `lnInst` attribute */
```
ca-d, line 2761: I'd say maximum "allowed" or "allowable" or "possible" or just "maximal value of `lnInst` attribute".

src/foundation.ts
```diff
@@ -2756,6 +2758,37 @@ export function getChildElementsByTagName(
...
+const maxLnInst = 99;
+const lnInstRange = Array(maxLnInst)
+  .fill(1)
+  .map((_, i) => `${i + 1}`);
+
+/** generator function returning unique `lnInst` for a `lnClass` `LNode` element */
```
ca-d, line 2767: While this is completely correct, I believe it could use a bit more explanation for the user, like:

```ts
/**
 * @param parent - The LNodes' parent element to be scanned once for `lnInst` values already in use. Be sure to create a new generator every time the children of this element change.
 * @returns a function generating increasing unused `lnInst` values for `lnClass` LNodes within `parent` on subsequent invocations
 */
```

ca-d (approved):
LGTM, apart from some naming issues and one possible (?) stray import.

src/foundation.ts
```diff
@@ -2758,13 +2758,19 @@ export function getChildElementsByTagName(
   );
 }
 
-/** maxumum allow number for `lnInst` attribute */
+/** maxumum value for `lnInst` attribute */
```
ca-d, line 2761: `s/maxumum/maximum/`

PR #787, opened 2022-05-27
refactor(plugging,hosting): make plugin content loading synchronous

### LINKED ISSUE #725: Disappearing element parentNode after await
After `await`ing some promise inside an async method of any`HTMLElement`, both Chrome and Firefox will cause `this.parentNode === null`. This seems to be happening irrespective of both `LitElement` and our plugin system.

For convenience, here is a menu plugin you can use to test out the effect:

```
data:text/javascript;charset=utf-8,export%20default%20class%20TestPlugin%20extends%20HTMLElement%20%7B%0D%0A%20%20async%20run%28%29%20%7B%0D%0A%20%20%20%20const%20parentNode%20%3D%20this.parentNode%3B%0D%0A%20%20%20%20console.warn%28this.parentNode%29%3B%0D%0A%20%20%20%20await%20new%20Promise%28r%20%3D%3E%20setTimeout%28r%2C%201000%29%29%3B%0D%0A%20%20%20%20console.warn%28this.parentNode%29%3B%0D%0A%20%20%20%20console.log%28parentNode%29%3B%0D%0A%20%20%7D%0D%0A%7D
```

```js
export default class TestPlugin extends HTMLElement {
  async run() {
    const parentNode = this.parentNode;
    console.warn(this.parentNode);
    await new Promise(r => setTimeout(r, 1000));
    console.warn(this.parentNode);
    console.log(parentNode);
  }
}
```

You can see that the second `warn` call prints `null` to the console, while the final `log` call prints the same `parentNode` as the first invocation of `warn`.

### Workaround

This example shows that we can avoid using `querySelector('open-scd')` for finding a suitable event target by remembering the element's `parentNode` before awaiting the first promise in the async method.

However, this means we will have to think about how to handle the new `Setting` API in a way in which settings may be set from within async methods after awaiting a promise, something that would not be possible with the current proposal.

ca-d:
Resolves #725 .

This branch resolves our asynchronous plugin loading issues by generating the `tagName` from the `src` URI instead of the class name, thereby obviating the need for the `carehtml` dependency and making the loading of plugins into the DOM synchronous.


ca-d:
@dlabordus @Flurb @JakobVogelsang : Since this refactor substantially changes how we load plugin code and how we generate plugin tag names, please make sure to test this extensively with your own external plugins to make sure everything continues to work as expected! Thank you.

JakobVogelsang:
Oh man, that is so great!! Working in the substation editor feels like x time faster. Also, very nice, the wizard does not cause re-render and the scroll bar position stays. I did however found that logical nodes are not rendered in the IED editor. I must check why this is :)

From code, I did not fully understand why this can be loaded synchronious, I would love to have a session with you about that :) After that I can do some code review :)

JakobVogelsang:
More lines killed that added. A dream refactor :)

src/Plugging.ts
```diff
@@ -156,62 +227,57 @@ export function Plugging<TBase extends new (...args: any[]) => EditingElement>(
...
-              plugin.src,
-              await import(plugin.src).then(mod => mod.default)
-            );
-          return html`<${loadedPlugins.get(plugin.src)}
+        content: staticTagHtml`<${tag}
             .doc=${this.doc}
             .docName=${this.docName}
             .docId=${this.docId}
             .pluginId=${plugin.src}
-            .nsdoc=${await initializeNsdoc()}
```
JakobVogelsang, line 212: I do not see us giving the plugins the `nsdoc` property. This at the moment is required in the IED editor -  which is a bug - BUT it raises the question: when do we want to integrate this now or when creating open-scd?

Other than the IED editor, no other editor seems to be compromised, at least not one developed in the `open-scd` repository.

ca-d: Yeah we have to add that back in, I forgot. Thanks for pointing it out!

The simplest thing to do for the time being would be to just pass the `nsdoc` Promise to the plugin and have it await the `nsdoc` where needed. That would be something we can pass synchronously.

Long term (after The Grand Core Rewrite) I would incorporate the `*.nsdoc` files into our new `docs` property alongside all other XML documents (`*.xsd`, `*.nsd`, `*.scd` etc). Then we have a unified way of accessing all XML document contents by file name.

ca-d:
> I did not fully understand why this can be loaded synchronious

Basically, since we now generate our tag name from the URL rather than the class name, we do not have to wait for the class to be (asynchronously) imported in order to generate the HTML template element, because we know the URL ahead of time. We can therefore render the HTML template just from the static information in the plugins.js file, so that the only asynchronous task that remains is actually registering the custom element under its predetermined tag name.

Does this explanation help?

src/Plugging.ts
JakobVogelsang: I like the solution for now and the one for the `open-scd-core` refactor has been discussed and agreed on as well so far I remember. So good to go for me 👍

JakobVogelsang:
> > I did not fully understand why this can be loaded synchronious
> 
> Basically, since we now generate our tag name from the URL rather than the class name, we do not have to wait for the class to be (asynchronously) imported in order to generate the HTML template element, because we know the URL ahead of time. We can therefore render the HTML template just from the static information in the plugins.js file, so that the only asynchronous task that remains is actually registering the custom element under its predetermined tag name.
> 
> Does this explanation help?

Yes and no. I actually did understand that. And looking through the code, the only thing that I am not capable to understand is the `pluginTag` function. :) and the one I would like to :)

src/Plugging.ts
ca-d: Please disregard my above comment and the last commit which adds this change.

I just had a look at the actual `nsdoc.ts` code for the first time and I have no idea why this is actually a Promise or what we need to await there. If it's actually unnecessary we should fix that file in this branch and pass it synchronously again so that plugins don't need to be changed.

ca-d:
> I have fixed the missing `nsdoc` and the IED editor works again. However, with loaded *.nsdoc files OpenSCD becomes terribly slow. We need to further investigate here :)

Maybe the issue is related to the fact that the `initializeNsdoc` function is being invoked on every render? Do we actually need that function? Could we just invoke it once at the top of the file and pass the `nsdoc` statically?

src/Plugging.ts
```diff
@@ -23,10 +20,72 @@ import { Select } from '@material/mwc-select';
...
-import { initializeNsdoc } from './foundation/nsdoc.js';
+import { Nsdoc } from './foundation/nsdoc.js';
+
+const pluginTags = new Map<string, string>();
+/**
+ * Hashes `uri` using cyrb64 analogous to
+ * https://github.com/bryc/code/blob/master/jshash/experimental/cyrb53.js .
+ * @returns a valid customElement tagName containing the URI hash.
+ */
+function pluginTag(uri: string) {
```
JakobVogelsang, line 34: Isn't a return type missing here? do I miss something?

src/Plugging.ts
```diff
@@ -23,10 +20,72 @@ import { Select } from '@material/mwc-select';
...
+
+/**
+ * Passes its arguments to LitElement's `html` tag after baking the first and
+ * last binding into the first two and last two static strings, respectively.
+ * Throws unless the first and last bindings contain identical strings.
+ */
+function staticTagHtml(
+  oldStrings: ReadonlyArray<string>,
+  ...oldArgs: unknown[]
+) {
```
JakobVogelsang, line 67: Same here: A return type would help to read this not easy to understand function.

src/Plugging.ts
```diff
@@ -23,10 +20,72 @@ import { Select } from '@material/mwc-select';
...
+        ((h2 >>> 0).toString(16).padStart(8, '0') +
+          (h1 >>> 0).toString(16).padStart(8, '0'))
+    );
+  }
+  return pluginTags.get(uri)!;
+}
+
+/**
+ * Passes its arguments to LitElement's `html` tag after baking the first and
+ * last binding into the first two and last two static strings, respectively.
```
JakobVogelsang, line 61: May a link to the MDN doc would be helpful to understand the comment

JakobVogelsang (approved):
LGTM

Sander3003:
Great improvement, OpenSCD is much faster! Even big SCL files are loading much faster! Thanks @ca-d !

src/Plugging.ts
ca-d: Return type `string` is inferred here.

src/Plugging.ts
ca-d: Added the return types.

PR #800, opened 2022-05-31
feat(general): prevent losing data on user navigation

### LINKED ISSUE #743: Losing data when a user click the back button in the browser
As an engineer
I want to be informed to save the data first before I click the "back" button in modern browers
So I do not lose my changes

Show a warning when a user want to back in the browser.

**Requirements**
Only show the warning if a document is open in OpenSCD

A pop-up window should show up:
Are you sure? Your work will be lost

**Additional information**
Example:
https://stackoverflow.com/questions/38306510/alert-on-browser-back-button-click-or-page-refresh

juancho0202:
closes #743

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,8 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
-  <meta name="Description" content="Open Substation Communication Designer is a bottom-up system configuration tool for projects described using IEC 61850-6 Edition 2 or greater.">
+  <meta name="Description"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -37,6 +38,15 @@
     if ('serviceWorker' in navigator)
       navigator.serviceWorker.register('/sw.js');
   </script>
+  <script>
+    const openScd = document.querySelector("open-scd");
+    addEventListener("beforeunload", e => {
```
dlabordus, line 43: @ca-d, Should this code be inside the OpenSCD Component, so that it will always work if the component is used?

ca-d: @dlabordus In order to be able to embed `open-scd` in other web applications that may do their own thing, I'd prefer not to register any event listeners on the global `window` object from the `open-scd` component.

dlabordus (approved):

PR #802, opened 2022-06-01
refactor: remove OpenSCD as dispatcher

JakobVogelsang:
After having the ability to load plugins synchronously, we can now get rid of OpenSCD as the dispatcher within the plugins.
In addition to that this PR removes OpenSCD as dispatcher form Settings as well which was not necessary from the beginning.


src/editors/ied/da-wizard.ts
```diff
@@ -198,26 +204,10 @@ export function createDaInfoWizard(
...
-        .querySelector('open-scd')!
-        .dispatchEvent(newWizardEvent());
-      return [];
-    };
-  }
-
   return [
     {
       title: get('iededitor.wizard.daTitle'),
-      secondary: {
```
dlabordus, line 213: This changes the label from 'Close' to 'Cancel'. Because these dialogs don't have any action related, close is better. The user is not cancelling anything.

src/editors/ied/do-wizard.ts
```diff
@@ -153,26 +159,10 @@ export function createDoInfoWizard(
...
-        .querySelector('open-scd')!
-        .dispatchEvent(newWizardEvent());
-      return [];
-    };
-  }
-
   return [
     {
       title: get('iededitor.wizard.doTitle'),
-      secondary: {
```
dlabordus, line 168: This changes the label from 'Close' to 'Cancel'. Because these dialogs don't have any action related, close is better. The user is not cancelling anything.

dlabordus (changes requested):

dlabordus:
@JakobVogelsang: small remark, watch out to change the formatting to much, this makes reviewing much harder, because there are a lot of changes and the ones that really matter can be missed.

src/editors/ied/do-wizard.ts
ca-d: I suggest we fix this by switching the translation from "Cancel" to "Close" in general. Many wizards don't cancel something when you close them.

test/unit/validators/ValidateTemplates.test.ts
```diff
@@ -18,31 +14,39 @@ describe('ValidateTemplates', () => {
...
 
-      await fixture(html`
-        <open-scd .doc=${doc}
-          ><validate-templates .doc=${doc}></validate-templates
-        ></open-scd>
+      const parent = await fixture(html`
+        <mock-editor-logger
+          ><validate-templates
+            .doc=${doc}
+            .pluginId="${'/src/validators/ValidateTemplates.js'}"
```
ca-d, line 36: Why do we need to set a `pluginId` here? Why didn't we need to do this before? Why set this string as a property rather than as an attribute?

ca-d (changes requested):

src/editors/ied/do-wizard.ts
dlabordus: I think it depends how a user looks at it. When I click on something like "Edit Substation", it's quit normal to press "Cancel" if you want to stop editing the substation. 
But in these cases the user click on "Show Info" or something like that. So in these 2 cases for me "Cancel" seems a little bit strange, because I just want to close the info wizard.

test/unit/validators/ValidateTemplates.test.ts
ca-d: We used to set this lower down. It's needed for the current validation API. We can probably change this in `open-scd-core` along the lines of the new `Setting` API.

src/editors/ied/do-wizard.ts
ca-d: I have been wanting to change the wording in wizards from "cancel" to "close" for quite a while, because we really can't assume that closing a wizard in general will cancel some action. At least some of the time, an action will have already been committed by an interactive element in the wizard or a subwizard at the point where the dialog is closed. This has lead to confusion before. I just added a commit with this suggested change. Please let me know what you think about the result!

dlabordus (approved):

dlabordus (approved):

PR #806, opened 2022-06-03
feat(menu/VirtualTemplateIED): automatically create virtual IEDs

### LINKED ISSUE #742: Transform function structure to server structure (function to logical devices).
As a system specification engineer, I want to further specify the allocation of function to specific IEDs.

** Requirement **
- every function - `Function` or `EqFunction` - and non leaf sub-function  - `SubFunction` and `EqSubFunction` is transformed into a logical device `LDevice` where functions `name` attribute is transformed to logical device `ldInst` attribute.
- `name` and `ldInst` can have different pattern. Indicate if transformation is not possible
- if any of the function name is not identical, use parent name(s) until a unique logical device name can be constructed
- function hierarchy shall be represented with the `grRef` 
- leaf sub-function (sub function with only one `LNode`) name shall be tagged with in the prefix in case the `LNode` has not prefix defined
- Allow to set `manufacturer`. E.g. the name of the organization creating the virtual IED
- set name attribute of IED element to `SPECIFICATION` 

** Mockup **
mock-LNode-selection-for-virtual-IED excalidraw

** Background **
The whole purpose of functional naming in SCL is to instantiate and allocate function aka logical nodes within/into the substation. Or in other words, to formulate how many instances of a specific function are needed - for what purpose those are needed. 

This can be done within the SSD file using the function structures* nested into the process structure**. 

After having done that, the question arises, who decide about the allocation of the instantiated functions - logical nodes - to the real IED that host the functions. There are two possible ways to do that:
1. The user hands over the SSD file to the IED vendor, and the vendor makes a proposal back to the user. The proposal potentially answering the questions: can it be done at all, how many IED are required and which IED cover which functions - logical nodes. 
2. The user goes one step further and formulate so called virtual IED and hands those over to the vendor. 
>The purpose of the issue is to allow the user to create virtual IED based on the function structure* in the substation section and thus speeding up this process.

- * function structure:  elements `Function`, `SubFunction`, `EqFunction` and `EqSubFunction`
- ** process structure: elements `Substation`, `VoltageLevel`, `Bay`, `ConductingEquipment`, `PowerTransformer`, `TransformerWinding` etc.

JakobVogelsang:
This is it, the plugin that allows to automatically create a virtual template IED - name `SPECIFICATION` - based on the function structure in the substation section. 

I decided to pack it into a menu type plugin for not. When we will extend the functionality around the virtual IED creation in the future, we can convert it into an editor type plugin. 

> To test, use the *.ssd file in the testfiles folder.

Closes #742

### DIFF
--- a/src/foundation.ts
+++ b/src/foundation.ts
@@ -598,19 +598,21 @@ function lNodeSelector(tagName: SCLTag, identity: string): string {
   if (identity.endsWith(')')) {
     const [parentIdentity, childIdentity] = pathParts(identity);
     const [lnClass, lnType] = childIdentity
-      .substring(1, identity.length - 2)
+      .substring(1, childIdentity.length - 1)
       .split(' ');
 
     if (!lnClass || !lnType) return voidSelector;
 
-    return tags[tagName].parents
-      .map(
-        parentTag =>
-          `${selector(
-            parentTag,
-            parentIdentity
-          )}>${tagName}[iedName="None"][lnClass="${lnClass}"][lnType="${lnType}"]`
-      )
+    const parentSelectors = tags[tagName].parents.flatMap(parentTag =>
+      selector(parentTag, parentIdentity).split(',')
+    );
+
+    return crossProduct(
+      parentSelectors,
+      ['>'],
+      [`${tagName}[iedName="None"][lnClass="${lnClass}"][lnType="${lnType}"]`]
+    )
+      .map(strings => strings.join(''))
       .join(',');
   }
 
--- a/public/js/plugins.js
+++ b/public/js/plugins.js
@@ -113,6 +113,15 @@ export const officialPlugins = [
     position: 'middle',
   },
   {
+    name: 'Create Virtual IED',
+    src: '/src/menu/VirtualTemplateIED.js',
+    icon: 'developer_board',
+    default: false,
+    kind: 'menu',
+    requireDoc: true,
+    position: 'middle'
+  },
+  {
     name: 'Subscriber Update',
     src: '/src/menu/SubscriberInfo.js',
     default: true,
--- a/src/menu/VirtualTemplateIED.ts
+++ b/src/menu/VirtualTemplateIED.ts
@@ -1,0 +1,259 @@
+import { html, LitElement, property, TemplateResult } from 'lit-element';
+import { get } from 'lit-translate';
+
+import '@material/mwc-dialog';
+import '@material/mwc-list';
+import '@material/mwc-list/mwc-list-item';
+import '@material/mwc-list/mwc-check-list-item';
+import '@material/mwc-list/mwc-radio-list-item';
+
+import '../filtered-list.js';
+import {
+  EditorAction,
+  getChildElementsByTagName,
+  getValue,
+  identity,
+  newWizardEvent,
+  selector,
+  Wizard,
+  WizardActor,
+  WizardInputElement,
+} from '../foundation.js';
+import { CheckListItem } from '@material/mwc-list/mwc-check-list-item';
+import {
+  getFunctionNamingPrefix,
+  getNonLeafParent,
+  getSpecificationIED,
+  getUniqueFunctionName,
+  LDeviceDescription,
+} from './virtualtemplateied/foundation.js';
+
+export type FunctionElementDescription = {
+  uniqueName: string;
+  lNodes: Element[];
+  lln0?: Element;
+};
+
+/** converts FunctionElementDescription's to LDeviceDescription's */
+function getLDeviceDescriptions(
+  functions: Record<string, FunctionElementDescription>,
+  selectedLNodes: Element[],
+  selectedLLN0s: string[]
+): LDeviceDescription[] {
+  const lDeviceDescriptions: LDeviceDescription[] = [];
+
+  Object.values(functions).forEach(functionDescription => {
+    if (
+      functionDescription.lNodes.some(lNode => selectedLNodes.includes(lNode))
+    ) {
+      const lLN0 = selectedLLN0s.find(selectedLLN0 =>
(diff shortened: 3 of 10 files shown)


danyill:
Thank you for this opportunity to provide comment :+1: 
I thought I'd do an initial usability test and fiddle around.

1. The first thing I did was enable the plugin and try to create an IED without any functions or subfunctions present. I had a valid Save button which did nothing.

    image

    Eventually, I remembered to follow the instructions and look at the SSD file and then saw a list of LNs but I think either a message to the user and/or a greyed out button might be appropriate.

2. It feels very odd to me to have a toggle button which functions as a check box for LN0. Is this to indicate that this is mandatory or recommended? I think this is here to ensure that a LDevice always meets at least one mandatory LN0 if it not specified in the Function breakdown...

    image

3. After creating a virtual IED, I thought, "I wonder what that did?" and looked at the IED editor to find an IED exists but with nothing in it? I was using `/test/testfiles/zeroline/functions.scd` which I realise is the incorrect file but it was a fun test.

    image

    In Pro Mode when I examine the source all I see is the container.

    image

    I'm not sure if I have erred in some way. I do not have any console errors.

4. Finally using the correct file in `test/testfiles/virtualied/specificfromfunctions.ssd` I am able to create an IED with LNs! I note that if I don't select a checkbox I can't create an IED but receive no error message (I think there should be some indication to the user as to "why it didn't work").

5. I notice the XML is not pretty-printed for this IED when I click edit in the IED editor. Probably out of scope for this PR I know unless it was to do with the way the elements were added to the document. However most 61850 SCL tools pretty-print output.
    
    image

I will study the standard and then play some more in a little while. These comments are me fumbling around and being pernickety :wink:

ca-d (commented):
The code itself LGTM, but I would still like to talk to @JakobVogelsang about the usability improvements proposed by @danyill .

JakobVogelsang:
Thank you, @danyill, for your valuable inputs. Based on your findings, I would suggest to make two changes:
1.) Save button should be available only if a valid virtual IED can be created
2.) I would replace the radio button with a select field for LLN0 type logical nodes. I think it makes it clearer for the user that there must be something - only one - LLN0 reference. It also differentiates better to the checkbox used for other type of logical nodes.

What do you think about that?

danyill:
> 1.) Save button should be available only if a valid virtual IED can be created


Sounds good.

> 2.) I would replace the radio button with a select field for LLN0 type logical nodes. I think it makes it clearer for the user that there must be something - only one - LLN0 reference. It also differentiates better to the checkbox used for other type of logical nodes.

I am not entirely sure what a select field is.
Perhaps another option is a checkbox which is greyed out and if any of the "children" are selected it is ticked automatically. I think that would be intuitive to the user "Oh yeah, each LDevice must have an LLN0".

I do want to do more testing but it could be done after merge because I have not found the time over the last few weeks and it all seems to work 😞

JakobVogelsang:
> > 1.) Save button should be available only if a valid virtual IED can be created
> 
> Sounds good.
> 
> > 2.) I would replace the radio button with a select field for LLN0 type logical nodes. I think it makes it clearer for the user that there must be something - only one - LLN0 reference. It also differentiates better to the checkbox used for other type of logical nodes.
> 
> I am not entirely sure what a select field is. Perhaps another option is a checkbox which is greyed out and if any of the "children" are selected it is ticked automatically. I think that would be intuitive to the user "Oh yeah, each LDevice must have an LLN0".
> 
> I do want to do more testing but it could be done after merge because I have not found the time over the last few weeks and it all seems to work 😞



> > 1.) Save button should be available only if a valid virtual IED can be created
> 
> Sounds good.
> 
> > 2.) I would replace the radio button with a select field for LLN0 type logical nodes. I think it makes it clearer for the user that there must be something - only one - LLN0 reference. It also differentiates better to the checkbox used for other type of logical nodes.
> 
> I am not entirely sure what a select field is. Perhaps another option is a checkbox which is greyed out and if any of the "children" are selected it is ticked automatically. I think that would be intuitive to the user "Oh yeah, each LDevice must have an LLN0".
> 
> I do want to do more testing but it could be done after merge because I have not found the time over the last few weeks and it all seems to work 😞

A select field is a drop-down element that only shows you the selected element, but hides the other potential options. See for example `mwx-select` or our `wizard-select`. We need something selectable in this function, as there might be more than on `LNodeType` of class `LLN0` in the `DataTypeTemplates` section. In case the `LLN0` is already predefined in the function, we disable the select field to indicate that this is the only option.

JakobVogelsang:
I have decided to remove the use of `wizard-dialog` and use the `mwc-dialog` instead. I don't think this wizard is really general enough to be part of the wizard-library in `open-scd-core`.

src/foundation.ts
```diff
@@ -598,19 +598,21 @@ function lNodeSelector(tagName: SCLTag, identity: string): string {
...
-          )}>${tagName}[iedName="None"][lnClass="${lnClass}"][lnType="${lnType}"]`
-      )
+    const parentSelectors = tags[tagName].parents.flatMap(parentTag =>
+      selector(parentTag, parentIdentity).split(',')
+    );
+
+    return crossProduct(
+      parentSelectors,
+      ['>'],
+      [`${tagName}[iedName="None"][lnClass="${lnClass}"][lnType="${lnType}"]`]
```
ca-d, line 613: Does the `'>'` need to be in a separate array here for some reason or could it go to the beginning of the template literal?

ca-d (approved):
Much nicer now with the improved UI! Thank you!

src/foundation.ts
JakobVogelsang: I had another look. It seems to be a common pattern we use in the selector function within the foundation. I also find it a bit more readable. I would therefore leave it as that. Is this ok with you?

PR #814, opened 2022-06-07
fix(editing): reactively update after changes to doc

### LINKED ISSUE #813: After updating Name in Wizard, Editor doesn't shown updated name
**Describe the bug**
Last something changed in the number of time Editors were refreshed and that seems to have a side-effect.
When updating the name of a Substation in the Substation Editor the change isn't display directly.
After switching between editors the change is displayed.

**To Reproduce**
1. Go to 'Substation Editor'
2. Update the name of the Substation using the standard wizard
3. After the wizard is closed using 'Save' the change isn't displayed in the Editor.
Switching to a other editor and back shows the updated value.

**Expected behavior**
New name should be displayed in the editor.

**Additional context**
-

ca-d:
Closes #813 .

I have locally been having quite some trouble with intermittently failing tests and error messages logged during tests (mostly due to `*.js` files not being available in web-test-runner due to them being called `*.ts` instead). I've fixed quite a few of them, but still have some.

Please advise whether this is a local phenomenon that only appears on my machine or whether you see intermittently failing tests (tests that initially fail but succeed on being re-run) on your machine as well. If tests run fine for *all* of you, I will merge this now and fix my local problems later, otherwise I will keep this pull request open until all local test inconsistencies have been resolved here. Either way, please do not merge this pull request on your own. Thanks!


dlabordus:
_I have locally been having quite some trouble with intermittently failing tests and error messages logged during tests (mostly due to *.js files not being available in web-test-runner due to them being called *.ts instead). I've fixed quite a few of them, but still have some._

I also have this problem since this morning. I see a big change in the package-lock.json also. How did you fix some? Because I don't see anything wrong in the ones going wrong.

dlabordus (approved):

PR #816, opened 2022-06-08
feat(ied-editor/init-dai): Initialize the DOI/SDI/DAI Structure when not available

### LINKED ISSUE #801: Instantiate data attributes in IED
As a system specification engineer, I want to instantiate data attributes in my IED. 

Each of the leaf data attribute nodes in the IED editor can be instantiated. Which means adding a `Val` element to its `DAI` representation with the IED section. This allows the system specification engineer to overwrite the `Val` element in the DataTypeTemplates section with an IED specific values.

** Requirement **
- only leaf nodes can be initialized. Leaf nodes are `DA`/`BDA` elements with a type **NOT** `Struct`
- when initializing a potentially missing parent structure must be added as well that consists of `DOI` | (`SDI`) ... | `DAI`
   e.g. this structure within the `DataTypeTemplates` section MMXU -> A(DO) -> phsA(SDO) -> cVal(DA) -> mag(BDA) -> f (BDA) would translate like that
```xml
 <DOI name="A">
   <SDI name="phsA">
      <SDI name="cVal">
         <SDI name="mag">
            <DAI name="f">
            </DAI>
         </SDI>
      </SDI>
   </SDI>
</DOI>
```
- in the upper example, either no, some or the whole parent structure might be initialized already. All the three cases must be covered. 
- add a create wizard that allows to specify the `Val` inner text
- the input shall depend on the `bType` 
- use `playlist_add` icon to indicate that a leaf data attribute can be instantiated and allocate in the leaf data attribute da-container header
- do not show the `playlist_add` icon if the data is already instatiated

dlabordus:
closes #801


Flurb (approved):

ca-d (commented):
I did not have time to really review or test this code, but it's already merged into main.

I would absolutely love it if we could find some way to slow down our "review and merge" process so that there is time for in-depth code review helping us raise the quality of our contributions. What do you think?

PR #822, opened 2022-06-14
feat(wizards/ied): added new textfields for IED properties 

juancho0202:
Addition of \<wizard-textfield> elements for the following IED properties:
- type
- manufacturer
- configVersion
- originalSclVersion*
- engRight
- owner

\* To display the information regarding SCL Schema it was assumed that there is no value in showing in separate fields for _version_, _revision_ and _release_. Instead, a function to concat the three values was included here. If this assumption is wrong, I will go back and create correspondent fields for revision and release to render them separately.

Since this is my second contribution to open-scd I would expect having missed something in the way, hence the "Reviewed: Help wanted" tag. I appreciate very much your feedback on my approach to this issue. Thanks!


src/wizards/ied.ts
```diff
@@ -78,6 +120,13 @@ function renderIEDReferencesWizard(references: Delete[]): TemplateResult[] {
   ];
 }
 
+function validatedVersionAttribute(element: Element): string | null {
```
dlabordus, line 123: Is it not possible to do something like 
`
return (element.getAttribute('originalSclVersion') ?? '')
  .concat(element.getAttribute('originalSclRevision') ?? '')
  .concat(element.getAttribute('originalSclRelease') ?? '')
`

Because it now displays what is in the XML, even if someone did something strange, but not undefined/null if the attribute isn't there.

test/unit/wizards/ied.test.ts
```diff
@@ -46,7 +46,30 @@ describe('Wizards for SCL element IED', () => {
       await element.requestUpdate();
       inputs = Array.from(element.wizardUI.inputs);
     });
-
+    it('contains a wizard-textfield with a non-empty "type" value', async () => {
+      expect((<WizardTextField[]>inputs).find(textField => textField.label == "type")?.value).not.to
+        .be("-");
```
dlabordus: Probably should be "to.be.equal('-')"

src/wizards/ied.ts
juancho0202: I like your take because for every other property we are actually showing whatever is in the XML even if someone did something strange, so it seems coherent to do the same for this one as well.

test/unit/wizards/ied.test.ts
juancho0202: Yes, correct. I fixed it in the last push.

dlabordus (approved):

PR #834, opened 2022-06-23
feat(104-plugin): New 104 Plugin to configure 104 Values and Networks

dlabordus:
The new 104 plugin to add/edit Private Address Elements for the 104 protocol and configure 104 networks (with or without redundancy)


src/editors/protocol104/base-container.ts
```diff
@@ -0,0 +1,6 @@
+import { LitElement, property } from 'lit-element';
+
+export class Base104Container extends LitElement {
+  @property()
+  doc!: XMLDocument;
```
ca-d, line 5: Does it really make sense to refactor out this one property definition into its own "Base Class"? Is the `Base104Container` a sensible abstraction? If it's really needed, could we find a more descriptive name for it, like `LitElementWithDocProperty` or something?

src/editors/protocol104/foundation/p-types.ts
```diff
@@ -0,0 +1,59 @@
...
+  'StationType': 'protocol104.network.connectedAp.wizard.stationTypeHelper',
+  'IP': 'protocol104.network.connectedAp.wizard.ipHelper',
+  'IP-SUBNET': 'protocol104.network.connectedAp.wizard.ipSubnetHelper',
+  'W-FACTOR': 'protocol104.network.connectedAp.wizard.wFactorHelper',
+  'K-FACTOR': 'protocol104.network.connectedAp.wizard.kFactorHelper',
+  'TIMEOUT-0': 'protocol104.network.connectedAp.wizard.timeout0Helper',
+  'TIMEOUT-1': 'protocol104.network.connectedAp.wizard.timeout1Helper',
+  'TIMEOUT-2': 'protocol104.network.connectedAp.wizard.timeout2Helper',
+  'TIMEOUT-3': 'protocol104.network.connectedAp.wizard.timeout3Helper',
+};
```
ca-d, line 59: Do you know why a few of your files are sometimes missing the newline at the end of the file?

src/translations/en.ts
```diff
@@ -356,6 +356,102 @@ export const en = {
...
+      },
+      logicLink: {
+        wizard: {
+          title: {
+            edit: 'Edit Logic Link',
+            add: 'Add Logic Link',
+          },
+          logicLinkNumberLabel: 'Logic Link number',
+          addedLogicLink:
+            'Added Logic Link group from SubNetwork[name="{{ subNetworkName }}"] > ConnectedAP[apName="{{ apName }}"][iedName="{{ iedName }}"]',
```
ca-d, line 423: Do we want to assume our users understand CSS Selector syntax? I understand this is nice and compact, but I'd prefer if we came up with a more natural formulation instead even if it's longer.

src/wizards/connectedap.ts
```diff
@@ -85,6 +85,34 @@ function existConnectedAp(accesspoint: Element): boolean {
   return (connAp && isPublic(connAp)) ?? false;
 }
 
+/**
+ * Creates a TypeRestriction checkbox for a given ConnectedAP wizard.
+ * @param element - The ConnectedAP of the wizard.
+ * @returns The checkbox within a formfield.
+ */
+export function createTypeRestrictionCheckbox(element: Element): TemplateResult {
```
ca-d, line 93: Do we really need to export functions like this from within our wizards library or would it be an option to duplicate this code where it is needed?

ca-d (commented):
I didn't manage to do a proper review of this branch so far, but it still looks to me like it could benefit from one. I mostly looked at the files outside the `protocol104` directory.

I also didn't manage to provide all the German translations yet, since I don't fully understand the English translations myself. I hope @JakobVogelsang will be able to help with this.

src/foundation.ts
```diff
@@ -537,6 +537,16 @@ export function getInstanceAttribute(element: Element): string | undefined {
   return inst ? inst : undefined;
 }
 
+/**
+ * Extract the 'type' attribute from the given XML element.
+ * @param element - The element to extract instance from.
+ * @returns the value, or undefined if there is no instance.
+ */
+export function getTypeAttribute(element: Element): string | undefined {
```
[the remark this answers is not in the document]
dlabordus: There was already a getNameAttribute and getDescriptionAttribute there, so that was the reason to put it there.
And I like to have such function at least in the IED Editor and 104 Plugin, because there type is retrieved a lot.
So I can move them to the foundation of these editors.

src/editors/protocol104/base-container.ts
dlabordus: Yes, I like these kind of classes. And I don't like the name LitElementWithDocProperty, because you can't put anything else in there anymore. I think in 1 place I also added nsdoc there, because it was also in every container.

I started with this class, because of the need to pass every container the doc property and didn't like to copy it to every container.

src/editors/protocol104/foundation/p-types.ts
dlabordus: @Flurb, was the formatter not yet configured when creating this class?
@ca-d, we only recently configured the formatter, so this may still be something from before that.

src/translations/en.ts
dlabordus: @Flurb, is this something that accidentally was added to this message, because it could be something like `'Added Logic Link group to SubNetwork (name="{{ subNetworkName }}") and ConnectedAP (AccessPoint Name="{{ apName }}", IED Name="{{ iedName }}")'

src/wizards/connectedap.ts
dlabordus: I don't know why this is done here, but I personally see that I kind of need it for Unit Testing, but I'm not sure it's needed here.

@Flurb, was it needed here?

src/foundation.ts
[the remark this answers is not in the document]
dlabordus: Moved the code to the foundation.ts of the 104 Container Directory.

src/editors/protocol104/foundation/p-types.ts
dlabordus: Fixed them. And it looks like the formatter not being configured locally.

src/translations/en.ts
dlabordus: Changed label.

src/wizards/connectedap.ts
dlabordus: Done

ca-d (approved):
LGTM

PR #843, opened 2022-07-01
fix(Pluggin/Hosting): allocate menu plugin content next to its list item

### LINKED ISSUE #842: Not possible to use own dialogs in menu type plugins
**Describe the bug**
Working with dialogs in menu type actions is not possible at the moment. One have to use the `wizarding` events working with `wizard-dialog` web component. If I want to work with own dialog like the `mwc-dialog` I cannot close the dialog.

**To Reproduce**
1. Create new menu type plugin
2. Add new dialog like so
    ```ts
      ...
      @query('mwc-dialog') dialog!: Dialog;
      ....
      async run (): Promise<void> {
          this.dialog.open = true;
      }
      ...
      render(): TemplateResult {
          return html`<mwc-dialog></mwc-dialog>`
      }
      
    ```
4. Click on the menu type plugin. You will see that the dialog is opened. But you cannot close it other then pressing ESC.

**Expected behavior**
The dialog should close in secondary and primary action

**Additional context**
The reason why this is happening is that in `Hosting` the content of the menu type plugin is rendered into the `mwc-list-item`. Which means clicking on the dialog also means clicking on the `mwc-list-item` which triggers the `async run()` again :)

JakobVogelsang:
Close #842 

Regression tests are missing here, sorry. In `open-scd` we do not have a regression test that would allow testing this specific situation. We should take it into account in `open-scd-core`.

The solution was to make menu plugin content the next sibling of it, `list-item` not its child. Doing so, a click on the content is not interpreted as a click in the list-item that triggers the run method.

### DIFF
--- a/src/Plugging.ts
+++ b/src/Plugging.ts
@@ -1,4 +1,5 @@
 import { html, query, TemplateResult } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map';
 import { translate } from 'lit-translate';
 
 import '@material/mwc-button';
@@ -291,6 +292,12 @@ export function Plugging<TBase extends new (...args: any[]) => EditingElement>(
             .docId=${this.docId}
             .pluginId=${plugin.src}
             .nsdoc=${this.nsdoc}
+            class="${classMap({
+              plugin: true,
+              menu: plugin.kind === 'menu',
+              validator: plugin.kind === 'validator',
+              editor: plugin.kind === 'editor',
+            })}"
           ></${tag}>`,
       };
     }
--- a/src/Hosting.ts
+++ b/src/Hosting.ts
@@ -73,7 +73,7 @@ export function Hosting<
               newPendingStateEvent(
                 (<MenuPlugin>(
                   (<unknown>(
-                    (<List>ae.target).items[ae.detail.index].lastElementChild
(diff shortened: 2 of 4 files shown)


src/Plugging.ts
```diff
@@ -291,6 +292,12 @@ export function Plugging<TBase extends new (...args: any[]) => EditingElement>(
             .docId=${this.docId}
             .pluginId=${plugin.src}
             .nsdoc=${this.nsdoc}
+            class="${classMap({
+              plugin: true,
+              menu: plugin.kind === 'menu',
+              validator: plugin.kind === 'validator',
+              editor: plugin.kind === 'editor',
```
ca-d, line 299: Could we use `[plugin.kind]: true` instead of these three lines?

ca-d (approved):
LGTM

src/Plugging.ts
JakobVogelsang: I do not get how this is solving the same problem. Could you be a bit more detailed?

JakobVogelsang:
I think we have to still do something here @ca-d. We move the web-component of the menu type plugin out of the list item, but they are still being in the list and therefore the `@action` triggers when clicking in a list item within the plugin.

Issue #711, opened 2022-04-27
Add SampledValueControl bar


JakobVogelsang:
As a user of OpenSCD I want to see all defined SampledValueControl in the project

** Requirement **
- group SampledValueControl per IED
- allow filter for SampledValueControl as well as its parents

** Mockup **
As a suggestion. Use the full width so long the other panes are not implemented.
![SampledValueControlSidebar](https://user-images.githubusercontent.com/66802940/166685752-7dbdea17-e373-4c27-89aa-cd42b5f62d29.png)

Issue #710, opened 2022-04-27
Add GSEControl select bar


JakobVogelsang:
As a user of OpenSCD I want to see all defined GSEControl element in the project.

** Requirement **
- group GSEControl per IED
- allow filtering for GSEControl and its parents

** Mockup **
As a suggestion. Use the full width so long the other panes are not implemented

Issue #709, opened 2022-04-27
Add ReportControl select bar


JakobVogelsang:
As a user of OpenSCD I want to see the number of all defined `ReportControl` in the project. 

** Requirements **
- group ReportControl per IED
- allow filtering for ReportControl elements

** Mockup **
As a suggestion. So long the other panes are not implemented use the full width of the screen.

PR #844, opened 2022-07-03
feat(editors/publisher): filter for control blocks and DataSets

JakobVogelsang:
This PR is covering issues #709 #710 and #711 in showing a filterable list of `ReportControl`, `GSEControl` and `SampledValueControl` elements. I also took into account the discussion #561 and added defined `DataSet` elements as well. This will allow users to follow a workflow where `DataSet` element are created independently of control blocks and a re combined with those in a lager engineering step or even online.

### DIFF
--- a/public/js/plugins.js
+++ b/public/js/plugins.js
@@ -56,6 +56,13 @@ export const officialPlugins = [
     kind: 'editor',
   },
   {
+    name: 'Publisher',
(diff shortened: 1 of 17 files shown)


ca-d:
Should this already open some wizard when clicking the list items or is this just for displaying the lists?

ca-d (approved):
LGTM

PR #852, opened 2022-07-07
fix(editors/substation): update on action

### LINKED ISSUE #850: Creating Function in Substation section fails
**Describe the bug**
It is not possibe to create a function at any level. 

**To Reproduce**
1. on the voltage level 
2. Click right and select Function
3. See error

**Expected behavior**
A function similar to a equipment.

**Screenshots**
grafik

JakobVogelsang:
Close #850 

I did not add regression tests as I think they cannot be done without visual regression tests. Btw @ca-d the snapshot type regression tests you did for the `Communication` editor fixing this issue are incorrect. I did not have a close enough look at those during review. You can see by looking at the number of children. This never differs, independent whether one is deleted or added.

### DIFF
--- a/src/editors/substation/bay-editor.ts
+++ b/src/editors/substation/bay-editor.ts
@@ -41,6 +41,9 @@ function childTags(element: Element | null | undefined): SCLTag[] {
 /** [[`SubstationEditor`]] subeditor for a `Bay` element. */
 @customElement('bay-editor')
 export class BayEditor extends LitElement {
+  /** The document being edited as provided to editor by [[`Zeroline`]]. */
+  @property({ attribute: false })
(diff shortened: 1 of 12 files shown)


ca-d:
Do you want me to fix the broken snapshot tests here? Or do we want a separate bug issue for that?

JakobVogelsang:
@ca-d I do not see we can cover that with the snapshots. At this stage, I think we have to either add a visual regression test or take the PR in as is. However, maybe I am missing how to properly implement these test with snapshots :)

ca-d (approved):
I'm in favor of taking this over without extra regression tests as is.

LGTM

Issue #655, opened 2022-04-01
Subscriptions filtering by device and message name


danyill:
After some very excellent work on subscriptions, I thought I'd have a look.

I expect the filtered list in the subscriptions to allow me to scope the IED basic properties (using the identity function) as well as the GOOSE message names (I looked at the GOOSE subscriptions). (I speculate we could concatenate the identity functions).

I would like to be able to scope to the IED name, manufacturer or type.

I tested the GOOSE subscriptions using https://openscd.github.io/ just now.

On the publishing side when I start typing an IED name it filters the IEDs.

![image](https://user-images.githubusercontent.com/674783/161213743-98c1db0f-39fb-463c-8825-239f5460c66f.png)

But this filtering prevents the GOOSE messages from being displayed.

![image](https://user-images.githubusercontent.com/674783/161213835-33a6b1d1-e861-4fc9-b751-9f61fb136724.png)

JakobVogelsang:
See the issue before yours #654 :)

Flurb:
Bit more background, the IED names are `noninteractive` in the filtered-list, which means they cannot be interacted with (duh!). filtered-lists as we use cannot handle noninteractive items in their filtering, so something needs to be done right here.

danyill:
Thanks for the explanations.
It may be the same issue, it is not clear to me.

I was expecting to be able to filter by both the GOOSE message name and the IED characteristics. As it stands now I can filter by the GOOSE name. When I add to this a filter by the IED name. I lose the GOOSE message name.

I don't mind that the other IED objects remain in the filtered list as this will be resolved in #653 but that may not get to the heart of this.

Let me try to be more clear:

* I want to filter all D60 devices with a goose name beginning with `gcb`.
* I start by writing `gcb` and realise my list is still quite long for the task I set out to do.
   ![image](https://user-images.githubusercontent.com/674783/161311041-31a1f0cd-e063-43b3-9423-9b26d7d5e75e.png)
* I stop and think and realise that actually my task only involves D60 devices (whatever they are). I add D60 to my filter `gcb d60` expecting now to have all goose messages beginning with `gcb`, `d60` or the IED "characteristics" which have `d60` in them. 
* Instead I find I have filtered out the children of the object of interest.
    ![image](https://user-images.githubusercontent.com/674783/161311585-c866e809-6871-4a3a-8021-883dfcc61b73.png)
* Is this a good use case, I think so...

[...]

Issue #560, opened 2022-02-24
Create a search bar in the IED editor


randy993:
As a engineer 
I want to search for element types: LN, IED, LD, DO, DA in the IED editor
So I can quickly find e.g. LN of the type PTOC etc. 

This topic has been discussed a few times already, but it would be usefull to have a search function in the IED editor. This search function could be somewhat like the search function in GitHub, which enables the user to create a search query with the help of a filter bar.

![image](https://user-images.githubusercontent.com/47896624/155551686-8053f101-2cb7-4ab9-891a-28ba9e5dabca.png)

I think it would be nice that this bar 'sticks' to the top of the IED editor, so that the bar will always be visible. Furthermore, I think we should limit this story to searching only inside the selected IED and not implementing the fancy filter button yet.. Which we can possibly extend this later after we have used this functionality and gained some experience.

For later stories:
- pathname, values
- search on textfields related to NSDdoc fields
- Autocomplete?
- Select IED's

**First version**

Syntax:
LD:`<inst>`
LN: `<lnClass>`
DO/DA: `<name>`

Result:
Only show the related fields, is case of "LN: LLN0" shows: 
![image](https://user-images.githubusercontent.com/7427358/157454495-cd189183-36e2-4e52-8878-08d10795dab0.png)
The search function should search in all attributes of the element, and the values.

Exact visual representation has to be discusses later.

This issues relates to:
https://github.com/openscd/open-scd/issues/554

Sander3003:
@randy993 can you do a suggestion for search strings?

It would be nice if we can search for: LN, IED, LD, DO, DA, pathname

This issues relates to:
https://github.com/openscd/open-scd/issues/554

randy993:
I think the above mentioned would be sufficient for this version. Regarding the NSD textfields, in my opinion this is a must. As mentioned in the previous post, I think we should implement functions as searching within IEDs/autocomplete in a later stage.

Sander3003:
The list should show the children and parents and exclude the IED/LD/DO/DA that do not meet the search criteria (or parents or children).

PR #858, opened 2022-07-08
feat(editor/publisher): add IED filter button

JakobVogelsang:
Pure prototype for a nerdy Friday :)

### DIFF
--- a/src/editors/publisher/data-set-editor.ts
+++ b/src/editors/publisher/data-set-editor.ts
@@ -6,7 +6,7 @@ import {
   property,
   TemplateResult,
 } from 'lit-element';
-import { compareNames, identity } from '../../foundation.js';
+import { compareNames, identity, isPublic } from '../../foundation.js';
 
 @customElement('data-set-editor')
 export class DataSetEditor extends LitElement {
@@ -18,6 +18,7 @@ export class DataSetEditor extends LitElement {
     return html`<filtered-list
       >${Array.from(this.doc.querySelectorAll('IED'))
         .sort(compareNames)
+        .filter(isPublic)
         .flatMap(ied => {
           const ieditem = html`<mwc-list-item
               class="listitem header"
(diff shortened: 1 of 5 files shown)


JakobVogelsang:
@danyill what you think of the filter button in the GSEControl list next to the search text fiels?

danyill:
I quite like it, as a prototype I guess we'd need to do a little work on the UI. This could be a powerful way to interact with large files.

Part of me says, though that if I'm filtering e.g. GSEControl and the IED name is in the secondary text then a single flat filter should be able to pick that up too and that IEDs with non-matching GSEControls simply shouldn't appear.

It may be a UI crime however to provide a parent/child and then "munge" them together within the filter bar and sooner or later I can see there will be too many fields/options and the user experience will suffer.

But maybe with the parent being the IED and the child being say the GSEControl name, then it might be OK to filter on IED name and GSE control name within a single filter bar?

:grey_question: 

This is somewhat similar to: https://github.com/openscd/open-scd/issues/655 and #560 

However, having a filter is probably a more "future-proof solution". I know it's only a prototype but I'd consider putting the filter icon on the right and having a pop-out for the IEDs name - putting at the top and displacing the existing list is a little confusing.

JakobVogelsang:
@danyill I have made some more CSS. I think this is coming close to the recommendations. I also made `noninteractive` elements in the filtered list filterable. Now you have both the solutions combined. I would love to get your opinion.

danyill:
Thank you for looking at this feedback. I like it! With an "IED name" heading in the filter I think we'd almost be there but there are some performance issues in large files:

* The fact that it is reactive on the list as each item selection is made, makes the user interface "sticky" with a large scd file.
(I contemplate whether we need service workers indexing things in the background if want this kind of reactive interface).
*  I wonder if the filter should default to enabling everything. Better performance would be to enable nothing and mean that the user didn't first have to "deselect all".
* Hitting "deselect all" or "select all" in the filter is _very_ sticky in a large scd file. I wonder if it is running multiple updates or whether this is just the penalty of a large file.
* I think it would be OK in the filter that the reactive update didn't occur until the panel was X-ed. That would dramatically improve performance.
* I wonder if the filter should be capped to 100% of vh and scrollable without scrolling the "outer pane", so the filter only occupies one screen height.

PR #872, opened 2022-07-13
fix(editor/ied): fixed styling how DA(I) values are displayed.

pascalwilbrink:
This breaks the text onto a new line if the text is too long. if there are multiple words, it tries to break it into logic pieces.

### DIFF
--- a/src/editors/ied/da-container.ts
+++ b/src/editors/ied/da-container.ts
@@ -59,7 +59,7 @@ export class DAContainer extends Container {
    */
   private getValue(): TemplateResult {
     if (this.instanceElement) {
-      return html`<b style="word-break: break-word; white-space: pre-wrap;"
+      return html`<b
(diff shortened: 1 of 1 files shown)


src/editors/ied/da-container.ts
```diff
@@ -59,7 +59,7 @@ export class DAContainer extends Container {
    */
   private getValue(): TemplateResult {
     if (this.instanceElement) {
-      return html`<b
+      return html`<b style="word-break: break-word; white-space: pre-wrap;"
```
dlabordus, line 62: This will only work if the value is from the DAI, but if there is no DAI the value comes from the DA, if there (line 67).
Maybe it's possible to do something at line 169 with the H4 element.

And also we prefer to add the styling to the style variable (line 211). But that hasn't been done everywhere in this file I see.

dlabordus (changes requested):

dlabordus (approved):

PR #874, opened 2022-07-14
fix(editor/104): Small improvements and fix.

dlabordus:
Small changes to the Element Queries and one fix that LN can also be directly below a AccessPoint.


JakobVogelsang (approved):
The changes do make sense. However, been honest, I cannot double-check the behavior. The speed on my M1 mac still annoyingly slow, but it is not crashing now.

PR #875, opened 2022-07-14
fix: Adding Subnetwork to a configuration without a Communication Ele…

### LINKED ISSUE #847: Adding SubNetwork element in COMMUNICATION does not work right away
**Describe the bug**
In the `Communication` plugin editor the fab button  `ADD SUBNETWORK` does not work right away, if a file does not contain a `Communication` element. You have to click a second time for a `SubNetwork` element to be added.

**To Reproduce**
1. Load a file without `Communication` element
2. Navigate to COMMUNICATION editor type plugin
3.Click on `ADD SUBNETWORK` button
4. Nothing happens

**Expected behavior**
The `SubNetwork` element should be added on the first click.

pascalwilbrink:
closes #847
…ment failed for the first time

When creating a `Communication` Element in Communication.ts, the created Element is also returned to use it as a parent for the wizard

The `Should create a Communication Element` integration test is extended to also check that the Wizard is closed properly

### DIFF
--- a/test/integration/editors/communication/Communication.test.ts
+++ b/test/integration/editors/communication/Communication.test.ts
@@ -1,18 +1,16 @@
 import { html, fixture, expect } from '@open-wc/testing';
 
-import '../../../mock-wizard.js';
-import { MockWizard } from '../../../mock-wizard.js';
+import '../../../mock-wizard-editor.js';
+import { MockWizardEditor } from '../../../mock-wizard-editor.js';
 
-import Communication from '../../../../src/editors/Communication.js';
-import { Editing } from '../../../../src/Editing.js';
-import { Wizarding } from '../../../../src/Wizarding.js';
+import Communication from '../../../../src/editors/Communication.js'; 
 import { Dialog } from '@material/mwc-dialog';
 import { WizardTextField } from '../../../../src/wizard-textfield.js';
 
 describe('Communication Plugin', () => {
   customElements.define(
     'communication-plugin',
-    Wizarding(Editing(Communication))
+    Communication
   );
   let element: Communication;
   beforeEach(async () => {
@@ -45,18 +43,23 @@ describe('Communication Plugin', () => {
 
   describe('with a doc loaded missing a communication section', () => {
     let doc: XMLDocument;
-    let parent: MockWizard;
+    let parent: MockWizardEditor;
     let fab: HTMLElement;
+    let element: Communication;
 
     beforeEach(async () => {
       doc = await fetch('/test/testfiles/missingCommunication.scd')
         .then(response => response.text())
         .then(str => new DOMParser().parseFromString(str, 'application/xml'));
-      parent = <MockWizard>(
+      
+      element = await fixture(
+        html`<communication-plugin .doc="${doc}"></communication-plugin>`
+      );
+
+      parent = <MockWizardEditor>(
         await fixture(
-          html`<mock-wizard
-            ><communication-plugin .doc=${doc}></communication-plugin
-          ></mock-wizard>`
+          html`<mock-wizard-editor
+            >${element}/mock-wizard-editor>`
         )
       );
       await element.updateComplete;
@@ -78,25 +81,28 @@ describe('Communication Plugin', () => {
   
     it('Should create a Communication Element', async () => {
           expect(parent.wizardUI.dialogs.length).to.equal(0);
-          fab.click();
-          await parent.updateComplete;
-          expect(parent.wizardUI.dialogs.length).to.equal(1);
+          expect(element.doc.querySelector('Communication')).is.null;
 
-          const dialog: Dialog = parent.wizardUI.dialogs.item(0);
+          await fab.click();
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
+          await parent.updateComplete;
+
+          expect(parent.wizardUI.dialogs.length).to.equal(1);
+          
+          const dialog: Dialog = parent.wizardUI.dialog!;
+          expect(dialog).to.not.be.undefined;
 
           const nameInput: WizardTextField = dialog.querySelector<WizardTextField>('wizard-textfield[label="name"]')!;
-
           nameInput.value = 'Test';
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
 
           const saveButton: HTMLElement = dialog.querySelector('mwc-button[slot="primaryAction"]')!;
+          await saveButton.click();
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
 
-          await parent.updateComplete;
-
-          saveButton.click();
-
-          await parent.updateComplete;
-
-          expect(parent.wizardUI.dialogs.length).to.equal(0);
+          expect(parent.wizardUI.dialog).not.to.exist;
+          expect(element.doc.querySelector('Communication')).not.is.null;
+          expect(element.doc.querySelector('Communication > SubNetwork[name="Test"]')).to.exist;
     });
   });
 });
--- a/test/integration/editors/communication/Communication.test.ts
+++ b/test/integration/editors/communication/Communication.test.ts
@@ -98,7 +98,6 @@ describe('Communication Plugin', () => {
           await saveButton.click();
           await new Promise(resolve => setTimeout(resolve, 100)); // await animation
 
-          expect(parent.wizardUI.dialog).not.to.exist;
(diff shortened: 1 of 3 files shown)


test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +75,28 @@ describe('Communication Plugin', () => {
       await parent.updateComplete;
       expect(parent.wizardUI.dialogs.length).to.equal(1);
     });
+  
+    it('Should create a Communication Element', async () => {
+          expect(parent.wizardUI.dialogs.length).to.equal(0);
+          fab.click();
+          await parent.updateComplete;
+          expect(parent.wizardUI.dialogs.length).to.equal(1);
```
JakobVogelsang, line 83: Has been checked in the test before. I would skip it :)

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +75,28 @@ describe('Communication Plugin', () => {
...
+
+          const dialog: Dialog = parent.wizardUI.dialogs.item(0);
```
JakobVogelsang, line 85: There is a property called `parent.wizardUI.dialog` that does the same

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +75,28 @@ describe('Communication Plugin', () => {
...
+
+          const saveButton: HTMLElement = dialog.querySelector('mwc-button[slot="primaryAction"]')!;
+
+          await parent.updateComplete;
+
+          saveButton.click();
+
+          await parent.updateComplete;
+
+          expect(parent.wizardUI.dialogs.length).to.equal(0);
```
JakobVogelsang, line 99: The important check is missing. What you want to see is that the `Communication` section including the `SubNetwork[name="Test"]` has been created. So something like this 

`expect(doc.querySelectot('Communication > SubNetwork[name="Test"]').to.exist`

JakobVogelsang, line 99: This test is check that the wizard closed is not needed here, I would suggest.

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +75,28 @@ describe('Communication Plugin', () => {
...
+          expect(parent.wizardUI.dialogs.length).to.equal(1);
+
+          const dialog: Dialog = parent.wizardUI.dialogs.item(0);
+
+          const nameInput: WizardTextField = dialog.querySelector<WizardTextField>('wizard-textfield[label="name"]')!;
+
+          nameInput.value = 'Test';
+
+          const saveButton: HTMLElement = dialog.querySelector('mwc-button[slot="primaryAction"]')!;
+
```
JakobVogelsang, line 92: I would use empty lines with care. I usually use them to tell the reader of the code that something new is happening. If you have an empty line after each line that does not make the test any more readable than leaving the empty line out completely is my feeling.

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +75,28 @@ describe('Communication Plugin', () => {
...
+  
+    it('Should create a Communication Element', async () => {
+          expect(parent.wizardUI.dialogs.length).to.equal(0);
+          fab.click();
+          await parent.updateComplete;
+          expect(parent.wizardUI.dialogs.length).to.equal(1);
+
+          const dialog: Dialog = parent.wizardUI.dialogs.item(0);
+
+          const nameInput: WizardTextField = dialog.querySelector<WizardTextField>('wizard-textfield[label="name"]')!;
```
JakobVogelsang, line 87: You formatter in not working. This line should break otherwise :)

JakobVogelsang (changes requested):

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: Fixed

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: Fixed

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: I deleted some empty lines

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +78,28 @@ describe('Communication Plugin', () => {
       await parent.updateComplete;
       expect(parent.wizardUI.dialogs.length).to.equal(1);
     });
+  
+    it('Should create a Communication Element', async () => {
+          expect(parent.wizardUI.dialogs.length).to.equal(0);
+          expect(element.doc.querySelector('Communication')).is.null;
+
+          await fab.click();
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
JakobVogelsang, line 87: I don't think we need those. They are there for debugging. Also try to get rid of any `element.updateComplete` and use them only where you need it :)

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +78,28 @@ describe('Communication Plugin', () => {
       await parent.updateComplete;
       expect(parent.wizardUI.dialogs.length).to.equal(1);
     });
+  
+    it('Should create a Communication Element', async () => {
+          expect(parent.wizardUI.dialogs.length).to.equal(0);
```
JakobVogelsang, line 83: This is tested with the test before. Try to reduce the line in the test to what you want to show with the test. It makes the test more readable :)

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +78,28 @@ describe('Communication Plugin', () => {
...
+          expect(element.doc.querySelector('Communication')).is.null;
+
+          await fab.click();
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
+          await parent.updateComplete;
+          
+          const dialog: Dialog = parent.wizardUI.dialog!;
```
JakobVogelsang, line 90: This is not the purpose of the test. The dialog is just a thing that you need to use here. We wanne see that the `Communication` was not in the doc and after the both `Communication` and a specific `SubNetwork` is in the doc :)

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +78,28 @@ describe('Communication Plugin', () => {
...
+          expect(dialog).to.not.be.undefined;
+
+          const nameInput: WizardTextField = dialog.querySelector<WizardTextField>('wizard-textfield[label="name"]')!;
+          nameInput.value = 'Test';
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
JakobVogelsang, line 95: Debug leftover?

test/integration/editors/communication/Communication.test.ts
```diff
@@ -73,5 +78,28 @@ describe('Communication Plugin', () => {
...
+
+          const saveButton: HTMLElement = dialog.querySelector('mwc-button[slot="primaryAction"]')!;
+          await saveButton.click();
+          await new Promise(resolve => setTimeout(resolve, 100)); // await animation
```
JakobVogelsang, line 99: Degug leftover?

JakobVogelsang (changes requested):
Some line in the test are not needed and make the test less readable. I would suggest refactoring a bit more.

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: Fixed

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: This one is needed for the dialog to open, checked it in the test

test/integration/editors/communication/Communication.test.ts
pascalwilbrink: This one is needed because the Subnetwork Element is created asynchronously (checked it in the test)

JakobVogelsang: I am not so sure. Creating such an element is really not time-consuming. I think it will work without any awaits here :)

pascalwilbrink: when i removed it, the test failed. because an event is risen, it needs to wait

JakobVogelsang (approved):
LGTM

PR #876, opened 2022-07-14
feat(filtered-list): filter on list item value

JakobVogelsang:
After two prototypes and some discussion, I decided to change the filter behavior of filtered-list. What has changed is that not only the list item's primary and secondary span is taken into account, but also its value attribute. This has in my opinion the advantage to be simple in design and very flexible for the user to use. As value is not visible, everybody can decide himself what the value can be and therefor also what the user can filter for. 

Best to test is the publisher plugin or both the subscriber plugins :)

### DIFF
--- a/src/editors/publisher/data-set-editor.ts
+++ b/src/editors/publisher/data-set-editor.ts
@@ -61,5 +61,9 @@ export class DataSetEditor extends LitElement {
     .listitem.header {
       font-weight: 500;
     }
+
+    mwc-list-item.hidden[noninteractive] + li[divider] {
+      display: none;
+    }
   `;
 }
--- a/src/editors/publisher/data-set-editor.ts
+++ b/src/editors/publisher/data-set-editor.ts
@@ -28,7 +28,7 @@ export class DataSetEditor extends LitElement {
                   const id = identity(element) as string;
                   return typeof id === 'string' ? id : '';
                 })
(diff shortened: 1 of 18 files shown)


JakobVogelsang:
@danyill I think this is better than having a heavyweight extra filtered-list for the filtered-list. What do you think?

JakobVogelsang:
@danyill I found a nice CSS solution for the multiple `li` elements.

src/editors/publisher/data-set-editor.ts
```diff
@@ -23,6 +23,12 @@ export class DataSetEditor extends LitElement {
               class="listitem header"
               noninteractive
               graphic="icon"
+              value="${Array.from(ied.querySelectorAll('DataSet'))
+                .map(element => {
+                  const id = identity(element) as string;
+                  return typeof id === 'string' ? id : '';
+                })
+                .join('')}"
```
dlabordus, line 31: Is it not better to join on ' ' (space) or '\n' (like also done in the filtered-list?
You now have the change that 2 words are joined together and that may give some strange results.

Also the other joins.

JakobVogelsang: True that! I will fix it using a space, as we use it in `filtered-list` as well. Thx

dlabordus (approved):

PR #877, opened 2022-07-17
Fix filtering of control blocks in Cleanup editor, closes #873

### LINKED ISSUE #873: Filtering in Cleanup Editor for Control Blocks broken
**Describe the bug**

**To Reproduce**

1. Load an scd file with controls blocks which refer to missing or invalid datasets
2. Attempt to filter as shown in (1) below.
3. Observe near (2) below that no filtering occurs
4. Observe that the Unreferenced Datasets widget works at (3) (with correct filtering) and regain a little strength.
5. Nonetheless burst into tears :cry:, this used to work...

Perhaps something to do with the reactive change? But updates after button clicks appear to work fine.
I shall investigate in the fullness of time.

**Expected behavior**

The filter filters based on the text.

**Screenshots**

image

danyill:
I shall be delighted to receive a review, closes #873

### DIFF
--- a/src/editors/cleanup/control-blocks-container.ts
+++ b/src/editors/cleanup/control-blocks-container.ts
@@ -20,9 +20,11 @@ import '@material/mwc-checkbox';
 
 import { Button } from '@material/mwc-button';
 import { Checkbox } from '@material/mwc-checkbox';
-import { List, MWCListIndex } from '@material/mwc-list';
+import { MWCListIndex } from '@material/mwc-list';
 
-import '../../filtered-list.js';
+import { IconButtonToggle } from '@material/mwc-icon-button-toggle';
+
+import { FilteredList } from '../../filtered-list.js';
 
 import {
   Delete,
@@ -92,7 +94,7 @@ export class CleanupControlBlocks extends LitElement {
   cleanButton!: Button;
 
   @query('.cleanupList')
-  cleanupList: List | undefined;
+  cleanupList: FilteredList | undefined;
 
   @queryAll('mwc-check-list-item.cleanupListItem')
   cleanupListItems: NodeList | undefined;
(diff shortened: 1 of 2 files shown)


danyill:
> This does make sense, I think, as we never know what other rules might be applied to the list that might prevent the filtering of the list items. And we want them to be always filter. ca-d do I miss something here?

An interesting solution but I think you miss something! If I understand what line 150 is. Would it be:

```css
    ::slotted(.hidden) {
      display: none !important;
    }
```

Forgiving the crime of using `!important`, let's try this test :wink:

1.     Start with `test/testfiles/cleanup.scd`
2.     Turn on the R filter
3.     Enter `gcb` in the search window (2 results)
4.     Turn off L,R,G,S filter (0 results, greyed out)
5.     Turn on S filter (lavender)
6.     Notice `MSVCB01_A` appears when it should not
7.     You could turn on any filter button and get everything in there shown

The problem is that we cannot overload the use of the hidden class without re-applying both the filtering from the filtered list _and_ the filtering from the buttons to be sure it has occurred correctly without an order of operations problems (I think).

JakobVogelsang:
I think I have found an elegant solution. To be able to work on it, I created an own branch. Test must be fixed, but the code seems to work. Btw. the first commit in the PR is to make the linter happy. I have hard time to review your web components as the coloring is ruined so long the linter is not happy.

danyill:
> Btw. the first commit in the PR is to make the linter happy. I have hard time to review your web components as the coloring is ruined so long the linter is not happy.

Are you using the repository's `.editorconfig`?

My console suggests that Prettier is using `.editorconfig`, so I'm not sure why we would be receiving linting inconsistencies.

```
["INFO" - 9:13:22 pm] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
```

danyill:
The solution of #910 seems better so we should close this after that is merged.

PR #888, opened 2022-07-19
fix(mergeWizard): insert element at valid position

ca-d:
The merge wizard was implemented before our "choose a valid position
automatically" feature was implemented for insert and move actions in
Editing (`reference: undefined`).

Therefore, all insert actions fired by the merge wizard insert the
element at as the parent's last child, resulting in schema invalid files
in many cases (e.g. insert a Substation section in a file with an IED or
DataTypeTemplates). This fixes that issue, making our Merge and
UpdateSubstation plugins schema compliant.

### DIFF
--- a/src/wizards.ts
+++ b/src/wizards.ts
@@ -73,7 +73,7 @@ function mergeWizardAction(
       for (const diff of selectedChildDiffs)
         if (!diff.ours)
           actions.push({
-            new: { parent, element: diff.theirs, reference: null },
+            new: { parent, element: diff.theirs },
           });
         else if (!diff.theirs)
           actions.push({


ca-d:
@JakobVogelsang this itty bitty bug fix is still looking for a bug issue and a regression test. If you find the time, I'd much appreciate you adding them. Otherwise I'll do it myself next week at the latest.

JakobVogelsang:
Closes #890

JakobVogelsang:
This functionality is not tested at all - schame on me :(. So adding regression tests means adding fresh tests and that is a bigger issue. I would therefore add the fix as it is a pretty obvious one and create another issue to add tests for this functionality. @ca-d what you think?

JakobVogelsang:
See #892

JakobVogelsang (approved):
LGTM

PR #889, opened 2022-07-19
fix(editing): don't validate after no-op action

ca-d:
After committing an empty complex action (e.g. clicking "merge" in a
merge wizard without any additions or deletions selected) the `Editing`
mixin needlessly dispatches a ValidateEvent causing lengthy validation
of an already validated document. This fixes that issue.

### DIFF
--- a/src/Editing.ts
+++ b/src/Editing.ts
@@ -419,7 +419,7 @@ export function Editing<TBase extends LitElementConstructor>(Base: TBase) {
             action: event.detail.action,
           })
         );
-      }
+      } else return;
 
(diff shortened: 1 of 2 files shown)


ca-d:
@JakobVogelsang this itty bitty buggy fixy also still needs an issue and a regression test, I'll furnish those next week if nobody else finds the time to do it. Until that time, this shall remain a draft pull request.

JakobVogelsang:
Closes #891

JakobVogelsang (approved):

PR #899, opened 2022-07-20
fix(editing): set false attribute values on update

ca-d:
Since we switched from Replace to Update, we now no longer set falsey attribute values (like the empty string `''` or `0`). This pull request fixes that issue.


ca-d:
@JakobVogelsang this itty bitty buggy fixy again does not come with a regression test or a bug issue and is therefore still a draft. If nobody else finds the time to add those this week, I'll add them next week and turn the draft into a pull request then.

ca-d:
> I'll add them next week

Oh - my youthfully naive optimism from July is coming back to haunt me... :sweat_smile:

JakobVogelsang (approved):
LGTM

PR #900, opened 2022-07-20
Accept multiple IEDs from same file in import IEDs. Closes #897

### LINKED ISSUE #897: Importing an IED only works once
**Describe the bug**

@mtilbrook has been doing some testing for me.
It appears that from a file with a single IED, it can only be imported once using the "Import IEDs" menu option.

**To Reproduce**

* [XAT_PCS-221S.zip](https://github.com/openscd/open-scd/files/9145591/XAT_PCS-221S.zip)

1. Open the attached zip file, `XAT_PCS-221S.zip`.
2. Open from within the zip file, the `XAT_Prot1_Base_Configuration.scd` file.
3. Click on 'Import IEDs', select the `XAT_PCS-221S.icd`. This only has a single IED.
4. The IED is imported successfully.
5. Now in the Substation editor, click on the IED , go to Edit and modify the name so it is different from the original file.
6. Click Save (note another small bug that the IED names are not refreshed in the Substation Editor) (I will file a separate issue for this).
7. Go to Import IEDs.
8. Select the same file.
9. Nothing happens.

**Expected behavior**

I should be able to import a device from the same SCL file more than one time, particularly if it has a different name from all other devices in the project.
It should not depend on the location of where the file came from on my computer.

 - OS: Windows 10
 - Browser: Chrome
 - Version: Version 103.0.5060.114 (Official Build) (64-bit)

**Additional context**

This appears to work if the SCL file being added is from a different location on the user's computer.
It seems to only be a problem with the "last added file", if two copies are created on a computer the use can alternate between them.

danyill:


JakobVogelsang (approved):
LGTM

danyill:
Record PR response time! Gracias.

PR #901, opened 2022-07-20
fix(validate-schema): cache validator workers

ca-d:
Due to a simple oversight we are currently not actually caching schema validator workers at all, meaning we instantiate a new worker on each validation, leaking memory and wasting time. This fixes that issue.

This pull request should result in a noticeable performance increase for repeat validations and in less memory usage.


ca-d:
LGTM

PR #902, opened 2022-07-20
fix(editor/substation): Updated IED name is shown in IED container

### LINKED ISSUE #898: Editing an IED's name from the Substation Editor doesn't refresh the name
**Describe the bug**

@mtilbrook has been doing some testing for me and noticed that editing an IED's name from the Substation Editor doesn't refresh the name on https://openscd.github.io/

**To Reproduce**
1. Open an SCL file with some IEDs.
2. Go to the Substation Editor.
3. Edit the IED and change the name. Click on Save.
4. When returning to the Substation editor, the IED name has not changed.
5. Go to another edit tab and then return.
6. The IED is now correctly updated.

**Expected behavior**

After editing an IED's name, it should update after returning from the wizard to the main editor display.

 - OS: Windows 10
 - Browser: Chrome
 - Version: Version 103.0.5060.114 (Official Build) (64-bit)

danyill:
Closes #898.

If I learnt anything from the reactive updates discussion, then perhaps this is an acceptable fix :wink: 
It does seem to work, which makes me happy.

### DIFF
--- a/src/editors/substation/zeroline-pane.ts
+++ b/src/editors/substation/zeroline-pane.ts
@@ -113,7 +113,7 @@ export class ZerolinePane extends LitElement {
 
     return ieds.length
       ? html`<div id="iedcontainer">
-          ${ieds.map(ied => html`<ied-editor .element=${ied}></ied-editor>`)}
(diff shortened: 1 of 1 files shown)


ca-d (commented):
Yes this is an acceptable fix, thank you very much for writing it!

I think it might still use a regression test.

PR #903, opened 2022-07-20
feat(menu/compareied): Add new menu option to compare a IED with one from template project

### LINKED ISSUE #836: Compare IED from Actual Substation with IED from Template
As a test engineer
I want to compare SCL IED Templates with IED Implementations
So I see the differences, based on the differences, I want to judge the amount of testing required

1. Open the SCL File containing the IED Templates; This will pop-up the File Open Dialog of the browser, just like with Open Project.

2. Next step is that the user selects the IED to compare.
On the left side select an IED from the implementation and on the right side selects a IED template from the opened file.
image
https://excalidraw.com/#json=V5SUeZ88Qn8T3Viirh3f7,R8XqBOeEkqNSTQDenBIurw

3. The user will start the compare function.
With the compare function, the IED can be compared with the IED of the template.
In CoMPAS there is now a compare between versions and showing all difference in one wizard.
We like to move this functionality to OpenSCD and use that here to show the differences.

**Requirements**
What are the minimal requirements you need to have for this feature wish

**Additional information**
The compare will show the difference below the IED Element, so up to the LN(0), including the DOI/SDI/DAI.

** open question **
Is this functionality to have in OpenSCD? 
If not CoMPAS will do, if we do, move the compare function from CoMPAS to OpenSCD.

dlabordus:
closes #836


JakobVogelsang:
Hi @dlabordus I have a first look on the functionality and have some little UI remarks and one bigger functional one that I think we need to discuss:

**UI**
- first dialogs title is the same one as the second and is misleading if you ask me. It should say something like `Select IED for comparison`, `IEDs auswählen`
- title of each of the columns is plural `IEDs ...`, `IEDs im Projekt`, `IEDs aus Vorlage`
- the german translation for the button is better `Starte Vergleich`
- second dialog noninteractive header shows only one IED in my case. Say I load two IEDs with different IED names it will say `Child elements from IED(AA1E1Q01BCU)` but below that also children of the other one are organized. I do not know if this is on purpose, as they differ by the icon. With files with a lot of changes this is misleading, though, I would argue.
- I was thinking that the identity string in `()` on the second page could be in secondary span as well. They tend to be very long the dialog has an unnecessary with sometimes

dlabordus:
Hi @JakobVogelsang,

Reaction on your comments in bold:

- first dialogs title is the same one as the second and is misleading if you ask me. It should say something like Select IED for comparison, IEDs auswählen
**All 3 dialog have their own title now, that's more clear indeed. But now a extra German title is needed :-)**
- title of each of the columns is plural IEDs ..., IEDs im Projekt, IEDs aus Vorlage
**Changed it**
- the german translation for the button is better Starte Vergleich
**I think that was already there, but please check**
- Second dialog noninteractive header shows only one IED in my case. Say I load two IEDs with different IED names it will say Child elements from IED(AA1E1Q01BCU) but below that also children of the other one are organized. I do not know if this is on purpose, as they differ by the icon. With files with a lot of changes this is misleading, though, I would argue.
**I accidentally switch the two element. I also changed the names of the variables, but that stays a little but difficult. Also I made the intermediate titles bold, this way it more clear where the changes below to. Please check if it's now more clear.**
- I was thinking that the identity string in () on the second page could be in secondary span as well. They tend to be very long the dialog has an unnecessary with sometimes
**I changed it, but the identity is not bold, but these looks kind of oke.**

JakobVogelsang (approved):
LGTM. Nice to read thank you

PR #911, opened 2022-07-23
feat(editors/publisher): add read only data set element editor

JakobVogelsang:
This PR adds a read only DataSet element editor that shows referenced DataSets for `ReportControl`, `GSEControl`, `SampledValueControl` as well as `DataSet`s itself.

I have not bothered to create issues for this one. I did test a lot of CSS in the PR and therefore no Mock was present.

### DIFF
--- a/src/editors/publisher/gse-control-editor.ts
+++ b/src/editors/publisher/gse-control-editor.ts
@@ -3,25 +3,66 @@ import {
   customElement,
   html,
   LitElement,
-  property,
+  property as state,
+  query,
   TemplateResult,
 } from 'lit-element';
+import { translate } from 'lit-translate';
 
+import '@material/mwc-button';
 import '@material/mwc-list/mwc-list-item';
-import '@material/mwc-icon';
+import { Button } from '@material/mwc-button';
+import { ListItem } from '@material/mwc-list/mwc-list-item';
 
+import './data-set-element-editor.js';
 import '../../filtered-list.js';
-import { compareNames, identity } from '../../foundation.js';
+import { FilteredList } from '../../filtered-list.js';
+
 import { gooseIcon } from '../../icons/icons.js';
+import { compareNames, identity, selector } from '../../foundation.js';
+import { styles } from './foundation.js';
 
 @customElement('gse-control-editor')
 export class GseControlEditor extends LitElement {
   /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
-  @property({ attribute: false })
+  @state({ attribute: false })
   doc!: XMLDocument;
 
-  renderList(): TemplateResult {
+  @state()
+  selectedGseControl?: Element | null;
+
+  @query('.selectionlist') selectionList!: FilteredList;
+  @query('mwc-button') selectGSEControlButton!: Button;
+
+  private selectGSEControl(evt: Event): void {
+    const id = ((evt.target as FilteredList).selected as ListItem).value;
+    const gseControl = this.doc.querySelector(selector('GSEControl', id));
+
+    if (gseControl) {
+      this.selectedGseControl = gseControl.parentElement?.querySelector(
+        `DataSet[name="${gseControl.getAttribute('datSet')}"]`
+      );
+      (evt.target as FilteredList).classList.add('hidden');
+      this.selectGSEControlButton.classList.remove('hidden');
+    }
+  }
+
+  private renderElementEditorContainer(): TemplateResult {
+    if (this.selectedGseControl !== undefined)
+      return html`<div class="elementeditorcontainer">
+        <data-set-element-editor
+          .element=${this.selectedGseControl}
+        ></data-set-element-editor>
+      </div>`;
+
+    return html``;
+  }
+
+  renderSelectionList(): TemplateResult {
     return html`<filtered-list
+      activatable
+      @action=${this.selectGSEControl}
+      class="selectionlist"
       >${Array.from(this.doc.querySelectorAll('IED'))
         .sort(compareNames)
         .flatMap(ied => {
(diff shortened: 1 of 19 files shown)


src/editors/publisher/gse-control-editor.ts
```diff
@@ -3,25 +3,66 @@ import {
...
+        `DataSet[name="${gseControl.getAttribute('datSet')}"]`
+      );
+      (evt.target as FilteredList).classList.add('hidden');
+      this.selectGSEControlButton.classList.remove('hidden');
+    }
+  }
+
+  private renderElementEditorContainer(): TemplateResult {
+    if (this.selectedGseControl !== undefined)
+      return html`<div class="elementeditorcontainer">
```
danyill, line 52: I would like to beg for the use of kebab case on classes. Complete lowercase is pretty hard to read and parse on the run.

https://cssguidelin.es/#hyphen-delimited

test/unit/editors/publisher/sampled-value-control-editor.test.ts
```diff
@@ -19,6 +19,21 @@ describe('Editor for SampledValueControl element', () => {
     );
   });
 
-  it('looks like the latest snapshot', async () =>
+  it('initialy looks like the latest snapshot', async () =>
```
danyill, line 22:
```suggestion
  it('initially looks like the latest snapshot', async () =>
```

test/unit/editors/publisher/data-set-editor.test.ts
```diff
@@ -17,6 +17,21 @@ describe('Editor for DataSet element', () => {
     );
   });
 
-  it('looks like the latest snapshot', async () =>
+  it('initialy looks like the latest snapshot', async () =>
```
danyill, line 20:
```suggestion
  it('initially looks like the latest snapshot', async () =>
```

test/unit/editors/publisher/__snapshots__/sampled-value-control-editor.test.snap.js
```diff
@@ -1,86 +1,196 @@
...
-snapshots["Editor for SampledValueControl element looks like the latest snapshot"] = 
-`<filtered-list>
-  <mwc-list-item
-    aria-disabled="false"
-    class="header listitem"
-    graphic="icon"
-    noninteractive=""
-    tabindex="-1"
-    value=""
+snapshots["Editor for SampledValueControl element initialy looks like the latest snapshot"] =
```
danyill, line 4:
```suggestion
snapshots["Editor for SampledValueControl element initially looks like the latest snapshot"] = 
```

test/unit/editors/publisher/__snapshots__/report-control-editor.test.snap.js
```diff
@@ -1,103 +1,230 @@
...
-snapshots["Editor for ReportControl element looks like the latest snapshot"] = 
-`<filtered-list>
-  <mwc-list-item
-    aria-disabled="false"
-    class="header listitem"
-    graphic="icon"
-    noninteractive=""
-    tabindex="-1"
-    value=""
+snapshots["Editor for ReportControl element initialy looks like the latest snapshot"] =
```
danyill, line 4:
```suggestion
snapshots["Editor for ReportControl element initially looks like the latest snapshot"] = 
```

test/unit/editors/publisher/__snapshots__/gse-control-editor.test.snap.js
```diff
@@ -1,120 +1,264 @@
...
-snapshots["Editor for GSEControl element looks like the latest snapshot"] = 
-`<filtered-list>
-  <mwc-list-item
-    aria-disabled="false"
-    class="header listitem"
-    graphic="icon"
-    noninteractive=""
-    tabindex="-1"
-    value="IED1>>CircuitBreaker_CB1>GCB IED1>>CircuitBreaker_CB1>GCB2"
+snapshots["Editor for GSEControl element initialy looks like the latest snapshot"] =
```
danyill, line 4:
```suggestion
snapshots["Editor for GSEControl element initially looks like the latest snapshot"] = 
```

test/unit/editors/publisher/__snapshots__/data-set-editor.test.snap.js
```diff
@@ -1,139 +1,302 @@
...
-snapshots["Editor for DataSet element looks like the latest snapshot"] = 
-`<filtered-list>
-  <mwc-list-item
-    aria-disabled="false"
-    class="header listitem"
-    graphic="icon"
-    noninteractive=""
-    tabindex="-1"
-    value="IED1>>CircuitBreaker_CB1>GooseDataSet1"
+snapshots["Editor for DataSet element initialy looks like the latest snapshot"] =
```
danyill, line 4:
```suggestion
snapshots["Editor for DataSet element initially looks like the latest snapshot"] = 
```

danyill (commented):
Looks good to me. A few small comments:

image

(1) Could we have a Pro-mode edit XML button against the dataset? It's great being able to drop into XML quickly.
(2) I couldn't filter on `GGIO1`, perhaps because of the spaces? But I think I should be able to.
(3) Before and after the slash on `System` in the screenshot, should we have a space? How do we decide when to use a `.`, `/` or ` ` ? If I want to filter on e.g. `CFG/LLN0`, I should be able to but I can't currently. Maybe the slash is a mistake in the DataSet editor?
(4) I'm not sure if it is intended, but it feels a little odd in the Report, GOOSE and SampledValue radio button selections to see a dataset when there is no connecting element on the left-hand side. I guess this may come together in future PRs but it might be nice to see the word "Dataset" somewhere to make this clear.
(5) I like how the mobile responsive "Select Other Dataset etc" works.

src/editors/publisher/gse-control-editor.ts
JakobVogelsang: That is an iteresting guideline. Let's go with the hyphen first and see what else we can adopt :)

JakobVogelsang:
Thanks for the input @danyill. I did some changes and some need to be addressed later. See my remarks below.
> Looks good to me. A few small comments:
> 
> image
> 
> (1) Could we have a Pro-mode edit XML button against the dataset? It's great being able to drop into XML quickly. 

That is a single issue as is not so easy to do. I would like to have a dialog similar to the wizard but only with the code and that is not possible yet. Hope you can wait another couple of weeks for that. I would first have all read only elements and would then start to enrich it with buttons.

>(2) I couldn't filter on `GGIO1`, perhaps because of the spaces? But I think I should be able to. 

You can do `GGIO 1` which would do the trick. I still changed the display of the string to better search for details

>(3) Before and after the slash on `System` in the screenshot, should we have a space? How do we decide when to use a `.`, `/` or ` ` ? If I want to filter on e.g. `CFG/LLN0`, I should be able to but I can't currently. Maybe the slash is a mistake in the DataSet editor? 

You can do something like `CFG LLN0` that would do. You do not neccassery need to have the delimiter in your search string.

>(4) I'm not sure if it is intended, but it feels a little odd in the Report, GOOSE and SampledValue radio button selections to see a dataset when there is no connecting element on the left-hand side. I guess this may come together in future PRs but it might be nice to see the word "Dataset" somewhere to make this clear. (5) I like how the mobile responsive "Select Other Dataset etc" works.

You are right if you look into the mockup there will be other panes showing other information as well. I wanted to start with the one that is common to all others. Still, I will change the header a bit for it to be more clear.

src/editors/publisher/gse-control-editor.ts
JakobVogelsang: I just tried and saw that our linter complains. I would like to make a bit more conscious  decision. Please forgive me for not changing that. BUT I would like to have style guide for CSS applied. So this is a really for now :)

PR #913, opened 2022-07-30
feat(plugin): add read-only report-control-element-editor

JakobVogelsang:
This one is adding read-only field showing the settings of a `ReportControl` element and its child elements. Per default it is not displayed to save the room for the data set. I am curious on your take on it.

### DIFF
--- a/src/editors/Publisher.ts
+++ b/src/editors/Publisher.ts
@@ -21,6 +21,7 @@ export default class PublisherPlugin extends LitElement {
   /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
   @property({ attribute: false })
   doc!: XMLDocument;
+
   @state()
   private publisherType: 'Report' | 'GOOSE' | 'SampledValue' | 'DataSet' =
     'GOOSE';
--- a/src/editors/publisher/data-set-editor.ts
+++ b/src/editors/publisher/data-set-editor.ts
@@ -113,5 +113,9 @@ export class DataSetEditor extends LitElement {
 
   static styles = css`
     ${styles}
+
+    data-set-element-editor {
+      flex: auto;
+    }
   `;
 }
--- a/src/editors/publisher/data-set-element-editor.ts
+++ b/src/editors/publisher/data-set-element-editor.ts
@@ -18,6 +18,10 @@ import { identity } from '../../foundation.js';
 
 @customElement('data-set-element-editor')
 export class DataSetElementEditor extends LitElement {
+  /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
+  @property({ attribute: false })
+  doc!: XMLDocument;
+  /** The element being edited as provided to plugins by [[`OpenSCD`]]. */
   @property({ attribute: false })
   element!: Element | null;
 
@@ -47,17 +51,21 @@ export class DataSetElementEditor extends LitElement {
       </wizard-textfield>
       <filtered-list
         >${Array.from(this.element!.querySelectorAll('FCDA')).map(fcda => {
-          const [ldInst, prefix, lnClass, lnInst, doName, daName] = [
+          const [ldInst, prefix, lnClass, lnInst, doName, daName, fc] = [
             'ldInst',
             'prefix',
             'lnClass',
             'lnInst',
             'doName',
             'daName',
+            'fc',
           ].map(attributeName => fcda.getAttribute(attributeName) ?? '');
 
           return html`<mwc-list-item selected twoline value="${identity(fcda)}"
-            ><span>${doName + '.' + daName}</span
+            ><span
+              >${doName}${daName
+                ? '.' + daName + ' ' + '[' + fc + ']'
+                : ' ' + '[' + fc + ']'}</span
             ><span slot="secondary"
               >${ldInst + '/' + prefix + lnClass + lnInst}</span
             >
@@ -77,7 +85,10 @@ export class DataSetElementEditor extends LitElement {
       </div>`;
 
     return html`<div class="content">
-      <h2>${translate('publisher.nodataset')}</h2>
+      <h2>
+        <div>DataSet</div>
+        <div class="headersubtitle">${translate('publisher.nodataset')}</div>
+      </h2>
     </div>`;
   }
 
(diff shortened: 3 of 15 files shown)


danyill:
Thank you for this interface improvement and the opportunity to make some comments. 

I think it's looking excellent and is much more intuitive to find. It makes better use of the screen space compared to the "old way". I like that the user doesn't have to store the linking parameters in their head for the different dialogs and it's easy to look at another report or IED and "flick between them".

I find it more intuitive in the view with both the ReportControl and Dataset shown side-by-side and think this would be a better default than the minimised view.

I wonder if it would be better if the Dataset and ReportControl windows were always either expanded side by side or vertically stacked and without the minimise button, which feels a little confusing to me (we don't usually minimise to left or right like this, I think).

image

Visually, I wonder if:

(1) There is less space required between the list of reports and the detail

(2) I notice each data object or data attribute now ends in a `.`

(3) I wonder if the columns need either more space or a dividing line between them, particularly the toggle of the first column seems close to the check box on the second column (trigger options and optional fields). Either the different coloured background for a column gap between them or more space or a vertical line might help? WDYT?

Also:

(4) It's slightly irritating to scroll the list on the left when there are many reports without datasets (true for many IEDs by default) in a large file. Should we be able to remove empty/unused reports with a tick box or filter? Maybe for future?

(5) When there are no reports in a new scd file which I open, the state from the previous file is retained on the right. I think that the right-hand side is not being regenerated when a new file is loaded.

image

(6) When a dataset is small, and the ReportControl has no parameters, the gap at the top between ReportControl and DataSet seems a little awkward. Perhaps a flex align to top?

image

(7)  I like the titles for the elements, but I think the title should be retained even if there is no DataSet, and beneath that, it should say "No DataSet referenced" (in a slightly smaller font).

image

I didn't look at the code, but here are some immediate thoughts.

danyill:
Also, I like the displays for the Trigger Options and Optional Fields.
I think they're clever and make good sense.
But I think they're not intuitive -- I have to sit and look at them.
I appreciate that our UI is very minimalist, but I think we need to add something here.

Could we have a header?

image

I'm not sure what the best words are, perhaps one of the following:

* Enabled / Supported
* Configured / Available
* Selected / Possible (< bit too loose)

I think I favour the first one.

JakobVogelsang:
> Thank you for this interface improvement and the opportunity to make some comments.
> 
> I think it's looking excellent and is much more intuitive to find. It makes better use of the screen space compared to the "old way". I like that the user doesn't have to store the linking parameters in their head for the different dialogs and it's easy to look at another report or IED and "flick between them".
> 
> I find it more intuitive in the view with both the ReportControl and Dataset shown side-by-side and think this would be a better default than the minimised view.

I have the feeling that a non-experienced user might be overwhelming by the 20+ settings. I think we need to hide this complexity, either by default or by user choice. But you are right, my proposal is not good yet. I decided to not have this feature and at it at a later stage.

> 
> I wonder if it would be better if the Dataset and ReportControl windows were always either expanded side by side or vertically stacked and without the minimise button, which feels a little confusing to me (we don't usually minimise to left or right like this, I think).

I have the feeling that a non-experienced user might be overwhelming by the 20+ settings. I think we need to hide this complexity, either by default or by user choice. But you are right, my proposal is not good yet. I decided to not have this feature and at it at a later stage.

> 
> image
> 
> Visually, I wonder if:
> 
> (1) There is less space required between the list of reports and the detail

I would fix it in a later stage. I have not found a good solution yet other than the percentage divide, which with little screens make sense but with big once does not.

> 
> (2) I notice each data object or data attribute now ends in a `.`

That is a bug with FCD type DataSets that are used in reports a lot.

> 
> (3) I wonder if the columns need either more space or a dividing line between them, particularly the toggle of the first column seems close to the check box on the second column (trigger options and optional fields). Either the different coloured background for a column gap between them or more space or a vertical line might help? WDYT?

Good idea, added a vertical line.

> 
> Also:
> 
> (4) It's slightly irritating to scroll the list on the left when there are many reports without datasets (true for many IEDs by default) in a large file. Should we be able to remove empty/unused reports with a tick box or filter? Maybe for future?

Yeah I would say a feature for later stage.

> 
> (5) When there are no reports in a new scd file which I open, the state from the previous file is retained on the right. I think that the right-hand side is not being regenerated when a new file is loaded.

That was a bug and is fixed here right away.

> 
> image
> 
> (6) When a dataset is small, and the ReportControl has no parameters, the gap at the top between ReportControl and DataSet seems a little awkward. Perhaps a flex align to top?

Could not reproduce that. Would you share the file?

> 
> image
> 
> (7) I like the titles for the elements, but I think the title should be retained even if there is no DataSet, and beneath that, it should say "No DataSet referenced" (in a slightly smaller font).

Thanks good idea!

> 
> image
> 
> I didn't look at the code, but here are some immediate thoughts.

Thanks much appreciate as always!

JakobVogelsang:
@danyill I fixed most of the issue. Some I see for future adoption. One of those is the heading in wizard-checkbox. I think it is not so critical now that there are vertical lines. I am currious what you say :)

danyill:
I think the vertical lines look great and are nicely in keeping with the visual style. Very nice :tada: 

>I have the feeling that a non-experienced user might be overwhelming by the 20+ settings. I think we need to hide this complexity, either by default or by user choice. But you are right, my proposal is not good yet. I decided to not have this feature and at it at a later stage.

If we are worried about the busyness, my suggestion is to make the Trigger Options and the Optional Fields items that could drop down / be expanded but were compressed by default. With that, there would be a reasonable minimum amount of information shown (core parameters to setup a report without worrying about the details).

>>    (6) When a dataset is small, and the ReportControl has no parameters, the gap at the top between ReportControl and DataSet seems a little awkward. Perhaps a flex align to top?

>Could not reproduce that. Would you share the file?

I think this has gone away with other changes, in fact, you may have fixed a bug somehow because I can't reproduce it, now I get what is shown below (which seems better) :+1: 

image

>One of those is the heading in wizard-checkbox. I think it is not so critical now that there are vertical lines. 

Let's let this rest for now :-)

danyill (approved):
Thank you for taking this feedback on board, I think this is great. I have added one more comment if we want it to look less busy, but I'd be happy to merge from a user's point of view.

PR #917, opened 2022-08-02
feat(editors/publisher): add read only gse-control-element-editor

JakobVogelsang:
Adding read only GSEControl element editor with same UI as for Report. Would love to hear your review.


danyill (approved):

JakobVogelsang:
@danyill I will do so with the SampledValue read-only setting. Thx for hint.

PR #920, opened 2022-08-03
feat(editors/publisher): add sampled-value-control-element-editor

JakobVogelsang:
The last read-only part. From now on it becomes interesting :)


src/editors/publisher/sampled-value-control-editor.ts
```diff
@@ -40,12 +58,15 @@ export class SampledValueControlEditor extends LitElement {
...
+
+    this.selectedSampledValueControl = smvControl;
 
     if (smvControl) {
-      this.selectedSampledValueControl =
+      this.selectedDataSet =
         smvControl.parentElement?.querySelector(
           `DataSet[name="${smvControl.getAttribute('datSet')}"]`
-        );
+        ) ?? null;
```
dlabordus, line 69: Is this part (?? null) needed? QuerySelector returns (Element |  null) by default.

src/editors/publisher/sampled-value-control-element-editor.ts
```diff
@@ -0,0 +1,269 @@
...
+@customElement('sampled-value-control-element-editor')
+export class SampledValueControlElementEditor extends LitElement {
+  /** The document being edited as provided to plugins by [[`OpenSCD`]]. */
+  @property({ attribute: false })
+  doc!: XMLDocument;
+  /** The element being edited as provided to plugins by [[`OpenSCD`]]. */
+  @property({ attribute: false })
+  element!: Element;
+  @property({ attribute: false })
+  get sMV(): Element | null | undefined {
```
dlabordus, line 32: Why return both null and undefined? Is it really needed and I prefer to choose one of the two.

src/editors/publisher/sampled-value-control-element-editor.ts
```diff
@@ -0,0 +1,269 @@
...
+        maxLength="${maxLength.cbName}"
+        dialogInitialFocus
+      ></wizard-textfield>
+      <wizard-textfield
+        label="desc"
+        .maybeValue=${desc}
+        nullable
+        helper="${translate('scl.desc')}"
+      ></wizard-textfield>
+      ${multicast === 'true'
```
dlabordus, line 158: Isn't a little bit strange to only show the checkbox when multicast is false?

src/editors/publisher/sampled-value-control-element-editor.ts
```diff
@@ -0,0 +1,269 @@
...
+        `SMV[ldInst="${ldInst}"][cbName="${cbName}"]`
+    );
+  }
+
+  private renderSmvContent(): TemplateResult {
+    const sMV = this.sMV;
+    if (!sMV)
+      return html` <h3>
+        <div>Communication Settings (SMV)</div>
+        <div class="headersubtitle">No connection to SubNetwork</div>
```
dlabordus, line 50: Shouldn't this be done through lit-translate?

src/editors/publisher/sampled-value-control-element-editor.ts
```diff
@@ -0,0 +1,269 @@
...
+    const hasInstType = Array.from(sMV.querySelectorAll('Address > P')).some(
+      pType => pType.getAttribute('xsi:type')
+    );
+
+    const attributes: Record<string, string | null> = {};
+
+    ['MAC-Address', 'APPID', 'VLAN-ID', 'VLAN-PRIORITY'].forEach(key => {
+      if (!attributes[key])
+        attributes[key] =
+          sMV.querySelector(`Address > P[type="${key}"]`)?.innerHTML.trim() ??
```
dlabordus, line 62: Why not use ".textContent", ".innerHTML" looks a little strange for a XML Document.

test/unit/editors/publisher/sampled-value-control-element-editor.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  let doc: XMLDocument;
+  let element: SampledValueControlElementEditor;
+
+  beforeEach(async () => {
+    doc = await fetch('/test/testfiles/wizards/sampledvaluecontrol.scd')
+      .then(response => response.text())
+      .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+  });
+
+  describe('with multicast attribute set to false - deprecated', () => {
```
dlabordus, line 16: deprecated? Shouldn't just be removed?

dlabordus (changes requested):

src/editors/publisher/sampled-value-control-editor.ts
JakobVogelsang: I would argue yes: with the `DataSet` I use the undefined to either render or not render the element editor. Undefined means that something not control block is selected. Null means that a control block an item is selected but it is an empty control block. That is why the `data-set-element-editor` has an element property `Element | null`. That is for example not the case for the control block itself where it can be undefined or must be an element.

src/editors/publisher/sampled-value-control-element-editor.ts
JakobVogelsang: That is true I will change that :)

src/editors/publisher/sampled-value-control-element-editor.ts
JakobVogelsang: Yea, that is the result from a discussion we had on the edit wizard for `SampledValueControl`. We agreed there that highlighting that multicast is not set to true should be done to allow users to a potential configuration error. Other that that is just disturbing to have as that thing is deprecated. My guess is that nobody will ever see it :)

src/editors/publisher/sampled-value-control-element-editor.ts
JakobVogelsang: That is true. That is the case for all the other header. I will correct it. Thx for the hint

src/editors/publisher/sampled-value-control-element-editor.ts
JakobVogelsang: True! Thx

test/unit/editors/publisher/sampled-value-control-element-editor.test.ts
JakobVogelsang: See comment above. I think it is valid to highlight potential errors in the file. If today a file with multicast error appears, that is rather an error.

dlabordus: I would prefer to not just use the term 'deprecated' like this, because most of the time it means cleaning up the code to developers.
Maybe change the text to 'with "deprecated" multicast attribute set to false'. This way it's more clear that the attribute is deprecated and not the software itself.

JakobVogelsang: I see that is true. Will cahnge that :)

dlabordus (approved):

PR #921, opened 2022-08-03
feat(editors/publisher): add code-editor-dialog

JakobVogelsang:
@danyill I was thinking to add a code editor to the publisher on every element as you suggested. That would be a very fast way to allow editing of the elements and would make the editor already quite usable. 

The tricky part here was to re-render properly. There is no clean way to distinguish between doc change on open project or doc change due to an editor action at the moment because in both cases we exchange the doc. So I had to do an assumption:
>When the new doc has an element at the same location in the file, I assume there was a change due an editor action. That means also that when you load an older version of a file that has also the same element (same identity string) then the element stays selected, and the values are updated.

I would love to have your review for this PR a code review as well. Maybe there is a smarter way to implement the pure code editor. 

Tests are still missing. I would do those as soon as I know the principal solution is good enough

### DIFF
--- a/src/editors/publisher/code-editor-dialog.ts
+++ b/src/editors/publisher/code-editor-dialog.ts
@@ -1,0 +1,103 @@
+import {
+  customElement,
+  css,
+  LitElement,
+  property,
+  TemplateResult,
+  html,
+  query,
+} from 'lit-element';
+import { get, translate } from 'lit-translate';
+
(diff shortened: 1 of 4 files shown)


JakobVogelsang:
With some feedback from @ca-d I will close the PR as is. We need something a bit more general here I think than this web component that can be reused in other cases.

PR #922, opened 2022-08-08
feat(menu/history): move history to menu plugin

### LINKED ISSUE #916: Extract SCL History from Log Dialog
With the new OpenSCD Core repository being created we want to separate the SCL History from the Logging Dialog. 
Currently the Hitem Element are loaded when the XML Document is loaded and converted to LogDetail through events.
We want to separate this logic and only show logging event in the logging dialog.

In the new situation a new menu item will be created as a menu plugin to display the SCL History in it's own window.
So we need to create a new menu plugin that will display a dialog with the information about the Hitem Elements in the SCL File.
The loading part of the Hitme Element in Logging.ts can be removed, including all related stuff, like SclhistoryEntry and SclhistoryDetail.

The new dialog will show a line for each Hitem found in the Header/History Element. 
Use the first line of A Two-Line MWC List Item to show the 'When' followed by 'What' attributes.
On the the second line show, 'who' and 'why' separated by a column. `${who} : ${why}`
Look at the logging window of the current implementation for more details how the styling is set.

**Requirements**
New menu items to show the History Items of a SCL File.

**Additional information**

juancho0202:
Closes #916


src/menu/History.ts
```diff
@@ -0,0 +1,124 @@
...
+  ): SclhistoryEntry {
+    return {
+      kind: 'sclhistory',
+      title: what ?? 'UNDEFINED',
+      message: this.createMessage(who, why),
+      time: this.convertToDate(when),
+    };
+  }
+
+  get sclHistory(): SclhistoryEntry[] {
```
dlabordus, line 56: Why not just return a Element Array and use the createMessage in the render function?
This way the object SclHistoryEntry can probably also be removed, because it isn't used any more.

src/menu/History.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  async run(): Promise<void> {
+    this.historyLog.open = true;
+  }
+
+  renderSclHistoryEntry(entry: SclhistoryEntry): TemplateResult {
+    return html` <abbr title="${entry.title}">
+      <mwc-list-item
+        class="${entry.kind}"
```
dlabordus, line 76: Is always 'sclhistory', so maybe move the styling to this dialog and just hard-coded the class here?

src/menu/History.ts
```diff
@@ -0,0 +1,124 @@
...
+        graphic="icon"
+        ?twoline=${!!entry.message}
+      >
+        <span>
+          <tt>${entry.time?.toLocaleString()}</tt>
+          ${entry.title}</span
+        >
+        <span slot="secondary">${entry.message}</span>
+        <mwc-icon
```
dlabordus, line 85: I don't think this is needed like this anymore, because there is no toggle anymore.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [0.21.0](https://github.com/openscd/open-scd/compare/v0.20.0...v0.21.0) (2022-08-08)
```
dlabordus, line 5: What happened here? This should be a change.
Also for the files manifest.json and package*.json

There are all changes here that are already in main since yesterday. 
So maybe first look what happened here before starting the review again.

dlabordus (changes requested):

src/translations/de.ts
```diff
@@ -84,6 +84,7 @@ export const de: Translations = {
     new: 'Neues projekt',
     title: 'Menü',
     viewLog: 'Protokoll anzeigen',
+    viewSclHistory: 'View SCL history',
```
dlabordus, line 87: Where is this label used?

src/menu/SclHistory.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  private renderSclHistory(): TemplateResult[] | TemplateResult {
+    if (this.sclHistory.length > 0)
+      return this.sclHistory
+        .slice()
+        .reverse()
+        .map(this.renderSclHistoryEntry, this);
+    else
+      return html`<mwc-list-item disabled>
+        <span>${translate('sclHistory.noEntries')}</span>
```
dlabordus, line 74: This label is not there (anymore).
I think something went wrong in the merge.

Because you are using two labels 'sclHistory.noEntries' and 'history.name'.

src/translations/de.ts
```diff
@@ -165,6 +165,10 @@ export const de: Translations = {
       placeholder: 'Keine Fehler',
     },
   },
+  history: {
+    name: '????',
```
JakobVogelsang, line 169:
```suggestion
    name: 'SCL History',
```

src/translations/de.ts
```diff
@@ -165,6 +165,10 @@ export const de: Translations = {
...
+    noEntries: '????',
```
JakobVogelsang, line 170:
```suggestion
    noEntries: 'Keine Einträge in der SCL History',
```

src/menu/SclHistory.ts
JakobVogelsang, line 74:
```suggestion
        <span>${translate('history.noEntries')}</span>
```

dlabordus: I suggested sclHistory, because history is so general.

JakobVogelsang: My bet sorry!!!! My remark is that with no history it shows me sclHistor.noEntries in the UI.

JakobVogelsang: The translation files to not meet the string here yet.

dlabordus (approved):

JakobVogelsang (approved):
LGTM

PR #926, opened 2022-08-09
feat(menu/compareied): ignore attributes/elements when comparing IED Elements

pascalwilbrink:
Added the possibility to ignore attributes and/or elements when checking for differences between 2 elements.


src/menu/CompareIED.ts
```diff
@@ -138,10 +144,71 @@ export default class CompareIEDPlugin extends LitElement {
     ></mwc-button>`;
   }
 
+  protected renderFilterCheckbox(): TemplateResult {
+    return html`<mwc-formfield
+      label="Filter mutables">
```
JakobVogelsang, line 149: Please translate! My suggestion:

en: `Filter preject specific settings`
de: `Projektspzifische Unterschiede ausblenden`

src/menu/CompareIED.ts
```diff
@@ -138,10 +144,71 @@ export default class CompareIEDPlugin extends LitElement {
...
+        <mwc-checkbox
+          ?checked=${this.filterMutables}
+          @change=${() => this.filterMutables = !this.filterMutables}>
+        </mwc-checkbox>
+      </mwc-formfield>
+      `
+  }
   protected renderCompare(): TemplateResult {
-    return html`${renderDiff(
+    const filterToIgnore: DiffFilter<Element> = {
```
JakobVogelsang, line 158: I would make it a global variable for now. If in a later stage we make this user configurable we can change. But here it makes the render method be hard to read.

test/testfiles/VS737-template.scd
```diff
@@ -0,0 +1,524 @@
+<SCL xmlns="http://www.iec.ch/61850/2003/SCL" version="2007" revision="B" release="4">
```
JakobVogelsang, line 1: I would change the name of the SCD file to be more descriptive. The issue number in Allianders Jira is not so much :)

test/testfiles/VS737-withchanges.scd
```diff
@@ -0,0 +1,524 @@
+<SCL xmlns="http://www.iec.ch/61850/2003/SCL" version="2007" revision="B" release="4">
```
JakobVogelsang, line 1: I would change the name of the SCD file to be more descriptive. The issue number in Allianders Jira is not so much :)

dlabordus (changes requested):

test/testfiles/VS737-template.scd
dlabordus: Our maybe see if we can do something with the two existing files, because there are already two files used.
Copy some extra IED's in there to do the trick.

src/foundation/compare.ts
```diff
@@ -14,6 +14,69 @@ export type Diff<T> =
...
+type DiffFilterConsumer<T> = boolean | ((value: T | null) => boolean);
+
+function getDiffFilterSelector(
+  elementToBeCompared: Element,
+  rootElementToBeCompared: Element,
+  filters: DiffFilter<Element>
+): DiffFilterSelector<Element> | undefined {
+  const querySelector: string | undefined =
+    rootElementToBeCompared === elementToBeCompared
+      ? ':root'
```
dlabordus, line 45: Shouldn't this be ':scope'?

test/testfiles/VS737-template.scd
pascalwilbrink: fixed: Moved it into the existing file

test/testfiles/VS737-withchanges.scd
pascalwilbrink: fixed: Moved it into the existing file

src/foundation/compare.ts
```diff
@@ -32,8 +95,15 @@ function describe(element: Element): string {
  */
 export function diffSclAttributes(
   elementToBeCompared: Element,
-  elementToCompareAgainst: Element
+  elementToCompareAgainst: Element,
+  filterToIgnore?: DiffFilter<Element>,
+  rootElementToBeCompared?: Element,
+  rootElementToCompareAgainst?: Element
 ): [string, Diff<string>][] {
+  rootElementToBeCompared = rootElementToBeCompared || elementToBeCompared;
```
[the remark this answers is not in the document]
pascalwilbrink: Fixed: removed it.

src/foundation/compare.ts
```diff
@@ -111,30 +203,58 @@ export function isSame(newValue: Element, oldValue: Element): boolean {
  */
 export function diffSclChilds(
   elementToBeCompared: Element,
-  elementToCompareAgainst: Element
+  elementToCompareAgainst: Element,
+  filterToIgnore?: DiffFilter<Element>,
+  rootElementToBeCompared?: Element,
+  rootElementToCompareAgainst?: Element
 ): Diff<Element>[] {
+  rootElementToBeCompared = rootElementToBeCompared || elementToBeCompared;
```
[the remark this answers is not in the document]
pascalwilbrink: fixed: Removed it

src/foundation/compare.ts
```diff
@@ -149,23 +269,37 @@ export function diffSclChilds(
  */
 export function renderDiff(
   elementToBeCompared: Element,
-  elementToCompareAgainst: Element
+  elementToCompareAgainst: Element,
+  filterToIgnore: DiffFilter<Element> = {},
+  rootElementToBeCompared?: Element,
```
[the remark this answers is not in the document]
pascalwilbrink: 1) Because of the question mark, the root/search elements are optional.
2) I renamed them to searchElement

I can extract the rendering function like this:

```code
function render(elementA, elementB) { return internalRender(elementA, elementB, elementA, elementB); }

function internalRender(elementA, elementB, searchElementA, searchElementB) { ... }
```

src/foundation/compare.ts
pascalwilbrink: Fixed: changed to scope (i saw that :root and :scope did the same thing)

src/foundation/compare.ts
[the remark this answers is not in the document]
dlabordus: This is better, because then the API is more clear if used in plugins.

src/foundation/compare.ts
dlabordus: I'm not full sure. 
Searching on the internet I think they are only the same when used in a Style Sheet File.
But in this case :root is the root document of the XML and for the IED that is the SCL Element. 
And :scope is the element on which the querySelector is executed, so here IED for instance.

```
14.1. :root pseudo-class
The :root pseudo-class represents an element that is the root of the document.

For example, in a DOM document, the :root pseudo-class matches the root element of the Document object. In HTML, this would be the html element (unless scripting has been used to modify the document).
```

```
8.6. The Reference Element Pseudo-class: :scope
In some contexts, selectors can be matched with an explicit set of :scope elements. This is a (potentially empty) set of elements that provide a reference point for selectors to match against, such as that specified by the querySelector() call in [DOM].

The :scope pseudo-class represents any element that is a :scope element. If the :scope elements are not explicitly specified, but the selector is scoped and the scoping root is an element, then :scope represents the scoping root; otherwise, it represents the root of the document (equivalent to :root). Specifications intending for this pseudo-class to match specific elements rather than the document’s root element must define either a scoping root (if using scoped selectors) or an explicit set of :scope elements.
```

src/menu/CompareIED.ts
pascalwilbrink: Fixed

src/menu/CompareIED.ts
pascalwilbrink: Fixed, moved outside class. Should this be in a different file as well?

src/foundation/compare.ts
```diff
@@ -189,21 +331,26 @@ export function renderDiff(
     attrDiffs.length > 0 ||
     childAddedOrDeleted.length > 0
   ) {
-    return html` ${attrDiffs.length > 0 || childAddedOrDeleted.length > 0
-      ? html` <mwc-list multi>
-          ${attrDiffs.length > 0
-            ? html` <mwc-list-item noninteractive ?twoline=${!!idTitle}>
+    return html` ${
```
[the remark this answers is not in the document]
pascalwilbrink: I don't know what happened. I used the standard prettier formatter

src/foundation/compare.ts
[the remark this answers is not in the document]
pascalwilbrink: Fixed

JakobVogelsang (approved):
LGTM

PR #940, opened 2022-08-12
feat(foundation/components): add pure UI oscd-dialog

JakobVogelsang:
For the merge with `open-scd-core` the API that opens a `wiazrd-dialog` will be available for standard SCL elements are defined in IEC 61850-6. We have however a lot of case where `wizard-dialog` is used for other purposes as well. I therefore invested some time to create a web component that copies some `wizard-dialog` behavior to allow a quicker refactor in the future. 

This is what the component can do for you:
1. Automatically renders primary action button when primary action is defined
2. Always renders secondary action `Close` button
3. Allow to define an element to show in a code editor. Allow to toggle between code editor and the dialog.
4. You can define your own code action. I do not want mix this pure UI element with any functionality. But we can of course define the function used in `wizard-dialog` somewhere in the foundations and reuse it.
5. Allows to slot any elements into the dialog you want. 

This is what it cannot to compared to `wizard-dialog`
1. Closing and opening of the dialog is the user's obligation now. There is no mechanism to watch when which dialog is opened. So if you want to have multiple dialogs to opened in a sequence, you have to add as many `oscd-dialog`s you need and open/close those in a specific order you need.
2. `oscd-dialog` other than `wizard-dialog` consist only of one page only. You cannot define a multi pages dialog, where the dialog handles the jumps to the next and/or the previous page. 
3. There is not checking of validity of the inputs. So when clicking in primary action, the dialog will be closed even on invalid inputs. 

The component is not ready to use yet!! I wanted to have your opinion on the feature set it has to have before proceeding.
 - Point 1:  I cannot think of a way to add 1 with the way `oscd-dailog` is implemented as a pure UI component. Do you?
 - Point 2: Do we really need that. From looking through the code I have not seen it be used other than `LNode`, `GSEControl`, `Report` and `SampledValuesControl` and those are going to be covered by `open-scd-core` wizard library
 - Point 3: Do we need that? I think yes, but am not sure at all here.

Please have a look at the tests, if you want to see how the web.component can be used.

### DIFF
--- a/src/foundation/components/oscd-dialog.ts
+++ b/src/foundation/components/oscd-dialog.ts
@@ -1,0 +1,140 @@
+import {
+  css,
+  customElement,
+  html,
+  LitElement,
+  property,
+  query,
+  TemplateResult,
+} from 'lit-element';
+import { translate } from 'lit-translate';
+
+import '@material/mwc-dialog';
+import '@material/mwc-button';
+import '@material/mwc-icon-button-toggle';
+import { Dialog } from '@material/mwc-dialog';
+
+import 'ace-custom-element';
+
+@customElement('oscd-dialog')
+export class OscdDialog extends LitElement {
+  /** Whether dialog is open */
+  @property({ type: Boolean })
+  open = false;
+
+  /** dialogs header content*/
+  @property({ type: String })
+  heading = '';
+
+  /** callback on primary action button click */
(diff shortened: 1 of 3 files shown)


JakobVogelsang:
Will be developed in `open-scd-core` and can be used in `open-scd` through npm

Issue #929, opened 2022-08-11
Schema invalid file when subscribing


JakobVogelsang:
**Describe the bug**
Subscribing causes schema invalidity

**To Reproduce**
1. open project with two `IED`s and one `GSEControl`
2. got to Subscriber editor plugin
3. subscribe GOOSE
4. see error from schema validator

**Expected behavior**
OpenSCD action should not be schema invalid

**Additional context**
The error pops up, because the attribute `srcLNInst` in the element `ExtRef` cannot be an empty string. Instead of an empty string, the attribute should not be present in the file.

danyill:
I have tested and believe this also applies to sampled values in the Subscriber(SMV) plugin.

Additionally, I think it is possible `lnInst` is entirely missing, which is allowable (only) for LLN0 according to Part 6 (Table 8). At the moment we don't handle this correctly either if I am not mistaken. Given that LLN0 is where GSEControl and SampledValueControl elements must reside we need to cater for this.

I will have a PR for this in a day or two, I hope.

danyill:
Eek! And then I see @dlabordus got there first!

PR #944, opened 2022-08-17
feat(editor/laterbinding): Subscribe and unsubscribe from ExtRef for Later Binding (SMV)

### LINKED ISSUE #866: Subscriber Later Binding (SMV): Connect FCDA to ExtRef
As a user of OpenSCD I want to connect a `FCDA` element to the `ExtRef` element. 

**Requirements**
- add Ed1 attributes
1. copy`ldInst`, `prefix`, `lnClass`, `lnInst`, `doName` and  `daName` to the `ExtRef` element
2. add `iedName` that is the `name` attribute of the `IED` element `FCDA` resides in
- add Ed2 attributes
1. `serviceType` is `SMV`
2. `srcLDInst` is the `inst` attribute of the `LDevice` element where the source `SampledValueControl` resides in
5. `srcPrefix` is empty string as `SampledValueControl` are always in `LN0`
6. `srcLNClass` is `LLN0` as `SampledValueControl` are always in `LN0`
7. `srcLNInst` is null as `SampledValueControl` are always in `LN0` (empty string is schema invalid)
8. `srcCBName` is the `name` attribute of the `SampledValueControl` 

- do only connect without present attributes `pLN`, `pDO`, `pDA` and `pServT` (check for those in extra issue,  see also additional information)
- make this a default plugin: set default setting in plugin registry to true.

**Additional Information**

Checking `pDO` means to check if the `CDC` meet. That, however, means 
1. based on the attributes `lnClass` and `doName` in the `FCDA` element, parse the `DOType` and take the `cdc` attribute
2. do the same in the source IED with the attributes `pLN` and `pDO` and get the `cdc` from the `DOType` element.
3. compare those
! `doName` can be something like this: A.phsA!

Checking `pDA` means checking if the `bType` is met. This is the same procedure as with the `pDO` instead that the element carrying is `DA` or `BDA`. That is only possible if all `pLN`, `pDO` and `pDA` are given.
! `daName` can be something like this: `instMag.i`

### LINKED ISSUE #868: Subscriber Later Binding (SMV): Disconnect FCDA from ExtRef
As a user of OpenSCD I want to disconnect an existing later binding connection (`ExtRef` with existing `intAddr`)

**Requirements:**
Remove attributes from `ExtRef`
- `iedName`
- `ldInst`
- `prefix`
- `lnClass`
- `lnInst`
- `doName`
- `daName`
- `serviceType`
- `srcLDInst`
- `srcPrefix`
- `srcLNClass`
- `srcLNInst`
- `srcCBName`

Leave attributes in `ExtRef`
- `intAddr`
- `pLN`
- `pDO`
- `pDA`
- `desc`

dlabordus:
closes #866 and closes #868 

And because of re-using parts and setting up the code like the other subscribing also solved #929


test/integration/editors/GooseControlSubscription.test.ts
```diff
@@ -334,11 +332,9 @@ describe('GOOSE subscriber plugin', () => {
                       ?.getAttribute('inst')}"][srcPrefix="${
                       fcda.closest('LN0')?.getAttribute('prefix') ?? '' //prefix is mendatory in ExtRef!!
                     }"][srcLNClass="${fcda
-                      .closest('LN0')
-                      ?.getAttribute('lnClass')}"][srcLNInst="${fcda
                       .closest('LN0')
                       ?.getAttribute(
-                        'inst'
+                        'lnClass'
```
JakobVogelsang, line 337: Same here, it does not seem correct?

test/integration/editors/SMVSubscriberLaterBinding.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    element = await fixture(
+      html` <smv-subscribe-later-binding-plugin
+        .doc=${doc}
+      ></smv-subscribe-later-binding-plugin>`
+    );
+    await element.requestUpdate();
+  });
+
+  it('when subscribing a available ExtRef then the list are changed', async () => {
```
JakobVogelsang, line 31:
```suggestion
  it('when subscribing an available ExtRef then the list are changed', async () => {
```

test/integration/editors/SampledValuesSubscription.test.ts
```diff
@@ -136,11 +136,9 @@ describe('Sampled Values Plugin', () => {
                     ?.getAttribute('inst')}"][srcPrefix="${
                     fcda.closest('LN0')?.getAttribute('prefix') ?? '' //prefix is mendatory in ExtRef!!
                   }"][srcLNClass="${fcda
-                    .closest('LN0')
-                    ?.getAttribute('lnClass')}"][srcLNInst="${fcda
                     .closest('LN0')
                     ?.getAttribute(
-                      'inst'
+                      'lnClass'
```
JakobVogelsang, line 141: Does not seem to be correct.

test/integration/editors/SampledValuesSubscription.test.ts
```diff
@@ -325,11 +323,9 @@ describe('Sampled Values Plugin', () => {
                       ?.getAttribute('inst')}"][srcPrefix="${
                       fcda.closest('LN0')?.getAttribute('prefix') ?? '' //prefix is mendatory in ExtRef!!
                     }"][srcLNClass="${fcda
-                      .closest('LN0')
-                      ?.getAttribute('lnClass')}"][srcLNInst="${fcda
                       .closest('LN0')
                       ?.getAttribute(
-                        'inst'
+                        'lnClass'
```
JakobVogelsang, line 328: Does not seem to be correct.

JakobVogelsang (changes requested):
Thx for this PR! Overall, it looks very good. I have found just some little things and in my opinion four things that need to be big fixed in the tests. 

I would like to play around with some files that I have, just to make sure I did not overlook thinks while writing the requirements.

src/foundation.ts
```diff
@@ -486,6 +486,16 @@ export function referencePath(element: Element): string {
   return path;
 }
 
+export type SclEdition = '2003' | '2007B' | '2007B4';
```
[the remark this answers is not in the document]
dlabordus: But version is a bit misleading, because there is a attribute version. Add how do we call Version + Revision + Release?

test/integration/editors/GooseControlSubscription.test.ts
```diff
@@ -139,11 +139,9 @@ describe('GOOSE subscriber plugin', () => {
...
```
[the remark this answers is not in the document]
JakobVogelsang: See it now! Sorry

src/editors/subscription/foundation.ts
```diff
@@ -137,19 +147,111 @@ export function createExtRefElement(
...
+    srcLDInst,
+    srcPrefix,
+    srcLNClass,
+    srcLNInst: srcLNInst ? srcLNInst : null,
+    srcCBName,
+  });
+}
+
+/**
+ * Create a clone of the passed ExtRefElement and updated or set the required attributes oon the cloned element
```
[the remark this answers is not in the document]
dlabordus: Fixed.

src/editors/subscription/foundation.ts
```diff
@@ -137,19 +147,111 @@ export function createExtRefElement(
...
+    controlElement.closest('LDevice')?.getAttribute('inst') ?? '';
   const srcPrefix =
-    controlBlock.closest('LN0,LN')?.getAttribute('prefix') ?? '';
+    controlElement.closest('LN0,LN')?.getAttribute('prefix') ?? '';
   const srcLNClass =
-    controlBlock.closest('LN0,LN')?.getAttribute('lnClass') ?? '';
-  const srcLNInst = controlBlock.closest('LN0,LN')?.getAttribute('inst') ?? '';
-  const srcCBName = controlBlock.getAttribute('name') ?? '';
+    controlElement.closest('LN0,LN')?.getAttribute('lnClass') ?? '';
+  const srcLNInst =
```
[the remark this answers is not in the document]
dlabordus: Fixed it, but at line 264 it's still needed, because when changing line 248 to 
`const srcLNInst = controlElement.closest('LN0,LN')?.getAttribute('inst');`
The variable srcLNInst can be a `string | null | undefined`, because of 'closest'.

So at line 264 (now became 263) the check was still needed to change undefined to null.
And also to fix if the attribute 'inst' was there, but as empty string to null.

src/foundation.ts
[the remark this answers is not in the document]
dlabordus: As discussed changed it to 'getSclSchemaVersion'.

test/integration/editors/GooseControlSubscription.test.ts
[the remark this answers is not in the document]
dlabordus: As discussed, this was correct, but the diff is a bit misleading, because how the formatter changed it.

test/integration/editors/SMVSubscriberLaterBinding.test.ts
dlabordus: Fixed it.

src/editors/subscription/foundation.ts
```diff
@@ -137,19 +147,109 @@ export function createExtRefElement(
...
+    lnClass,
+    lnInst,
+    prefix,
+    doName,
+    daName,
+    srcLDInst,
+    srcPrefix,
+    srcLNClass,
+    srcLNInst: srcLNInst ? srcLNInst : null,
+    srcCBName,
```
danyill, line 175: This would solve the issue that is identified in #929 but I think this logic is still incorrect since if I understand correctly, `inst` can be entirely missing in LLN0, and we should cater for this.

I was thinking of doing something like this (although I suspect it could be written more elegantly!):

```ts
  // a null value for srcLNInst means this parameter isn't written if an instance number isn't found
  // or if the instances attribute is missing
  const lnInstValue = controlBlock.closest('LN0,LN')?.getAttribute('inst') ?? '';
  const srcLNInst = lnInstValue === '' || lnInstValue === null ? null : lnInstValue;
```

JakobVogelsang (approved):
LGTM

src/editors/subscription/foundation.ts
JakobVogelsang: 1. No, this is not true. If you look closer into the schemas, than `inst` must always be set. With `LN0` it is fixed to an empty string (>Ed2). With Ed1 it is required but not fixed. In both cases `inst` must be present.
2. Even if it is not present in the files, we cover that by setting `srcLNInst` to null in this case. :)

JakobVogelsang: See line 174 also :)

danyill: Quite right, thank you. I think I should be :sleeping:

JakobVogelsang: Good night, and thanks for your detailed look on it. I do very much appreciate this!!!

Issue #860, opened 2022-07-11
Subscriber Later Binding (GOOSE): Add edit button to open `GSEControl` edit wizard


JakobVogelsang:
As a user of OpenSCD I want to be able to change the `GSEControl` element from within the `Subscriber Later Binding (GOOSE)` plugin.

**Requirement:**
- add edit button to the list item that opens the edit wizard for `GSEControl` element

Issue #865, opened 2022-07-12
Subscriber Later Binding (GOOSE): Show connected and available `ExtRef` on `FCDA` selection


JakobVogelsang:
As a user of OpenSCD I want to know if an existing data point in a `GSEControl` element is connected to a `ExtRef` element.

**Requirements**
- show all connected later binding `ExtRef` for the selected `FCDA`. That means the `ExtRef` element
1. has an existing attribute `intAddress`
2. has attributes `ldInst`, `prefix`, `lnClass`, `lnInst`, `doName` and `daName` that are equal to the same once in `FCDA`
3. has an attribute `iedName` equals to the `name` attribute of the `IED` element, where `FCDA` resides

Ed.2 and higher also
5. `serviceType="GOOSE"`
6.`srcLDInst`, `srcPrefix`, `srcLNClass`, `srcLNInst` and `srcCBName` fit to the `SampledValueControl` containing the `FCDA`

- mark all `ExtRef` with existing `intAddress` that are not connected (missing `iedName`, `ldInst`, `prefix`, `lnClass`, `lnInst`, `doName`, `daName`) as unsubscribed
- exclude all `ExtRef` from the source IED (same IED as the selected `FCDA`)
- disable elements in the unsubscribed list that cannot be connected (attributes `pLN`, `pDO`, `pDA` and `pServT` are present in the only criteria for now)

**Mockups**
![with_ext_ref_list excalidraw](https://user-images.githubusercontent.com/66802940/179539485-491958ee-05f3-4c00-a923-095d3302a5b7.png)

danyill:
It may be a useful feature to also display the `desc` field on an `ExtRef` to help the user in mappings. In some cases the `intAddr` is not sufficient to clarify meaning (e.g. in a SEL relay this describe a internal bit and does not name a function directly).

JakobVogelsang:
I updated the Mockup :)

PR #945, opened 2022-08-18
feat(editors/subscriber-later-binding): Add plugin with filterable FCDA list

juancho0202:
Refactor to have a single Later Binding left panel with filterable FCDA list that may be used either for SVC or GSEControl elements

(WIP)

Missing:

- Adding Later Binding Goose tests
- Updating Later Binding Sample Value tests
- Removing unused code once the team approves the refactor


src/editors/SMVSubscriberLaterBinding.ts
```diff
@@ -1,6 +1,6 @@
 import { css, html, LitElement, property, TemplateResult } from 'lit-element';
 
-import './subscription/smv-laterbinding/svc-laterbinding-list.js';
```
dlabordus, line 3: Shouldn't we cleanup this old file, including translations?

src/editors/GooseSubscriberLaterBinding.ts
```diff
@@ -0,0 +1,41 @@
...
+      </div>
+    </div>`;
+  }
+
+  static styles = css`
+    :host {
+      width: 100vw;
+    }
+
+    .container {
```
dlabordus, line 27: I think @JakobVogelsang likes the list be the full width for this issue and when picking up the next issue to add the second list make it 50%.

So for now that means removing 'display' and 'height' from the class 'container' and also the class 'column' isn't needed for now. When picking up issue #865, these lines can be added again.

JakobVogelsang: That is very true :)

dlabordus (changes requested):

dlabordus (approved):

juancho0202:
Closes #860

PR #956, opened 2022-08-25
Processing of Column Number from Validation Messages

dlabordus:
Processing of the XSD Messages in the Validator Services added a new value 'ColumnNumber'. This new value is also displayed at the second line of the Diagnostic Dialog. Also the vale 'Linenumber' changed to 'LineNumber'.


dlabordus:
Sorry wrong repository :-(

PR #957, opened 2022-08-25
fix(wizard/gsecontrol): Fix schema validation error after creating GSE Control (MinTime)

### LINKED ISSUE #848: Creating GOOSE results in schema invalid project
**Describe the bug**
Adding new `GSEControl` element results in schema invalidity due to incorrect sequence for `MinTime` element. 

**To Reproduce**
1. Open Project
2. Create GOOSE !Make sure that host IED is connected to Communication!
3. See schema validator. It says `This element is not expected: line: ... MinTime`

**Expected behavior**
The sequence of the elements should be handled by OpenSCD when adding a new element.

**Error source**
- The bug in `scr/wizards/gsecontrol.ts` in line 191 and 192. Those have the wrong sequence. It must be another way around. Even better to use the action event and the sequence check there to do that.

dlabordus:
closes #848

### DIFF
--- a/src/wizards/gsecontrol.ts
+++ b/src/wizards/gsecontrol.ts
@@ -27,6 +27,7 @@ import {
   newSubWizardEvent,
   newWizardEvent,
   selector,
+  SimpleAction,
   Wizard,
   WizardActor,
   WizardInputElement,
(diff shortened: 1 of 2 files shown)


JakobVogelsang (approved):
LGTM!! Thank you for fixing this one!

PR #959, opened 2022-08-29
feat(wizard/dai): Set/Update a value for a type 'Timestamp' of Data Attribute Instance

dlabordus:


test/unit/wizards/dai.test.ts
```diff
@@ -4,59 +4,205 @@ import '../../mock-wizard.js';
...
+      await element.requestUpdate();
+      inputs = Array.from(element.wizardUI.inputs);
+    });
+
+    it('update value should be updated in document', async function () {
+      await setWizardTextFieldValue(<WizardTextField>inputs[0], '');
+      await setWizardTextFieldValue(<WizardTextField>inputs[1], '');
+
+      const complexActions = createValue(
+        dai.closest('DOI')!,
```
JakobVogelsang, line 96: It is defined in line 79 already. One of those should be removed I think

JakobVogelsang (commented):
One tiny thing that I found. Other than that LGTM

test/unit/wizards/dai.test.ts
dlabordus: Forgot this one when refactoring the test. Found a second one also :-)

JakobVogelsang: :)

JakobVogelsang (approved):
LGTM

Issue #958, opened 2022-08-28
Feedback on Later Binding SMV Plugin


danyill:
**Is your feature request related to a problem? Please describe.**

This is some early, highly optional feedback on the excellent SMV Later Binding plugin. I did not manage to find time to comment during the review cycle, sorry.
We used this plugin (briefly) to configure some SMV subscriptions, and here is some feedback for enhancements at some stage.

**Describe the solution you'd like**

1. Initially, the person I tried out this extension on was confused about how to use it. I think perhaps some terminology changes would help.
    a. For the publisher on the left-hand side, I'd like to suggest a terminology change. Instead of "List of Sampled Values Controls with linked FCDAs" we could have (as per the other SMV plugin): "Sampled Values Publisher".
    b. For the subscriber on the left-hand side, I'd like to suggest we rename this from, "Selected Sampled Value data "smvcb0" / FCDA "AmpSv.instMag.i"" to: "Subscriber to". I'm not sure the control block name or the FCDA is useful if item 2 below is implemented. But if they are useful, I think including the IED name would be helpful. In many cases, merging units will have the same SMV control block name.

2. The selection on the LHS becomes lost when carrying out activity on the right-hand side. In other plugins (for instance, the Publisher plugin) we don't lose this, and it has a light green background (I think this is the `activated` property). This is valuable context for the user.

3. While the later binding plugin provides good functionality for the action of binding, it doesn't readily allow an overview of the configured ExtRefs which most tools provide, see the following screenshots from other applications which provide a tabular overview for each FCDA.

    a. AcSELerator Architect
        ![image](https://user-images.githubusercontent.com/674783/187053306-bd7bb32d-2f92-4a6c-9411-62ade3174198.png)

    b. Siemens DIGSI 5
        ![image](https://user-images.githubusercontent.com/674783/187053346-755e6265-a19f-4f84-9681-09099ebf1a5c.png)

    c. We could do this through a separate plugin. Perhaps this is a useful overview for all the different editors. But if you are doing the mappings for a three-phase set of currents and voltages, for example, it is helpful to be able to see what is "already done and what is still to do".
    d. Another idea might be to allow clicking on the IED, and when this happens for all items in the SMV control block to be a synchronised list between the values on the left and right-hand screen, but this might not readily fit a mobile form factor.

JakobVogelsang:
Regarding point 3 of your feedback, I would propose to add the number of connection to the list item representing the `FCDA` element like so

[...]

PR #991, opened 2022-09-03
refactor(editors/subscriber...): Naming in GOOSE/SMV Subscriber plugins

danyill:
Closes #729 

This is in response to #729 and partially resolves #958 (slightly different terminology used than what was proposed there).

I know this is just wording and possibly doesn't seem that important :grey_question:. However, just a few words can make things more or less intuitive. I will be happy to receive a review, and if this is not helpful, I will be happy for it to be closed. :shrug: 

GOOSE/Sampled Value Control:

* Simplify the radio buttons to "Publisher | Subscriber" and "Subscriber | Publisher". They are too hard to parse and users must learn by doing in any case.
* Reword the headings for IED lists to be more explicit about what is happening in the two parts of the screen

Later Binding:

* If we are going to describe the source, we need to qualify it with not just the DA but the LDevice and LNode to be meaningful
* This is a lot of information, so an overflow should wrap to be responsive because this title can be pretty important
* Generally avoid technical language (like FCDAs and ExtRefs) and replace them with Messages and Inputs. I think this is better for users, although less precise.
* I decided not to include the IED name because I think applying the `activated` attribute to the filtered list will be adequate to provide this context (arguably removes a need for the title at all but in many ways looking up to more significant text is better than looking left)

The German is Google Translate and possibly no good; happy for it to be reverted or improved.

Here are some screenshots:

### GOOSE

**Publisher**

**Before**
image

**After**
image

Subscriber

**Before**
image

**After**
image

SMV is similar.

### SMV (Later Binding) 

**Before**
image

**After**
image

PR #992, opened 2022-09-03
Add activatable to subscriber filtered lists

danyill:
This adds "activatable" to the filtered lists used in subscribers as suggested in #958. When we click on the RHS we should not lose the selection on the LHS.

However, this exacerbates the restoration of the RHS pane issue where if either opening a new file or going to a different editor and returning the RHS pane may be either incorrect (for a new file without the same element) or the selection on the left is not shown.

I think Jakob did something in the Publisher plugin for this and I must look at this more deeply. We cannot readily save state for each editor at the moment I think such that one can flick between editors without losing selections. Does that mean the RHS panel should be made "empty" i.e. deselected whenever entering the editors?


danyill:
Will be done in #991 as requested by Jakob

PR #995, opened 2022-09-05
feat(editors/subscriber-later-binding): Show connected and available ExtRef on FCDA selection (GOOSE)

### LINKED ISSUE #865: Subscriber Later Binding (GOOSE): Show connected and available `ExtRef` on `FCDA` selection

juancho0202:
closes #865


src/editors/subscription/later-binding/ext-ref-laterbinding-list.ts
```diff
@@ -97,6 +99,39 @@ export class ExtRefLaterBindingList extends LitElement {
     );
   }
 
+  get version(): string | null {
```
dlabordus, line 102: Use a function call 'getSclSchemaVersion' to check if it's a 2003 version.

src/editors/subscription/later-binding/ext-ref-laterbinding-list.ts
```diff
@@ -97,6 +99,39 @@ export class ExtRefLaterBindingList extends LitElement {
...
+    return this.doc.documentElement.getAttribute('version');
+  }
+
+  private checkSuscribedGooseRequirements(extRefElement: Element): boolean {
```
dlabordus, line 106: I think this method should also be implemented for SampledValueControl, @JakobVogelsang?
So then only the difference is in 'serviceType', I think between 'SMW' and 'GOOSE'.

dlabordus (changes requested):

src/editors/subscription/later-binding/ext-ref-laterbinding-list.ts
JakobVogelsang: Oh yes, we should do that. I was missing that for the SampledValueControl element type subscribing when defining the issue.

juancho0202: would this be the solution?

`extRefElement.getAttribute('serviceType') === this.controlTag !== 'GSEControl'? 'GOOSE':'SMV' &&`

juancho0202: are there any other differences aside from serviceType attribute for SMV?

src/editors/subscription/later-binding/ext-ref-laterbinding-list.ts
```diff
@@ -97,6 +100,33 @@ export class ExtRefLaterBindingList extends LitElement {
     );
   }
 
+  private checkSuscribedGooseRequirements(extRefElement: Element): boolean {
```
dlabordus, line 103: Change the name of the function, because it's not anymore about Goose Only :-)

dlabordus (changes requested):

dlabordus (approved):

PR #998, opened 2022-09-06
doc(menu/help): update and user doc

JakobVogelsang:
An update of the wiki files based on what we have in the wiki to make it available in the help plugin.


dlabordus (approved):
Didn't read them all ;-), but this is nice for the users.

PR #1002, opened 2022-09-12
Subscriber Later Binding (GOOSE): Disconnect FCDA from ExtRef

juancho0202:


src/editors/subscription/later-binding/foundation.ts
```diff
@@ -24,6 +29,22 @@ export function newFcdaSelectEvent(
   });
 }
 
+export function getFCDALaterBindingList(
```
dlabordus, line 32: I don't think this is a handy place for these 2 methods. 
They are not needed in production code.
And also the dependency is now in a loop, SMVSubscriberLaterBinding.ts depends on ext-ref-laterbinding-list.ts which depends on foundation.ts and which depends again on SMVSubscriberLaterBinding.ts. Not a good idea.

I created a file test-support.ts beside the test files (test/integration/editors/) to hold this kind of functions. The first idea was to also call this file foundation.ts, but that looked a bit strange if there is also a foundation.test.ts in that directory.

test/integration/editors/GooseSubscriberLaterBinding.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+
+    element = await fixture(
+      html`<goose-subscriber-later-binding-plugin
+        .doc="${doc}"
+      ></goose-subscriber-later-binding-plugin>`
+    );
+  });
+
+  it('when unsubscribing a subscribed ExtRef then the list are changed', async () => {
```
dlabordus, line 32: 'when unsubscribing a subscribed ExtRef then the **lists** are changed.'
Small typo. Also in the other test file already wrong, so change it for all tests (4x)

dlabordus (approved):

src/editors/subscription/later-binding/foundation.ts
juancho0202: I moved them into file test/integration/editors/test-support.ts as suggested

test/integration/editors/GooseSubscriberLaterBinding.test.ts
juancho0202: Updated 3x occurrences

PR #1007, opened 2022-09-18
Add ? and * wildcards to filtered-search, closes #1006.

### LINKED ISSUE #1006: Support wildcards in filtered list for Subscriber Later Binding SMV and GOOSE

The filtered lists in OpenSCD are powerful and an integral part of the workflow.
However, they do not always facilitate an efficient workflow for particular device semantics.

Consider the following example:

image

* <1> and <2> show the internal addresses which are semantically in this relay (a SEL-400 series relay)
* <3> shows a second attachment point which is used to aggregate SV streams internally within the device.

The nomenclature is:

* `I` = Current
* `A` or `B` or `C` = Phase
* `S` or `T` or `U` = Terminal
* `_2` or `_3` is the attachment point to allow multiple SMV streams to be combined.

As an end-user I wish to select the terminal first (the terminal might be, for instance, a side of a transformer).
Because of these particular device semantics, this is at the end of the internal address, making this hard to do.
Ideally, our filtering would be powerful enough that this would not matter.

I want the following order of operations:

* `IAS`
* `IBS`
* `ICS`
* `IAS_2`
* `IBS_2`
* `ICS_2`

I want to be able to filter: `I?S` and then `I?S_2` to improve my workflow.
This saves me from having to re-apply a filter for each phase.
At the moment, I think I only need a per-character wildcard.

**Test Files**

See attached:
[XAT_Prot1.scd.zip](https://github.com/openscd/open-scd/files/9587011/XAT_Prot1.scd.zip)

danyill:
This is just a prototype.

It does allow you to do a nice thing, i.e. line up phases between left and right.

image

### DIFF
--- a/src/filtered-list.ts
+++ b/src/filtered-list.ts
@@ -38,11 +38,15 @@ function hideFiltered(item: ListItemBase, searchText: string): void {
     value
   ).toUpperCase();
 
-  const terms: string[] = searchText.toUpperCase().split(' ');
-
-  terms.some(term => !filterTarget.includes(term))
-    ? slotItem(item).classList.add('hidden')
-    : slotItem(item).classList.remove('hidden');
+  const terms: string[] = searchText.toUpperCase().trim().split(' ');
+  terms.some(term => {
+    // regexp escape 
+    const termRegexed = `*${term}*`.replace(/[.+^${}()|[\]\\]/g, '\\$&'); 
+    const reTerm = new RegExp(termRegexed.replace(/\*/g,'.*').replace(/\?/g,'.'),'i');
+    !reTerm.test(filterTarget)
+      ? slotItem(item).classList.add('hidden')
+      : slotItem(item).classList.remove('hidden');
+  });
 }
 
 /**
--- a/src/filtered-list.ts
+++ b/src/filtered-list.ts
@@ -18,6 +18,7 @@ import { List } from '@material/mwc-list';
 import { ListBase } from '@material/mwc-list/mwc-list-base';
 import { ListItemBase } from '@material/mwc-list/mwc-list-item-base';
 import { TextField } from '@material/mwc-textfield';
+import { ReportControlElementEditor } from './editors/publisher/report-control-element-editor';
 
 function slotItem(item: Element): Element {
   if (!item.closest('filtered-list') || !item.parentElement) return item;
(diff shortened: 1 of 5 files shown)


ca-d:
@danyill I was really intrigued when you just mentioned the possibility of letting users filter by regex. Would a "regex toggle button" next to the search field be a viable alternative for this use case as well?

danyill:
> @danyill I was really intrigued when you just mentioned the possibility of letting users filter by regex. Would a "regex toggle button" next to the search field be a viable alternative for this use case as well?

Yes, that is a good idea. I am tempted to say let's just clone the Developer Tools from Chrome.

image

But I wonder if regex search shouldn't be a settings toggle or a "pro mode" feature, and more basic filtering like what we have now with just the addition of wildcards might be better for most users...

JakobVogelsang:
I would propose to start with the wildcard. The implementation looks good to me. But tests are missing ;)

danyill:
In fact, the implementation was not good, alas, but I hope it is better now. Please check carefully, we worry terribly about unintended consequences and the high impact of this change!

src/filtered-list.ts
```diff
@@ -38,11 +39,23 @@ function hideFiltered(item: ListItemBase, searchText: string): void {
...
-  const terms: string[] = searchText.toUpperCase().split(' ');
-
-  terms.some(term => !filterTarget.includes(term))
-    ? slotItem(item).classList.add('hidden')
-    : slotItem(item).classList.remove('hidden');
+  const terms: string[] = searchText
+    .toUpperCase()
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&')
+    .trim()
+    .split(' ');
```
JakobVogelsang, line 46: split on `/\s/g`

src/filtered-list.ts
```diff
@@ -38,11 +39,23 @@ function hideFiltered(item: ListItemBase, searchText: string): void {
...
   ).toUpperCase();
 
-  const terms: string[] = searchText.toUpperCase().split(' ');
-
-  terms.some(term => !filterTarget.includes(term))
-    ? slotItem(item).classList.add('hidden')
-    : slotItem(item).classList.remove('hidden');
+  const terms: string[] = searchText
+    .toUpperCase()
+    .replace(/[.+^${}()|[\]\\]/g, '\\$&')
```
JakobVogelsang, line 44: Where did you get the collection of "problematic" characters from?

JakobVogelsang (dismissed):
One little change and one question other that that VERY WELL DONE!

src/filtered-list.ts
danyill: I expect them somewhere on SO, but I couldn't find them again for looking. They look like the right ones for a regex on the face of it.

src/filtered-list.ts
JakobVogelsang: I know thank you :)

danyill:
Thank you for this review. I have made a little change to use a regex for multiple spaces for the split.

JakobVogelsang (approved):
LGTM

PR #1010, opened 2022-09-22
feat(editors/subscription): add subscription supervision support

### LINKED ISSUE #942: Add `LGOS` and `LSVS` acc. to IEC 61850-6 and IEC 61850-7-1
As a user of OpenSCD I want the tool to handle the `LGOS` and `LSVS` behavior acc. to the IEC 61850-6 2007B4. that is

**Background**
It is very easy for the tester of IEC 61850 communication to test if a GOOSE or Sampled Value Stream is published on the network as expected, as both are multicast messages that can be received on any port and compared to its setting in the SCD file. Testing of supervision is not so easy. With edition 1 it was a bad trial and error process until the IED did use the signal and the tester could see either measured values or some binary information triggering something....for that reason the logical node classes were introduced with edition 2: `LGOS` and `LSVS`. They allow per GOOSE and Sampled Value instance to supervise the subscription status. 

As all logical nodes `LGOS` and `LSVS` are allocated in the data model. In this case, in the receiving IED. With it the question arise who is responsible to put the data into the data model. With Ed2 this question was not answered fully and this lead to the discussion in [Tissue 1401](https://iec61850.tissue-db.com/tissue/1401). With the newest part of the 61850-6 this is clarified to an extent that we can be confident enough to implement it. See Table 11 in IEC 61850-6 for `SupSubscription` element for the wording of the -6.

**Requirements**
- only relevant for >= Ed2(2007B)
- Triggered on each newly subscribed GOOSE/SampledValue: no information from the GOOSE/SampledValue has been subscribed in the subscriber IED yet (query `ExtRef[iedName="sendingIED"][srcCBName="name of GOOSE/SampledValue"]` in the subscribing IED returns null)
- first try to add the control block reference to an existing but not used `LGOS`/`LSVS`
- if this is not possible, try to add new `LGOS`/`LSVS` to the subscribing IED
- warn the user if new instance of `LN[lnClass="LSVS"]` exceed the `maxSv` number

>NOTE: There are two possible scenarios that can happen and the requirements differ a bit: 1. There are already pre-defined `LGOS`/`LSVS` in the data model, and we need to instantiate the `DOI[name="GoCBRef"]>DAI[name="setSrcRef"]` or `DOI[name="SvCBRef"]>DAI[name="setSrcRef"]`, respectively. 2. There are no instantiated `LGOS`/`LSVS` and we need to instantiated new once in order to set the `DOI[name="GoCBRef"]>DAI[name="setSrcRef"]` or `DOI[name="SvCBRef"]>DAI[name="setSrcRef"]`, respectively. 

**Configure existing `LGOS`/`LSVS`**
1. check if the subscribing IED does have an unused instance of `LGOS`/`LSVS`
- get the first element `LN[lnClass="LGOS"]`/`LN[lnClass="LSVS"]` that does not have
- `DOI[name="GoCBRef"]>DAI[name="setSrcRef"]`/`DOI[name="SvCBRef"]>DAI[name="setSrcRef"]`
2. Add the GOOSE control block reference / the Sampled Values control block reference (e.g. `AA1E1Q01MUMod2_MU1/LLN0.MSVCB01`) as a value to the `DAI[name="setSrcRef"]`
>NOTE: the control block reference is a pointer to the `GSEControl`/`SampledValueControl` element that is structured as follows IEDNameLDInst/LLN0.CBName. (the `name` attribute of the receiving IED followed by the `inst` attribute of the parent LDevice follow by `/` follow by `LLN0` followed by `.` followed by the `name` attribute of the GOOSE/SampledValue control block)
3. Add a child `<Private type="OpenSCD.create"/>` to the `LGOS/LSVS` element

**First instantiate than configure `LGOS`/`LSVS`**
1. check all the pre-condition for adding LGOS/LSVS further down
2. if all conditions are met, add a new `LN` element into the logical device `LDevice` where all the other `LN[lnClass="LGOS"]`/`LN[lnClass="LSVS"]`
>NOTE: One pre-condition is that at least one such logical node exist already in the data model

**Pre-Condition to add new LN instance of LGOS/LSVS**
1. Check if `LGOS` instance can be added to the subscriber IED (all checks are related to the subscriber IED!)
- Check in the `Services` element - child of `IED` or `AccessPoint` if `Services>ClientServices>SubSubscription.maxGo` is bigger than 0 
- Check if the subscriber IED has instantiated logical node `LN[lnClass="LGOS"]`
- Check if the `lnType` attribute of instantiated logical node has a representation in the data type templates section `DataTypeTeplates>LNodeType[id="${lnType}"]`
- check if the number of the already instantiated `LGOS` is smaller than the allowed subscription for the subscriber IED: `querySelectorAll(LN[lnClass="LGOS"]).lenght < Services>ClientServices>SubSubscription.maxGo`
- check if the IED allows instantiating values `DataTypeTemplates>LNodeType[lnClass="LGOS"]>DO[name="GoCBRef"]>DA[name="setSrcRef"]` has attributes `valKind=Conf/RO` and `valImport=true`. 
> NOTE: `valImport` defines if instantiated data (instantiated by SCT) will change the data model when re-importing into the ICT. This is the case here. We want instantiate the `setSrcRef` data attribute and wan the IED to re-import and use it

2. Check if `LSVS` instance can be added to the subscriber IED (all checks are related to the subscriber IED!)
- Check in the `Services` element - child of `IED` or `AccessPoint` if `Services>ClientServices>SubSubscription.maxSv` is bigger than 0 
- Check if the subscriber IED has instantiated logical node `LN[lnClass="LSVS"]`
- Check if the `lnType` attribute of instantiated logical node has a representation in the data type templates section `DataTypeTeplates>LNodeType[id="${lnType}"]`
- check if the number of the already instantiated `LSVS` is smaller than the allowed subscription for the subscriber IED: `querySelectorAll(LN[lnClass="LSVS"]).lenght < Services>ClientServices>SubSubscription.maxSv`
- check if the IED allows instantiating values `DataTypeTemplates>LNodeType[lnClass="LSVS"]>DO[name="SvCBRef"]>DA[name="setSrcRef"]` has attributes `valKind=Conf/RO` and `valImport=true`. 
> NOTE: `valImport` defines if instantiated data (instantiated by SCT) will change the data model when re-importing into the ICT. This is the case here. We want instantiate the `setSrcRef` data attribute and wan the IED to re-import and use it

>NOTE: see the IEC 61850-7-1 Annex G: LGOS/LSVS engineering. This a very nice overview of the several checks we need to implement to make sure we are conflicting with the ICT (IEC configuration tool)

**Additional information**
- see the [tissue database](https://iec61850.tissue-db.com/default.mspx) and search for `LGOS`
- see IEC 61850-6 Table 11
- see IEC 61850-7-1 Annex G

juancho0202:
closes #942

### DIFF
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -4,6 +4,7 @@ import {
   compareNames,
   Create,
   createElement,
+  Delete,
   getSclSchemaVersion,
 } from '../../foundation.js';
 import {
@@ -228,6 +229,48 @@ export function instantiateSubscriptionSupervision(
     },
   ]);
 }
+
+/**
+ * Return an array with a single Delete action to delete the supervision element
+ * for the given GOOSE/SMV message and subscriber IED.
+ */
+export function removeSubscriptionSupervision(
+  controlBlock: Element | undefined,
+  subscriberIED: Element
+): Delete[] {
+  const supervisionType =
+    controlBlock?.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  const valElement = Array.from(
+    subscriberIED.querySelectorAll(
+      `LN[lnClass="${supervisionType}"]>DOI>DAI>Val,LN0[lnClass="${supervisionType}"]>DOI>DAI>Val`
+    )
+  ).find(val => val.textContent == controlBlockReference(controlBlock));
+  if (!valElement) return [];
+  const lnElement = valElement.closest('DOI')?.parentElement;
+  if (!lnElement || !lnElement.parentElement) return [];
+  // Check if that one has been created by OpenSCD (private section exists)
+  const isOpenScdCreated = lnElement.querySelector(
+    'Private[type="OpenSCD.create"]'
+  );
+  return isOpenScdCreated
+    ? [
+        {
+          old: {
+            parent: lnElement.parentElement,
+            element: lnElement,
+          },
+        },
+      ]
+    : [
+        {
+          old: {
+            parent: lnElement,
+            element: valElement.closest('DOI')!,
+          },
+        },
+      ];
+}
+
 /**
  * Checks if the given combination of GOOSE/SMV message and subscriber IED allows for subscription supervision.
  * @param controlBlock The GOOSE or SMV message element
@@ -279,6 +322,12 @@ export function findOrCreateAvailableLNInst(
     SCL_NAMESPACE,
     'LN'
   );
+  const openScdTag = subscriberIED.ownerDocument.createElementNS(
+    SCL_NAMESPACE,
+    'Private'
+  );
+  openScdTag.setAttribute('type', 'OpenSCD.create');
+  newElement.appendChild(openScdTag);
   newElement.setAttribute('lnClass', supervisionType);
   const instantiatedSibling = subscriberIED
     .querySelector(`LN[lnClass="${supervisionType}"]>DOI>DAI>Val`)
@@ -342,10 +391,10 @@ export function controlBlockReference(
   controlBlock: Element | undefined
 ): string | null {
   if (!controlBlock) return null;
-  const anyLN = controlBlock.closest('LN,LN0');
-  const prefix = anyLN?.getAttribute('prefix') ?? '';
-  const lnClass = anyLN?.getAttribute('lnClass');
-  const lnInst = anyLN?.getAttribute('inst') ?? '';
+  const eitherLN = controlBlock.closest('LN,LN0');
+  const prefix = eitherLN?.getAttribute('prefix') ?? '';
+  const lnClass = eitherLN?.getAttribute('lnClass');
+  const lnInst = eitherLN?.getAttribute('inst') ?? '';
   const ldInst = controlBlock.closest('LDevice')?.getAttribute('inst');
   const iedName = controlBlock.closest('IED')?.getAttribute('name');
   const cbName = controlBlock.getAttribute('name');
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -196,7 +196,7 @@ export function instantiateSubscriptionSupervision(
   if (!availableLN) return [];
   // First, create the templateStructure array
   const templateStructure = createTemplateStructure(availableLN, [
-    'SvCBRef',
+    controlBlock?.tagName === 'GSEControl' ? 'GoCBRef' : 'SvCBRef',
     'setSrcRef',
   ]);
   if (!templateStructure) return [];
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -194,7 +194,20 @@ export function instantiateSubscriptionSupervision(
     supervisionType
   );
   if (!availableLN) return [];
-  // First, create the templateStructure array
+  // First, we make suere that LN's inst is unique and non-empty
+  const inst = availableLN.getAttribute('inst') ?? '';
+  if (inst === '') {
+    const instNumber =
+      Array.from(
+        subscriberIED.querySelectorAll(`LN[lnClass="${supervisionType}"]`)
+      ).reduce((highestVal, ln) => {
+        return ln.getAttribute('inst')
+          ? Math.max(highestVal, +ln.getAttribute('inst')!)
+          : highestVal;
+      }, 0) + 1;
+    availableLN.setAttribute('inst', '' + instNumber);
+  }
+  // Then, create the templateStructure array
   const templateStructure = createTemplateStructure(availableLN, [
     controlBlock?.tagName === 'GSEControl' ? 'GoCBRef' : 'SvCBRef',
     'setSrcRef',
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -2,9 +2,15 @@ import { css, LitElement, query } from 'lit-element';
 import {
   cloneElement,
   compareNames,
+  Create,
   createElement,
   getSclSchemaVersion,
 } from '../../foundation.js';
+import {
+  createTemplateStructure,
+  determineUninitializedStructure,
+  initializeElements,
+} from '../../foundation/dai.js';
 import { getFcdaReferences } from '../../foundation/ied.js';
 
 export enum View {
@@ -158,6 +164,110 @@ export function getExtRef(
       `ExtRef[iedName="${iedName}"]${getFcdaReferences(fcda)}${controlCriteria}`
     )
   ).find(extRefElement => !extRefElement.hasAttribute('intAddr'));
+}
+
+export function instantiateSubscriptionSupervision(
+  controlBlock: Element,
+  subscriberIED: Element
+): Create[] {
+  if (getSclSchemaVersion(subscriberIED.ownerDocument) === '2003') return [];
+  const supervisionType =
+    controlBlock.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  if (subscriberIED.querySelector(`LN[lnClass="${supervisionType}"]`) === null)
+    return [];
+  if (
+    Array.from(
+      subscriberIED.querySelectorAll(
+        `LN[lnClass="${supervisionType}"]>DOI>DAI>Val`
+      )
+    ).find(val => val.textContent == controlBlockReference(controlBlock))
+  )
+    return [];
+  if (
+    maxSupervisions(subscriberIED, controlBlock) <=
+    instantiatedSupervisions(subscriberIED, controlBlock)
+  )
+    return [];
+  const availableLN = nextAvailableLNInst(subscriberIED, controlBlock);
+  if (!availableLN) return [];
+  // First, create the templateStructure array
+  const templateStructure = createTemplateStructure(availableLN, [
+    'SvCBRef',
+    'setSrcRef',
+  ]);
+  if (!templateStructure) return [];
+  // Determine where to start creating new elements (DOI/SDI/DAI)
+  const [parentElement, uninitializedTemplateStructure] =
+    determineUninitializedStructure(availableLN, templateStructure);
+  // // Next create all missing elements (DOI/SDI/DAI)
+  const newElement = initializeElements(uninitializedTemplateStructure);
+  newElement.querySelector('Val')!.textContent =
+    controlBlockReference(controlBlock);
+  return [
+    {
+      new: {
+        parent: parentElement,
+        element: newElement,
+      },
+    },
+  ];
+}
+
+export function nextAvailableLNInst(
+  subscriberIED: Element,
+  controlBlock: Element
+): Element | null {
+  const supervisionType =
+    controlBlock.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+
+  const firstEmptyLN = Array.from(
+    subscriberIED.querySelectorAll(`LN[lnClass="${supervisionType}"]`)
+  ).find(
+    ln =>
+      ln.querySelector('DOI>DAI>Val') === null ||
+      ln.querySelector('DOI>DAI>Val')?.textContent === ''
+  );
+  if (firstEmptyLN) return firstEmptyLN;
+  return null;
+}
+
+export function instantiatedSupervisions(
+  subscriberIED: Element,
+  controlBlock: Element
+): number {
+  const supervisionType =
+    controlBlock.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  const instantiatedValues = Array.from(
+    subscriberIED.querySelectorAll(
+      `LN[lnClass="${supervisionType}"]>DOI>DAI>Val`
+    )
+  ).filter(val => val.textContent !== '');
+  return instantiatedValues.length;
+}
+
+export function maxSupervisions(
+  subscriberIED: Element,
+  controlBlock: Element
+): number {
+  const maxAttr = controlBlock.tagName === 'GSEControl' ? 'maxGo' : 'maxSv';
+  const maxValues = parseInt(
+    subscriberIED
+      .querySelector('Services>ClientServices>SupSubscription')
+      ?.getAttribute(maxAttr) ?? '0'
+  );
+  return isNaN(maxValues) ? 0 : maxValues;
+}
+
+export function controlBlockReference(controlBlock: Element): string | null {
+  const anyLN = controlBlock.closest('LN,LN0');
+  const prefix = anyLN?.getAttribute('prefix') ?? '';
+  const lnClass = anyLN?.getAttribute('lnClass');
+  const lnInst = anyLN?.getAttribute('inst') ?? '';
+  const ldInst = controlBlock.closest('LD')?.getAttribute('inst');
+  const iedName = controlBlock.closest('IED')?.getAttribute('name');
+  const cbName = controlBlock.getAttribute('name');
+  if (!cbName && !iedName && !ldInst && !lnClass) return null;
+  return `${iedName}${ldInst}/${prefix}${lnClass}${lnInst}.${cbName}`;
 }
 
 export function canCreateValidExtRef(
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -234,6 +234,12 @@ export function instantiateSubscriptionSupervision(
 /**
  * Return an array with a single Delete action to delete the supervision element
  * for the given GOOSE/SMV message and subscriber IED.
+ *
+ * @param controlBlock The GOOSE or SMV message element
+ * @param subscriberIED The subscriber IED
+ * @returns an empty array if removing the supervision is not possible or an array
+ * with a single Delete action that removes the LN if it was created in OpenSCD
+ * or only the supervision structure DOI/DAI/Val if it was created by the user.
  */
 export function removeSubscriptionSupervision(
   controlBlock: Element | undefined,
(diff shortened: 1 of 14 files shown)


src/foundation/dai.ts
```diff
@@ -1,4 +1,5 @@
 import { SCL_NAMESPACE } from '../schemas.js';
+import { getTypeAttribute } from '../editors/protocol104/foundation/foundation.js';
```
dlabordus, line 2: I don't think it's handy to use the one from the 104 Protocol Editor.
I think there is also one in the foundation.js of the root.
If not just use .getAttribute('type')

src/foundation/dai.ts
```diff
@@ -99,3 +100,57 @@ export function initializeElements(
...
+  path.forEach(name => {
+    // There should be a DOType or DAType set for the current element in the list.
+    if (!typeElement) {
+      templateStructure = null;
+      return;
+    }
+    const dataElement = typeElement.querySelector(
+      `:scope > DO[name="${name}"], :scope > SDO[name="${name}"], :scope > DA[name="${name}"], :scope > BDA[name="${name}"]`
+    );
+    // If there is no DA/BDA Element found the structure is incorrect, so just stop.
```
dlabordus, line 131: Maybe also update this comments.

juancho0202:
@dlabordus I would like to meet to ask you guidance on how to approach the later-binding implementation. Currently the instantiateSubscriptionSupervision and removeSubscriptionSupervision return Create actions and Delete actions respectively, in later-binding I see we're using Replace actions for both subscription and unsubscription, I don't understand right now how to make those thing work together.

Also I appreciate any feedback about my test file `supervision.test.ts`, thanks.

juancho0202:
@JakobVogelsang I implemented the removeSubscriptionSupervision function now. I appreciate the feedback you have on how I approached it and of course in my code overall. Thanks!

JakobVogelsang:
@juancho0202 Another thing I have found: subscription from `DATA BINDING` does not add any supervision information to the sink IED

juancho0202:
Warn the user if new instance of LN[lnClass="LSVS/LGOS"] exceeds the maxSv/maxGo permitted

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+        `LN[lnClass="${supervisionType}"]>DOI>DAI>Val`
+      )
+    ).find(val => val.textContent == controlBlockReference(controlBlock))
+  )
+    return false;
+  if (
+    maxSupervisions(subscriberIED, controlBlock) <=
+    instantiatedSupervisionsCount(subscriberIED, controlBlock, supervisionType)
+  )
+    return false;
```
JakobVogelsang, line 302: Some empty lines inbetween would make the code read better. Here is a good example for it. It becomes hard to distinguish between return in line 302 and the one in the next line

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+    SCL_NAMESPACE,
+    'LN'
+  );
+  const openScdTag = subscriberIED.ownerDocument.createElementNS(
+    SCL_NAMESPACE,
+    'Private'
+  );
+  openScdTag.setAttribute('type', 'OpenSCD.create');
+  newElement.appendChild(openScdTag);
+  newElement.setAttribute('lnClass', supervisionType);
```
JakobVogelsang, line 334: Here I would add also the minimum inst as the minimum of all available in the IED

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+ * @param subscriberIED The subscriber IED
+ * @param controlBlock The GOOSE or SMV message element
+ * @returns The max number of LN instances with supervision allowed
+ */
+export function maxSupervisions(
+  subscriberIED: Element,
+  controlBlock: Element
+): number {
+  const maxAttr = controlBlock.tagName === 'GSEControl' ? 'maxGo' : 'maxSv';
+  const maxValues = parseInt(
```
JakobVogelsang, line 379: I think it is recommended to give it a basis e.g. `parseInt('someting',10)`.

JakobVogelsang: The return of `parseInt` is always a number. `0` for strings that are not numbers

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+ * Creates a string pointer to the control block element.
+ *
+ * @param controlBlock The GOOSE or SMV message element
+ * @returns null if the control block is undefined or a string pointer to the control block element
+ */
+export function controlBlockReference(
+  controlBlock: Element | undefined
+): string | null {
+  if (!controlBlock) return null;
+  const eitherLN = controlBlock.closest('LN,LN0');
```
JakobVogelsang, line 397: in schema this is called `anyLn` :) we use it already in some places to say `LN` or `LN0`. I would recommend to use this.

src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
```diff
@@ -115,27 +115,48 @@ export class ExtRefLaterBindingList extends LitElement {
...
     });
 
-    return {
+    const replaceAction = {
       old: { element: extRefElement },
       new: { element: clonedExtRefElement },
     };
+
+    const subscriberIed = extRefElement.closest('IED') || undefined;
+    const removeSubscriptionActions = removeSubscriptionSupervision(
```
JakobVogelsang, line 124: Does this mean we remove the supervision on any unsubscribe? If so, then it is incorrect. Imagine the following situation:

You have two subscriptions to two data attributes from one `GOOSE` and for this GOOSE you have a subscription supervision. You unsubscribe one of the two data attributes you would lose the subscription supervision, although there is still one data attribute connected.

src/editors/subscription/later-binding/ext-ref-ln-binding-list.ts
```diff
@@ -168,6 +179,15 @@ export class ExtRefLnBindingList extends LitElement {
     // Check if empty Input Element should also be removed.
     actions.push(...emptyInputsDeleteActions(actions));
 
+    // we need to extend the actions array with the actions for removing the supervision
+    const subscriberIed = lnElement.closest('IED') || undefined;
+    actions.push(
+      ...removeSubscriptionSupervision(
```
JakobVogelsang, line 185: Same as for the Later binding plugin. We must remove the subscription supervision only if any data attribute from a given GOOSE or Sample Value Stream in unsubscribed.

src/translations/de.ts
```diff
@@ -359,6 +359,8 @@ export const de: Translations = {
         publisherView: 'Publisher | Subscriber',
         subscriberView: 'Subscriber | Publisher',
       },
+      subscribeActionMessage: '???',
```
JakobVogelsang, line 362: Why not reuse the generic translation `subscription.connect` and `subscription.disconnect` here. The english terms at least are the same

test/unit/editors/subscription/supervision.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  describe('when subscribing to a SampledValueControl', () => {
+    beforeEach(async () => {
+      doc = await fetch(
+        '/test/testfiles/editors/VS893-LaterBindingSMV-LSVS.scd'
+      )
+        .then(response => response.text())
+        .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+    });
+
+    it('produces the correct structure where supervision is allowed', () => {
```
JakobVogelsang, line 24: That is a long test :D. I would break it down to smaller and more readable tests.

test/unit/editors/subscription/supervision.test.ts
```diff
@@ -0,0 +1,195 @@
+import { expect } from '@open-wc/testing';
+
+import {
+  instantiateSubscriptionSupervision,
+  removeSubscriptionSupervision,
+} from '../../../../src/editors/subscription/foundation.js';
+
+import { Create, Delete } from '../../../../src/foundation.js';
+
+describe('supervision', () => {
```
JakobVogelsang, line 10: There are tests missing for the following case: You have two data attributes subscribed from one e.g. GOOSE and you unsubscribe one of those. In this case no removal of the subscription supervision should be done.

src/foundation.ts
```diff
@@ -2797,6 +2797,18 @@ export function newLnInstGenerator(
   };
 }
 
+/**
+ * @param lnElements - The LN elements to be scanned for `inst`
+ * values already in use.
+ * @returns first available inst value for LN or undefined if no inst is available
+ */
+export function minAvailableLnInst(lnElements: Element[]): string | undefined {
```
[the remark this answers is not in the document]
juancho0202: Renamed ☑️

src/foundation.ts
```diff
@@ -2797,6 +2797,18 @@ export function newLnInstGenerator(
...
+  const lnInsts = new Set(lnElements.map(ln => ln.getAttribute('inst') || ''));
+  const uniqueLnInst = lnInstRange.find(lnInst => !lnInsts.has(lnInst));
+  if (uniqueLnInst) lnInsts.add(uniqueLnInst);
```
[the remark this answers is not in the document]
juancho0202: Updated to return the without adding anything to lnInsts ☑️

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+  if (!controlBlock || !subscriberIED) return [];
+  const supervisionType =
+    controlBlock?.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  const valElement = Array.from(
+    subscriberIED.querySelectorAll(
+      `LN[lnClass="${supervisionType}"]>DOI>DAI>Val,LN0[lnClass="${supervisionType}"]>DOI>DAI>Val`
+    )
+  ).find(val => val.textContent == controlBlockReference(controlBlock));
+  if (!valElement) return [];
+  const lnElement = valElement.closest('DOI')?.parentElement;
```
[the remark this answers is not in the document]
juancho0202: I wasn't sure the parent's value was either LN0 or LN exclusively at that moment. I modified to match with your suggestion `valElement.closest('LN0, LN')` ☑️

src/editors/subscription/foundation.ts
```diff
@@ -150,6 +159,252 @@ export function getExtRef(
...
+          },
+        },
+      ];
+}
+
+/**
+ * Checks if the given combination of GOOSE/SMV message and subscriber IED allows for subscription supervision.
+ * @param controlBlock The GOOSE or SMV message element
+ * @param subscriberIED The subscriber IED
+ */
```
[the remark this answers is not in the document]
juancho0202: Output defined and updated typedoc comments for all my other functions created for this story ☑️

src/editors/subscription/foundation.ts
juancho0202: Added empty line ☑️

src/editors/subscription/foundation.ts
juancho0202: Updated the comment accordingly ☑️

src/editors/subscription/foundation.ts
juancho0202: Updated name ☑️

danyill:
Thank you for your patience with this PR.

I have not looked at the code (at all) but have finally done a little user testing with an ICT. The test was:

* configure GOOSE and SMV mappings between a Schweitzer SEL-411L-2 and an NR PCS-221S relay using OpenSCD
* open in SEL Architect to see if they are recognised
* since they are not, modify the XML by hand to the point where SEL Architect will recognise the supervision LNs

The results are... not great and very likely what Jakob expected when we started this.

* `goCBRef` is insufficient to instantiate the supervision and SEL Architect does not recognise it.
* Once I add all the fields in the LNodeType that seem to be required, the file will still not save without errors relating to the `esel` namespace.
* Deleting the OpenSCD (hand-modified) supervision and recreating it in SEL Architect appears to work but the file remains with errors.
* Basically, SEL Architect does not support an SCT creating supervision nodes.
* Even if we were to make the supervision pluggable the requirements of SEL Architect are too expansive and relate to their own namespace so I think that changes must be made within the ICT before it is feasible to consider this.

For OpenSCD and this PR (for discussion):

* The supervision instance didn't appear to be created for the LSVS LN. I have not investigated why.
* These results perhaps indicate that we should disable supervision by default or find a whitelisting/blacklisting approach based on platform/software/devices tests.
* I would like to go back to SEL with some comments and see if they are willing to improve SEL Architect.
* My files used for testing and a write-up are attached: [LGOS_LSVS_SEL_and_NR_v0.01.zip](https://github.com/openscd/open-scd/files/9894595/LGOS_LSVS_SEL_and_NR_v0.01.zip). 
* For the write-up, extract the zip file and open `LGOS_LSVS_SEL_and_NR_v0.01.html`. @JakobVogelsang could you see if you agree. If so I'll take this up with SEL. For our engineering workflow we could do the supervision as a final step in an ICT but I would prefer this to be in the domain of the SCT.

I propose to do a test like this for Siemens, NR and GE in the next few weeks, but it will take me some time... Hopefully this is helpful in the meantime. Has anyone done a test like this with Siemens or other IEDs?

JakobVogelsang:
Good evening @danyill, THANK you very much for this detailed test! I love the level of detail you are adding. 

Before you approach SEL to allow this feature, and you should, one more detail. Acc. to the standard, there as a way vendors can express whether instantiated data is processed. This can be done through `valKind` and `valImport` attributes. See an example how this is done in SIEMENS

```xml
<DOType id="SIPROTEC5_DOType_ORG_external_V09.30.01_V09.30.01" cdc="ORG">
   <DA name="setSrcRef" bType="ObjRef" valKind="RO" dchg="true" fc="SP" valImport="true"/>
</DOType>
```
For this feature to work, `valImport` must be true and `valKind` either `RO` or `Conf`. 
In the fields you provided, this looks like:
```xml
<DOType  id="ORG_0_2" cdc="ORG">
   <DA name="setSrcRef" bType="ObjRef" dchg="true" fc="SP"/>
</DOType>
```

This means however that we cannot expect the vendor to accept this. SEL clearly expressed that this value instantiation is read back into the ICT.

I did test with DIGSI 9.30 and at least the update of `LGOS` and `LSVS` seems to work. We will do an online test in the next weeks to prove. I am confident however that is works.

I would propose therefore to still try to not overcomplicate the feature as when we did our work properly in the SEL there will not be any subscription supervision added.

@juancho0202 please forgive me this! It is in the requirements of this issue but I did not test yet nor have added it to the test files. So as a reminder this is the requirement

Check if the IED allows instantiating values `DataTypeTemplates>LNodeType[lnClass="LGOS"]>DO[name="GoCBRef"]>DA[name="setSrcRef"]` has attributes `valKind=Conf/RO` and `valImport=true`.
```

**Comment by danyill** [comment]
Somewhat incidentally, but perhaps in our upcoming supervision editor, we could add to the story a way of ensuring that the user is not confused about why some IEDs will provide supervision and others will not. 

IED manufacturers do not declare this, and I think we could provide an interface which shows the "calculated capability" and usage of LGOS and LSVS LNs.

File: src/editors/subscription/foundation.ts (hunk 5 above)
Comment (juancho0202) [reply src/editors/subscription/foundation.ts:394]: Updated the parseInt accordingly ☑️

File: src/translations/de.ts (hunk 9 above)
Comment (juancho0202) [reply src/translations/de.ts]: Removed `subscribeActionMessage` and reused `subscription.connect` name ☑️

File: src/editors/subscription/later-binding/ext-ref-later-binding-list.ts (hunk 7 above)
Comment (juancho0202) [reply src/editors/subscription/later-binding/ext-ref-later-binding-list.ts:124]: I am uncertain about how to implement this logic because at any given point we only know the list of subscribed ExtRefs for a particular selected FCDA in a ControlBlock. It could be that there is a subscription for a data attribute in a different FCDA element within the same ControlBlock, do I need to consult every FCDA element in the ControlBlock for its subscriptions and see if my potential removal of supervision would affect them?

File: src/editors/subscription/later-binding/ext-ref-ln-binding-list.ts (hunk 8 above)
Comment (juancho0202) [reply src/editors/subscription/later-binding/ext-ref-ln-binding-list.ts]: Verification added with fn `canRemoveSubscriptionSupervision` ☑️

File: test/unit/editors/subscription/supervision.test.ts (hunk 10 above)
Comment (juancho0202) [reply test/unit/editors/subscription/supervision.test.ts]: I changed this to `describe()` and broke down everything into smaller tests as suggested ☑️

File: test/unit/editors/subscription/supervision.test.ts (hunk 11 above)
Comment (juancho0202) [reply test/unit/editors/subscription/supervision.test.ts:10]: I added this test for both LGOS and LSVS ☑️

File: src/foundation/dai.ts [hunk 16]
```diff
@@ -99,3 +99,57 @@ export function initializeElements(
...
+/**
+ * Use the path configuration of a Common Data Class to search for all DO/SDO/DA/BDA Elements to create
+ * a structure for which DOI/SDI/DAI Elements should be created later. Null will be returned when an invalid
+ * Template Structure is described by the path.
+ *
+ * @param lnElement - The LN Element to use for searching the starting DO Element.
+ * @param path    - The (S)DO/(B)DA Elements to find in the template structure.
+ * @returns List of Elements starting with the DO Element followed by one or more SDO or (B)DA Elements describing the structure.
+ */
+export function createTemplateStructure(
```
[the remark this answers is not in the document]
juancho0202: Tests added for both LSVS and LGOS cases ☑️

src/foundation.ts
```diff
@@ -2797,6 +2797,18 @@ export function newLnInstGenerator(
...
+  return uniqueLnInst;
```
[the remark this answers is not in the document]
juancho0202: Tests for getting the min inst number added ☑️

JakobVogelsang (approved):
LGTM. I have by hand tested the last two open issues and i works. From my point of view it can be merged

PR #1012, opened 2022-09-23
fix(menu/importieds): allow import to new projects

### LINKED ISSUE #1011: Importing IED in new project fails
**Describe the bug**
Adding new IED in a project without `DataTypeTemplates` element does not work

**To Reproduce**
1. Create new project
2. Add some IED with the Import IED plugin
3. No IED is imported

**Additional context**
This bug was introduced with the performance improvement and synchronious loading of plugins. The import IED plugin does check whether `DataTypeTemplates` element is there. In this case a new actionElement is fired and the doc is updated. The doc the plugin is working with has then no documentElement anymore :(

JakobVogelsang:
Closes #1011 

See description of the bug in the issue.

### DIFF
--- a/src/menu/ImportIEDs.ts
+++ b/src/menu/ImportIEDs.ts
@@ -178,10 +178,10 @@ function hasConnectionToIed(type: Element, ied: Element): boolean {
 function addEnumType(
   ied: Element,
   enumType: Element,
-  doc: Document
+  parent: Element
 ): SimpleAction | undefined {
-  const existEnumType = doc.querySelector(
-    `:root > DataTypeTemplates > EnumType[id="${enumType.getAttribute('id')}"]`
+  const existEnumType = parent.querySelector(
+    `EnumType[id="${enumType.getAttribute('id')}"]`
   );
 
   if (existEnumType && enumType.isEqualNode(existEnumType)) return;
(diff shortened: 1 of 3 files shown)


src/menu/ImportIEDs.ts
```diff
@@ -319,21 +319,42 @@ function addLNodeType(
 function addDataTypeTemplates(ied: Element, doc: XMLDocument): SimpleAction[] {
   const actions: (SimpleAction | undefined)[] = [];
 
+  const dataTypeTemplates = doc.querySelector(':root > DataTypeTemplates')
+    ? doc.querySelector(':root > DataTypeTemplates')!
+    : createElement(doc, 'DataTypeTemplates', {});
+
+  if (!dataTypeTemplates) {
```
dlabordus, line 326: There will always be a Element now, because you will query it of create it at line 322.

dlabordus (changes requested):

src/menu/ImportIEDs.ts
```diff
@@ -178,9 +178,9 @@ function hasConnectionToIed(type: Element, ied: Element): boolean {
 function addEnumType(
   ied: Element,
   enumType: Element,
-  doc: Document
+  parent: Element
```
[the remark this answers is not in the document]
JakobVogelsang: Very true! Thanks!

src/menu/ImportIEDs.ts
JakobVogelsang: Very true! Thx

dlabordus (approved):

PR #1015, opened 2022-09-24
Document extensions and scroll behaviour on wiki, closes #557

### LINKED ISSUE #557: Tab bar scrolling
The menu/tab bar is now large enough (a sign of success) that on some form factor devices or small screens there can be too many entries.

image

We could solve this by showing left and right buttons and allowing scrolling of the tabs (especially wheel mouse scrolling which is what I'd expect to be available).

danyill:
Hopefully, this is about right. I would be pleased to take a review.


JakobVogelsang (approved):
Fantastic!! Thank for this proactive move

PR #1019, opened 2022-09-25
feat(wizard/connectedap): create GSE and SMV elements

### LINKED ISSUE #1013: Missing `SMV` and `GSE` element in `ConnectedAP` creation
**Describe the bug**
Creating new `ConnectedAP` in the Communication does not add `SMV` and `GSE` elements. 

**To Reproduce**
1. Import IED with GOOSE or SMV
2. Create new `Subnetwork` and `ConnectedAP`
3. See not `SMV` and `GSE` elements are created

**Expected behavior**
These elements are needed in interaction with other tools to check, if the GOOSE/SMV are connected to the same network.

**Requirements**
- check if the element is not yet added to another access point of the same IED (see `serverAt` feature)
- make sure that the MAC address is unique 
- make sure that `APPID` is unique
- VLANID is `000`
- VLAN-Priority is `7`
- MinTime is 10
- MaxTime is 10000

JakobVogelsang:
Closes #1013 

This PR shall add `GSE` and `SMV` elements on creating a new `ConnectedAP`. This is needed as a preparation for the subscribing plugins. The logic here is
- for each `GSEControl` add `GSE` element with unique `MAC-Address` and `APPID`
- for each `SampledValueControl` add `SMV` element with unique `MAC-Address` and `APPID`
- do this for access points that has a `Server` element and `ServerAt` element
- first come first serve ... when an IED has two access points fisrt has the server the second pointing to the server in the first and the second is added to subnetwork all GSE and SMV is added in this ConnectedAP. When after that the first is added to the same or another subnetwork no new GSE or SMV element is added

I did want to prevent UI components and with that have added two new generator functions to the foundation:
- MAC-Address
- APPID

they can be re-used to generate unique MAC-Addresses and APPIDs within the valid ranges of both.


src/wizards/connectedap.ts
```diff
@@ -46,6 +51,170 @@ function compareAccessPointConnection(
...
+        multiplier: 'm',
+      });
+      minTime.textContent = '10';
+      actions.push({ new: { parent: gSE, element: minTime } });
+
+      const maxTime = createElement(connectedAp.ownerDocument, 'MaxTime', {
+        unit: 's',
+        multiplier: 'm',
+      });
+      maxTime.textContent = '10000';
```
danyill, line 211: Do you think 1000 might be a better default? I think this would be more typical.

```suggestion
      maxTime.textContent = '1000';
```

src/foundation/generators.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return () => {
+    const uniqueMAC = range.find(mac => !macs.has(mac));
+    if (uniqueMAC) macs.add(uniqueMAC);
+    return uniqueMAC ?? '';
+  };
+}
+
+const maxGseAppId = 0x3fff;
+const minGseAppId = 0x0000;
```
danyill, line 46: Hmmm... 8-1 Ed 2.1 suggests:

> The reserved value range for GOOSE Type 1 is 0x0000 to 0x3FFF, for GOOSE Type 1A (Trip)
> the reserved value range is 0x8000 to 0xBFFF.

Is Type 1A a real thing? Should we allow for this?

test/unit/foundation/generators.test.ts
```diff
@@ -0,0 +1,158 @@
+import { expect } from '@open-wc/testing';
+import {
+  appIdGenerator,
+  mACAddressGenerator,
+} from '../../../src/foundation/generators.js';
+
+describe('generator function for new `MAC-Address` `P` element', () => {
```
danyill, line 7: nice!

danyill (commented):
Just a few little comments, generally very nice and a learning experience for me, gracias :+1:

danyill:
I had some fun testing this by completing removing all subnets in a large file, creating them and watching the SMV and GSE elements get created. LGTM.
Not all tools allow communication structures to be easily adapted so this is helpful to us.

> first come first serve ... when an IED has two access points fisrt has the server the second pointing to the server in the first and the second is added to subnetwork all GSE and SMV is added in this ConnectedAP. When after that the first is added to the same or another subnetwork no new GSE or SMV element is added

I did not quite understand. What I do see is the same GSE is created with the same MAC-address in each different ConnectedAP that a device supports. 

So for instance if a device has three ConnectedAPs (say `M1`, `G1`, `S1`) after using this I have three GSE elements each with the same MAC address.

Is that what was intended? I might have guessed only creating in one subnetwork is what was required.

src/foundation/generators.ts
JakobVogelsang: See I did not even know this and find it a bit odd. It would be better to mark in another way that the GOOSE is a critical one, but ok. For the generator, I would leave it as is. I cannot think of a way we can guess based on the data that we need to define a type `1A` GOOSE. It just depends on too many factors. At the end, the user must decide and know this. There will be an edit wizard on the `GSE` element, and it can already be done in the edit GOOSE wizards. @danyill do you have another suggestion?

test/unit/foundation/generators.test.ts
JakobVogelsang: Thanks

src/foundation/generators.ts
JakobVogelsang: I still would add this an `type1A` option to the generator, so that others can reuse it. In the long run, those will get to `open-scd-core`.

JakobVogelsang:
> I had some fun testing this by completing removing all subnets in a large file, creating them and watching the SMV and GSE elements get created. LGTM. Not all tools allow communication structures to be easily adapted so this is helpful to us.
> 
> > first come first serve ... when an IED has two access points fisrt has the server the second pointing to the server in the first and the second is added to subnetwork all GSE and SMV is added in this ConnectedAP. When after that the first is added to the same or another subnetwork no new GSE or SMV element is added
> 
> I did not quite understand. What I do see is the same GSE is created with the same MAC-address in each different ConnectedAP that a device supports.
> 
> So for instance if a device has three ConnectedAPs (say `M1`, `G1`, `S1`) after using this I have three GSE elements each with the same MAC address.
> 
> Is that what was intended? I might have guessed only creating in one subnetwork is what was required.

Not at all. This is the behaviour I am after: Say you have 3 accP `M1`, `M2` and `M3`. The first carrying the `Server` and the second pointing to the server `ServerAt apName="M1"` the third does not have a server or server at. Let's further assume the Server has 3 `GSEControl` elements:
1. case: no connections yet and you connect M1. outcome: three GSE in new connected ap with each unique MAC and APPID
2. case: no connections yet and you connect M2. outcome:  three GSE in new connected ap with each unique MAC and APPID
3. 2. case: no connections yet and you connect M3. outcome:  no new GSE as M1 does not have server nor use a server
4. case: M1 connected already and has one of the three GSE and you want to connect M2: two new GSE are created. the MACs in the new GSE are unique as well as APPID

I this the behavior you can see with your files?

src/wizards/connectedap.ts
JakobVogelsang: Yes way not. I have the default from the IEC 61850 system configurator of SIEMENs

src/foundation/generators.ts
danyill: I have found an IED which uses this type 1A but it is in the standard... I think adding to the generator is fine.

danyill:
Thanks for the explanation, I am somewhat convinced. We will need ways to quickly allow users to select and remove if we don't provide this option while creating the subnet.

In the attached, I have merged the two branches (read only and this GSE/SMV element creator), then removed all subnetworks. Then created one. Then I have added _all_ ConnectedAP elements to this subnet. Quite a few of the IEDs have ServerAt directives and multiple AccessPoints.

[XAT_Prot1 (2).scd.zip](https://github.com/openscd/open-scd/files/9678723/XAT_Prot1.2.scd.zip)

After this I extract and take a look at the Communication Section.

* I have 5 repeats of the MAC address: `01-0C-CD-01-00-01`
* I have 5 repeats of the APPID `0001`
* I have 24 repeats of the MAC address: `01-0C-CD-04-00-01`
* I have similar repeats of APPID `4000` and `4001`.

I think you intended for the MAC address and APPIDs to be globally unique. But we iterate based on the AccessPoint and reinstantiate the lists at the start.

JakobVogelsang:
I got the issue. It happens when you select to create multiple `ConnectedAP`s at once. I do initialize the generator too late and do not write the new access points yet to the file. And this is where the discrepancy is coming from. Thanks for your tests!!

JakobVogelsang:
Should be fixed now. Thanks for testing :)

JakobVogelsang:
@danyill would you be so kind and have another look at this one :)

danyill (approved):
LGTM.

PR #1021, opened 2022-09-25
feat(editors/communication): add GSE and SMV editor type elements

### LINKED ISSUE #1020: Show `GSE` and `SMV` elements in the `Communication` section
As a user of OpenSCD I want to see the allocation of `GSE` and `SMV` elements to the `ConnectedAP` elements.

**Requirements**
- show all `GSE` and `SMV` of the server for each `ConnectedAP`
- disable those that are connected to other access point in the same `IED`

JakobVogelsang:
Close #1020 

Playing around with real IED in the lab, we saw that it does make sense for the user to see to what `ConnectedAP` elements the `GSEControl` and `SampledValueControl` are allocated. This PR is read only but action buttons will follow.


src/editors/communication/gse-editor.ts
```diff
@@ -0,0 +1,41 @@
...
+  property,
+  state,
+  svg,
+} from 'lit-element';
+
+import '@material/mwc-icon';
+
+import '../../action-icon.js';
+
+const gooseIcon = svg`<svg viewBox="0 0 24 24">
```
dlabordus, line 15: Isn't handier to add this icon to the generic icon.ts of foundation?

src/editors/communication/smv-editor.ts
```diff
@@ -0,0 +1,42 @@
...
+  property,
+  state,
+  svg,
+} from 'lit-element';
+
+import '@material/mwc-icon';
+
+import '../../action-icon.js';
+
+const smvIcon = svg`
```
dlabordus, line 15: Isn't handier to add this icon to the generic icon.ts of foundation?

dlabordus (changes requested):

src/editors/communication/smv-editor.ts
JakobVogelsang: Good hint thx

dlabordus (approved):

danyill:
I was about to press comment, and then a computer meltdown occurred!

This will mercifully reduce the quantity of my comments!

Thank you for the opportunity to look over this PR. I did not look at the code but have a few UI comments. You are right we need this feature.

Most of these could be happily bypassed or become an issue for future improvement if you think it best.

1. I am doubtful about the need to show greyed out ConnectedAPs. I think it would be better to have some filtering across IED name, and ldInst / cbName.

2. As always it would be nice to have a code editor button for at least each access point or subnetwork somehow.

3. I think the icons could perhaps be smaller, 32 px? The `G` and the `S` don't seem to scale well. Also when clicking on the IED the opening of the screen feels a bit abrasive/abrupt as it opens. Perhaps this is easing, perhaps it is an unfamiliar UI experience. It's tempting to say on large form factor devices that opening to the right is better than opening down.

4. Perhaps the text of the access point name should be centered inside the container as we do in the substation editor?

5. Why default to only showing the Address field? When I click on this I get the ConnectedAP editor. But the ConnectedAP includes the GSE and SMV elements with their Address elements. It doesn't seem semantically correct. Should we instead be able to toggle SMV/GSE/IP addresses and show them all "expanded"?

6. How should a large number of Connected APs or GSE/SMV elements be moved between subnets? If actions are per item it may not be sufficient for tasks a user might reasonably do.

7. I think ldInst and cbName should be on separate lines if possible and that sufficient space should be available to support a 12 character IED name. I think at some stage we made this change in the substation editor.

JakobVogelsang:
Thanks for the contribution. Very much appreciated!

> I was about to press comment, and then a computer meltdown occurred!
> 
> This will mercifully reduce the quantity of my comments!
> 
> Thank you for the opportunity to look over this PR. I did not look at the code but have a few UI comments. You are right we need this feature.
> 
> Most of these could be happily bypassed or become an issue for future improvement if you think it best.
> 
>     1. I am doubtful about the need to show greyed out ConnectedAPs. I think it would be better to have some filtering across IED name, and ldInst / cbName.

I wanted to reuse the UI for the `ConnectedAP` as I got good feedback for it. I found it nice to see what other GSE of the same IED are there but not connected to the same `ConnectedAP`. I would go with it for now a make the icon smaller to not give it so much space. 
> 
>     2. As always it would be nice to have a code editor button for at least each access point or subnetwork somehow.
You can access to both through the edit wizard. This will be the case for the `GSE` and `SMV` elements as well.

> 
>     3. I think the icons could perhaps be smaller, 32 px? The `G` and the `S` don't seem to scale well. Also when clicking on the IED the opening of the screen feels a bit abrasive/abrupt as it opens. Perhaps this is easing, perhaps it is an unfamiliar UI experience. It's tempting to say on large form factor devices that opening to the right is better than opening down.

This is true. Unfortunately, you cannot transform the CSS display rule. But let me see. If I can hide the disabled elements somehow differently

> 
>     4. Perhaps the text of the access point name should be centered inside the container as we do in the substation editor?
> 
>     5. Why default to only showing the Address field? When I click on this I get the ConnectedAP editor. But the ConnectedAP includes the GSE and SMV elements with their Address elements. It doesn't seem semantically correct. Should we instead be able to toggle SMV/GSE/IP addresses and show them all "expanded"?

Because the wizards become more complicated this way, and potentially very big. Just imagine you have an `ConnectedAP` say have 40 `GSE` elements. Even if we use one page per `GSE` it is just too user-friendly enough. I think wizards should be used for digestible bits.

> 
>     6. How should a large number of Connected APs or GSE/SMV elements be moved between subnets? If actions are per item it may not be sufficient for tasks a user might reasonably do.

I have some ideas. There will be an extra issue for that :)

>     7. I think ldInst and cbName should be on separate lines if possible and that sufficient space should be available to support a 12 character IED name. I think at some stage we made this change in the substation editor.

That is not possible et with the action-icon. We could put an emphasis on the cbName like so: `cbName (ldInst)`. The good thing is that you can always see the complete iedName by hovering over the title of the pane and you can also see the complete name by hovering over the icon.

danyill:
All good, thank you for your consideration.

danyill (approved):

PR #1026, opened 2022-10-01
fix(editors/later-binding): Resolve absent prefix in ExtRef for later binding subscription

### LINKED ISSUE #1005: Later Binding on specific file does not work
**Describe the bug**
The GOOSE type later binding plugin does not show any `ExtRef` when clicking on data

**To Reproduce**
Open appending file
1. Go to Subscribe - Later Binding (GOOSE)
2. Click on '´data'
3. See no reaction in possible inputs

**Expected behavior**
A selection of inputs or a warning that there are none

**Additional context**
See file
[LaterBindingGOOSEError.scd.zip](https://github.com/openscd/open-scd/files/9584092/LaterBindingGOOSEError.scd.zip)

danyill:
Partially resolves #1005

### DIFF
--- a/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
+++ b/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
@@ -94,8 +94,8 @@ export class ExtRefLaterBindingList extends LitElement {
     attributeName: string
   ): boolean {
     return (
-      extRefElement.getAttribute(attributeName) ===
-      this.currentSelectedFcdaElement?.getAttribute(attributeName)
+      (extRefElement.getAttribute(attributeName) ?? '') ===
+        this.currentSelectedFcdaElement?.getAttribute(attributeName) ?? ''
     );
   }
 
@@ -138,7 +138,7 @@ export class ExtRefLaterBindingList extends LitElement {
       extRefElement.getAttribute('iedName') ===
         this.currentIedElement?.getAttribute('name') &&
       this.sameAttributeValue(extRefElement, 'ldInst') &&
-      (this.sameAttributeValue(extRefElement, 'prefix') || extRefElement.getAttribute('prefix') === null) &&
+      this.sameAttributeValue(extRefElement, 'prefix') &&
       this.sameAttributeValue(extRefElement, 'lnClass') &&
       this.sameAttributeValue(extRefElement, 'lnInst') &&
       this.sameAttributeValue(extRefElement, 'doName') &&
--- a/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
+++ b/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
@@ -138,7 +138,7 @@ export class ExtRefLaterBindingList extends LitElement {
       extRefElement.getAttribute('iedName') ===
         this.currentIedElement?.getAttribute('name') &&
       this.sameAttributeValue(extRefElement, 'ldInst') &&
-      this.sameAttributeValue(extRefElement, 'prefix') &&
+      (this.sameAttributeValue(extRefElement, 'prefix') || extRefElement.getAttribute('prefix') === null) &&
       this.sameAttributeValue(extRefElement, 'lnClass') &&
       this.sameAttributeValue(extRefElement, 'lnInst') &&
       this.sameAttributeValue(extRefElement, 'doName') &&
--- a/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
+++ b/src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
@@ -95,7 +95,7 @@ export class ExtRefLaterBindingList extends LitElement {
   ): boolean {
     return (
       (extRefElement.getAttribute(attributeName) ?? '') ===
-        this.currentSelectedFcdaElement?.getAttribute(attributeName) ?? ''
+        (this.currentSelectedFcdaElement?.getAttribute(attributeName) ?? '')
     );
   }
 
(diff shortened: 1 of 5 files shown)


JakobVogelsang:
Very nice! Thanks for tracing back the issue. Would it be possible to add a regression test?

src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
```diff
@@ -138,7 +138,7 @@ export class ExtRefLaterBindingList extends LitElement {
       extRefElement.getAttribute('iedName') ===
         this.currentIedElement?.getAttribute('name') &&
       this.sameAttributeValue(extRefElement, 'ldInst') &&
-      this.sameAttributeValue(extRefElement, 'prefix') &&
+      (this.sameAttributeValue(extRefElement, 'prefix') || extRefElement.getAttribute('prefix') === null) &&
```
dlabordus, line 141: I don't think this is the full solution, because if the value of 'this.currentSelectedFcdaElement' is not null they are not the same.
I think the solution is to see if we can make the function 'sameAttributeValue also handle null values correctly.

dlabordus (changes requested):

src/editors/subscription/later-binding/ext-ref-later-binding-list.ts
dlabordus: Maybe this implementation of sameAttributeValue will work better

```
  private sameAttributeValue(
    extRefElement: Element,
    attributeName: string
  ): boolean {
    return (
      extRefElement.getAttribute(attributeName) ?? '' ===
      this.currentSelectedFcdaElement?.getAttribute(attributeName) ?? ''
    );
  }```

Comment (danyill) [reply src/editors/subscription/later-binding/ext-ref-later-binding-list.ts:141]: Thank you both for this review. I will add a few regression tests.

>I don't think this is the full solution, because if the value of 'this.currentSelectedFcdaElement' is not null they are not the same.
>I think the solution is to see if we can make the function 'sameAttributeValue also handle null values correctly.

:+1: You are correct, `prefix` is optional on both the FCDA and the ExtRef.

But I am less sure about the solution. This is not generally true for all attributes in the FCDA or ExtRef and isn't it "sloppy parsing" to allow this through in general?

Shouldn't we require that the SCL files are well-formed and apply this only to optional attributes by exception?

So at risk of overcomplicating things:

```js
(this.sameAttributeValue(extRefElement, 'prefix') || extRefElement.getAttribute('prefix') === null || this.getAttribute('prefix') === null) 
```

?

dlabordus (approved):

danyill:
Please don't merge I must do more here...

danyill:
Thank you for the review, I understand better now, despite this being quite trivial :wink: :cry: 

In preparing the tests I ran the scd file through OpenSCD which is why there is some whitespace diff.

Without this:

```diff
@@ -71,6 +71,9 @@
                                                <Inputs>
                                                        <ExtRef intAddr="Pos;CSWI1/Pos/stVal" desc="Interlocking.Input"/>
                                                        <ExtRef intAddr="Pos;CSWI1/Pos/q" desc="Interlocking.Input" iedName="GOOSE_Publisher" serviceType="GOOSE" ldInst="QB2_Disconnector" lnClass="CSWI" lnInst="1" prefix="" doName="Pos" daName="q" srcLDInst="QB2_Disconnector" srcPrefix="" srcLNClass="LLN0" srcCBName="GOOSE2"/>
+                                                       <ExtRef intAddr="Pos;CILO/EnaOpn/stVal" desc="Interlocking.Input2" iedName="GOOSE_Publisher" serviceType="GOOSE" ldInst="QB1_Disconnector" lnClass="CILO" lnInst="1" prefix="" doName="EnaOpn" daName="stVal" srcLDInst="QB2_Disconnector" srcPrefix="" srcLNClass="LLN0" srcCBName="GOOSE2"/>
+                                                       <ExtRef intAddr="Pos;CILO/EnaCls/stVal" desc="Interlocking.Input3" iedName="GOOSE_Publisher" serviceType="GOOSE" ldInst="QB1_Disconnector" lnClass="CILO" lnInst="1" doName="EnaCls" daName="stVal" srcLDInst="QB2_Disconnector" srcPrefix="" srcLNClass="LLN0" srcCBName="GOOSE2"/>
+                                                       <ExtRef intAddr="Pos;CILO/EnaOpn2/stVal" desc="Interlocking.Input4" iedName="GOOSE_Publisher" serviceType="GOOSE" ldInst="QB2_Disconnector" lnClass="CILO" lnInst="1" doName="EnaOpn" daName="stVal" srcLDInst="QB2_Disconnector" srcPrefix="" srcLNClass="LLN0" srcCBName="GOOSE2"/>
                                                </Inputs>
                                        </LN>
                                        <LN prefix="" lnClass="CSWI" inst="1" lnType="Dummy.CSWI">
@@ -94,6 +97,9 @@
                                                <DataSet name="GOOSE2sDataSet">
                                                        <FCDA ldInst="QB2_Disconnector" prefix="" lnClass="CSWI" lnInst="1" doName="Pos" daName="stVal" fc="ST"/>
                                                        <FCDA ldInst="QB2_Disconnector" prefix="" lnClass="CSWI" lnInst="1" doName="Pos" daName="q" fc="ST"/>
+                                                       <FCDA ldInst="QB1_Disconnector" lnClass="CILO" lnInst="1" doName="EnaOpn" daName="stVal" fc="ST"/>
+                                                       <FCDA ldInst="QB1_Disconnector" prefix="" lnClass="CILO" lnInst="1" doName="EnaCls" daName="stVal" fc="ST"/>
+                                                       <FCDA ldInst="QB2_Disconnector" lnClass="CILO" lnInst="1" doName="EnaOpn" daName="stVal" fc="ST"/>
                                                </DataSet>
                                                <DataSet name="GOOSE1sDataSet">
                                                        <FCDA ldInst="QB1_Disconnector" prefix="" lnClass="CSWI" lnInst="1" doName="Pos" daName="stVal" fc="ST"/>
```

I have added CILO LN subscriptions with the following combinations, which I think cover the relevant cases.

```
FCDA        ExtRef
No prefix   Prefix
Prefix      No prefix
No prefix   No prefix
```

dlabordus (approved):

PR #1028, opened 2022-10-02
feat(hosting) Allow renaming project file

### LINKED ISSUE #1027: Allow user to rename project

As a user I'd like to be able to rename my project. This helps me keep things orderly without requiring it to be a step after going Save project and downloading or before opening a project in OpenSCD.

This is not especially important but it helps me track my workflow.

Perhaps the menu could be a little wider and an edit/rename icon beside the project name:

image

Even better might be the ability to click on the project name on the menu bar/drawer:

image

danyill:
Closes #1027

This may be an unwanted feature since the issue has yet to go through refinement.

My excuse is that I was curious to see if I could do it :wink: 

I couldn't figure out the correct CSS to allow the width of the filename using the `<input`> element to be longer (it should be longer by default then it is as the previous implementation had).

I'm not quite sure whether this requires tests or where best to put them.

### DIFF
--- a/src/Hosting.ts
+++ b/src/Hosting.ts
@@ -275,11 +275,19 @@ export function Hosting<
 
     private changeProjectName(e: Event): void {
       let proposedName = (<HTMLInputElement>e.target!).value;
-      proposedName = proposedName.indexOf('.')
-        ? `${proposedName}.scd`
-        : proposedName;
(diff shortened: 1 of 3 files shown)


danyill:
We await the future where this will be elegantly resolved.

PR #1034, opened 2022-10-05
fix(editors/cleanup) Cleanup of datatypes is buggy

### LINKED ISSUE #1008: Datatype Cleanup Buggy
**Describe the bug**

As Sander prompted me, I reluctantly confess... :zipper_mouth_face: 

The datatype cleanup editor has several problems. I am the cause of all of them :man_facepalming: :wink: 

When the remove subtypes is checked:

1.  It sometimes takes several passes to remove all the templates. This is not what was intended. This was found by testing larger scd files, removing a whole IED, and cleaning up the unconnected DataType references.
2. It has a hard-coded `MAX_STACK_DEPTH` of 10,000 to ensure there are no recursive references. However, when indexing all templates, to begin with on a decent sized scd file, it is not hard to get close to this on my trial files. Some IEDs have very verbose data models.
3. If `MAX_STACK_DEPTH` is exceeded, it will silently fail. That is not very good. As a minimum, a log entry? Or a console debug statement would seem reasonable.

Additionally, on openscd.github.io the filtering doesn't seem to be working correctly either, items become grayed out rather than removed from the filtered list.

Once the Remove button is pushed, the checkboxes should all become deselected. This also applies to the cleanup and dataset windows.

Point 1 is the slightly complex one. I'm still thinking about it, but I think the point of entry is incorrect. We should start off by something like:

1. Creating an index entry for all datatypes in the DataTypeTemplates section. All entries start with usage of zero.
2. Indexing recursively based on the LNodeTypes which are _not_ selected for removal.
3. After this, those templates used will have a positive value and those without will be at 0.

The tests also need to be a little bit better. The most important one is the LNodeType which has to be treated differently from disconnected DOType, DAType or EnumTypes.

danyill:
Closes #1008

I have not been able to break this for some time now. Most of my recent tests have been around removing IEDs in large scd files and see that the templates still validate (everything is there that should be) and there are no "left behind" templates.

I have tried to work more with ids than elements to improve speed. Still, it hangs in Chrome frequently, and we have to click "Wait" to make it continue on large files. I think we need a way to deal with long-running tasks in the UI in future. Should it block with a modal/dialog? Should it be turned into a "this task will be completed in the background" with a task queue? Not today's problem... I guess :wink: 

I will be grateful for a review and perhaps testing with some scd files or project files would be prudent.


src/editors/cleanup/datatypes-container.ts
```diff
@@ -266,29 +271,43 @@ export class CleanupDataTypes extends LitElement {
...
+
+      if (elementStack.length >= MAX_STACK_DEPTH) {
+        this.dispatchEvent(
+          newLogEvent({
+            kind: 'error',
+            title: get('cleanup.unreferencedDataTypes.title'),
+            message: get('cleanup.unreferencedDataTypes.stackExceeded', {
+              maxStackDepth: MAX_STACK_DEPTH.toString(),
+            }),
+          })
```
danyill, line 306: I am not sure but perhaps this is causing the test failure?

```
test/unit/Logging.test.ts:

 ❌ LoggingElement > cannot undo warning messages
      TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.
        at qy.<anonymous> (test/unit/Logging.test.ts:32:12)
```

src/translations/de.ts
```diff
@@ -745,7 +745,8 @@ export const de: Translations = {
       tooltip:
         'Datentypen, die nicht in einem logischen Knoten oder einem anderen verwendeten Datentyp referenziert werden',
       alsoRemoveSubTypes: 'Entfernen Sie auch Untertypen',
-    },
+      stackExceeded: 'Maximale Stapellänge überschritten. Maximal zulässig ist {{maxStackDepth}}. Die Reinigung des Datentyps ist unvollständig, und es ist möglicherweise ein Dateischaden aufgetreten.'
```
JakobVogelsang, line 748:
```suggestion
      stackExceeded: 'Maximale Aufrufe überschritten. Maximal zulässig sind {{maxStackDepth}}. Nicht alle überflüßigen Datentypen sind entfernt und das Project is potentiel beschädigt.'
```

JakobVogelsang (approved):
LGTM other than the german translation. I would not reproduce your test error. Maybe You can run a rebase or merge on/to main.

src/editors/cleanup/datatypes-container.ts
danyill: I can reproduce it consistently including on a fresh clone but it doesn't appear to be related to this code so I'll mark this as resolved.

danyill:
Thank you for reviewing this. :+1: 
I have made the changes requested, and will merge this, hoping that the test failure will disappear.

danyill:
Regrettably now `main` is broken (I am sorry :cry:), and I can see why now. 

When I do a `git diff main` from this branch, or I look at the changed files in Github I see:

```diff
\ No newline at end of file
diff --git a/test/unit/Logging.test.ts b/test/unit/Logging.test.ts
index 46fbd866..0627f57e 100644
--- a/test/unit/Logging.test.ts
+++ b/test/unit/Logging.test.ts
@@ -30,7 +30,7 @@ describe('LoggingElement', () => {
 
   it('cannot undo warning messages', () => {
     element.dispatchEvent(
-      newLogEvent({ kind: 'warning', title: 'test warning' })
+      ({ kind: 'warning', title: 'test warning' })
     );
```

image

I don't see this when I open `test/unit/Logging.test.ts` on either this branch or main.
This is a mysterious and new evil in my world.

Issue #1016, opened 2022-09-25
FCDA order doesn't follow dataset in GOOSE later binding editor


danyill:
**Describe the bug**

I was doing a little testing and noticed that the sort order was unexpected in the later binding GOOSE and SMV plugins.

Using the attached file:

[XAT_Prot2_Base_Configuration.scd.zip](https://github.com/openscd/open-scd/files/9641628/XAT_Prot2_Base_Configuration.scd.zip)

Where a GOOSE message is made including the following dataset:

```xml
<DataSet desc="A dataset for transformer trips" name="Ctl">
    <Private type="Siemens-GUID">9a196dba9bdb402191c79b9e8b78e0c5</Private>
    <FCDA ldInst="ANN" prefix="SVT" lnClass="GGIO" lnInst="4" doName="Ind01" daName="stVal" fc="ST" />
    <FCDA ldInst="ANN" prefix="SVT" lnClass="GGIO" lnInst="4" doName="Ind01" daName="q" fc="ST" />
    <FCDA ldInst="ANN" prefix="INC" lnClass="GGIO" lnInst="13" doName="Ind01" daName="stVal" fc="ST" />
    <FCDA ldInst="ANN" prefix="INC" lnClass="GGIO" lnInst="13" doName="Ind01" daName="q" fc="ST" />
</DataSet>
```

In OpenSCD the order appears reversed on the data attributes.

![image](https://user-images.githubusercontent.com/674783/192161978-f8c766c9-30e3-4227-85fc-1ae7f95f424e.png)

It is correct in the Publisher plugin:

![image](https://user-images.githubusercontent.com/674783/192161971-83495686-bc85-4f99-8b8e-c8ec983da69c.png)

**To Reproduce**

1. Open the attached scd file.
2. Go to Later Binding (GOOSE)
3. View the FCDAs on the left.

**Expected behavior**

The order should follow the dataset order by default.

dlabordus:
Is it not better to have a predictable ordering? 
Because if the order is how it's in the XML, searching can become a issue if someone added them in some random order.

danyill:
> Is it not better to have a predictable ordering?
> Because if the order is how it's in the XML, searching can become a issue if someone added them in some random order.

I think there's a good argument each way but:

* Searching can be done with the search bar?
* The FCDAs in datasets can be ordered by the end user. Shouldn't we respect this a little bit?
[...]

PR #1040, opened 2022-10-12
feat(editor/subscriber): Show counter for Subscriber Plugins (Logical Nodes / Later) (GOOSE/SMV)

dlabordus:
This will replace PR https://github.com/openscd/open-scd/pull/993.

Thanks @danyill for the code. 

It will added the counter to the Left Side of the Editor for both Data and Later Binding.
Through a event the counter will be updated if the subscription changes.

### DIFF
--- a/src/editors/subscription/fcda-binding-list.ts
+++ b/src/editors/subscription/fcda-binding-list.ts
@@ -24,7 +24,13 @@ import {
 import { gooseIcon, smvIcon } from '../../icons/icons.js';
 import { wizards } from '../../wizards/wizard-library.js';
 
-import { getFcdaTitleValue, newFcdaSelectEvent, styles } from './foundation.js';
+import {
+  getFcdaTitleValue,
+  newFcdaSelectEvent,
+  styles,
+  SubscriptionChangedEvent,
+} from './foundation.js';
+import { getSubscribedExtRefElements } from './later-binding/foundation.js';
 
 type controlTag = 'SampledValueControl' | 'GSEControl';
 
@@ -41,14 +47,18 @@ export class FcdaBindingList extends LitElement {
   doc!: XMLDocument;
   @property()
   controlTag!: controlTag;
+  @property()
+  includeLaterBinding!: boolean;
 
   // The selected Elements when a FCDA Line is clicked.
   @state()
-  selectedControlElement: Element | undefined;
+  private selectedControlElement: Element | undefined;
   @state()
-  selectedFcdaElement: Element | undefined;
+  private selectedFcdaElement: Element | undefined;
+  @state()
+  private extRefCounters = new Map();
 
-  iconControlLookup: iconLookup = {
+  private iconControlLookup: iconLookup = {
     SampledValueControl: smvIcon,
     GSEControl: gooseIcon,
   };
@@ -58,6 +68,12 @@ export class FcdaBindingList extends LitElement {
 
     this.resetSelection = this.resetSelection.bind(this);
     parent.addEventListener('open-doc', this.resetSelection);
+
+    const parentDiv = this.closest('.container');
+    if (parentDiv) {
+      this.resetExtRefCount = this.resetExtRefCount.bind(this);
+      parentDiv.addEventListener('subscription-changed', this.resetExtRefCount);
+    }
   }
 
   private getControlElements(): Element[] {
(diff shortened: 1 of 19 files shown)


dlabordus:
@danyill, which issue will this PR close? Would be nice to link it to this PR.

src/editors/subscription/fcda-binding-list.ts
```diff
@@ -59,13 +68,17 @@ export class FcdaBindingList extends LitElement {
...
+      parentDiv.addEventListener('subscription-changed', this.resetExtRefCount);
+    }
   }
 
   private getControlElements(): Element[] {
     if (this.doc) {
-      return Array.from(
-        this.doc.querySelectorAll(`LN0 > ${this.controlTag}`)
-      ).sort((a, b) => compareNames(`${identity(a)}`, `${identity(b)}`));
+      return Array.from(this.doc.querySelectorAll(`LN0 > ${this.controlTag}`));
```
danyill, line 81: Perhaps we should add closes #1016 somewhere? I am pleased to see this :smile:

src/editors/subscription/fcda-binding-list.ts
```diff
@@ -79,11 +92,40 @@ export class FcdaBindingList extends LitElement {
...
           )}] > FCDA`
         )
-      ).sort((a, b) => compareNames(`${identity(a)}`, `${identity(b)}`));
+      );
     }
     return [];
   }
 
+  private resetExtRefCount(event: SubscriptionChangedEvent): void {
+    if (event.detail.control && event.detail.fcda) {
```
danyill, line 101: For my education why do we pass the control and and fcda back. Do we not trust the selection in fcda-binding-list?

src/editors/subscription/fcda-binding-list.ts
```diff
@@ -92,6 +134,7 @@ export class FcdaBindingList extends LitElement {
   private resetSelection(): void {
     this.selectedControlElement = undefined;
     this.selectedFcdaElement = undefined;
+    this.extRefCounters = new Map();
```
danyill, line 137: Alas, this we cannot have, it completely removes all indexing :stuck_out_tongue_winking_eye:

danyill (changes requested):
Thank you for reviewing, tidying up and improving my efforts :clap: 

There is a one liner we must fix as this PR breaks the indexing completely:

* I tested with `IOP_2019_HV_v9_ed2.1.scd`, I found that it takes 5-10 seconds to update when I click on an FCDA in the GOOSE later-binding plugin for every different FCDA I select. 
* In my branch `issue-958-number-of-connections` it only seems to happen for the first one (as I would expect).

We reset the map where we should not, I have identified the place where it occurs.

src/editors/subscription/fcda-binding-list.ts
danyill: It occurred to me that we need something like this in the `updated` method so that if the user opens a new document after modifying the first one, the counters are reset.

Something like this:

```js
  protected updated(_changedProperties: PropertyValues): void {
    super.updated(_changedProperties);

    // When the document is updated, we will fire the event again.
    if (
      _changedProperties.has('doc') ||
      _changedProperties.has('selectedControlElement') ||
      _changedProperties.has('selectedFcdaElement')
    ) {
      this.dispatchEvent(
        newFcdaSelectEvent(
          this.selectedControlElement,
          this.selectedFcdaElement
        )
      );
    }
    
    // *add this bit here*
    if (_changedProperties.has('doc')) {
      this.extRefCounters = new Map();
    }
  }
```

dlabordus: You are right here. This will improve the performance correctly :-)

src/editors/subscription/fcda-binding-list.ts
dlabordus: I prefer passing the data and make it work independent, this way component firing the event determines what data to use.

src/editors/subscription/fcda-binding-list.ts
dlabordus: Added this issue to the PR to close it. 

The first sorting I added for later binding was on intAddr and that seems strange. So I just removed the sorting for the Binding (Later and Data)

test/integration/editors/GooseSubscriberDataBinding.test.ts
```diff
@@ -48,6 +49,7 @@ describe('GOOSE Subscribe Data Binding Plugin', async () => {
     expect(extRefListElement['getSubscribedLNElements']().length).to.be.equal(
       0
     );
+    expect(getSelectedSubItemValue(fcdaListElement)).to.be.null;
```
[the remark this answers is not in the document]
dlabordus: I think this is a way it will work. We could also check the Map for the entry we are updating.

I only removed the selectedItem variable from the component and just did the querying in a function in the test.
It is not a good practice to add code to a component that is only used for testing purpose.

src/editors/subscription/later-binding/foundation.ts
```diff
@@ -94,28 +97,72 @@ function checkEditionSpecificRequirements(
...
-  iedElement: Element | undefined,
+  controlTag: 'SampledValueControl' | 'GSEControl',
   controlElement: Element | undefined,
   fcdaElement: Element | undefined,
   extRefElement: Element
 ): boolean {
   return (
     extRefElement.getAttribute('iedName') ===
-      iedElement?.getAttribute('name') &&
+      fcdaElement?.closest('IED')?.getAttribute('name') &&
```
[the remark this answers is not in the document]
dlabordus: Is this a question? 

It may slowdown the code a little bit, but I prefer doing it like this, because the IED that should be used is the one of the FCDA/Control Element. Passing the IED can cause errors when developer start passing the wrong IED.

dlabordus:
@danyill, can you try again if the performance is better?

I fixed the update method. This was indeed needed to make it faster.

danyill (approved):

danyill:
LVGTM, thanks!

PR #1041, opened 2022-10-13
bug(editor/subscriber): When selecting ExtRef also include Control Reference Attributes (GOOSE/SMV)

dlabordus:
When using a newer version than 2003 the ExtRef also include the attributes 'srcLDInst', 'srcLNClass', 'srcLNInst' and 'srcCBName' of the Control Element (SMV/GOOSE) (and ServiceType was also missing). 
These attributes were not added when search for a matching ExtRef. So it found a matching ExtRef much faster and then not creating a new ExtRef.


test/integration/editors/__snapshots__/SMVSubscriberMessageBinding.test.snap.js
```diff
@@ -850,7 +850,7 @@ snapshots["Sampled Values Plugin in Subscriber view when selecting an IED and su
         aria-disabled="false"
         noninteractive=""
         tabindex="-1"
-        value="IED4>>CircuitBreaker_CB1>MSVCB02"
```
JakobVogelsang, line 853: Is that correct? I would expect the heading list item to not disappear when filtering for e.g. `MSVB02`

JakobVogelsang: Ahh I see that changed with the changed test file. Is that correct?

JakobVogelsang (commented):
In general LGTM. Just have one question. :) Thanks for properly updating the snapshots!

test/integration/editors/__snapshots__/SMVSubscriberMessageBinding.test.snap.js
JakobVogelsang: The reference is correct now. That is what matters.

JakobVogelsang (approved):
LGTM

PR #1043, opened 2022-10-15
fix(Logging) Correct toggle colors in log

### LINKED ISSUE #1035: Log filter buttons don't appear to toggle state
**Describe the bug**

Once upon a time the log filter buttons changed colour when they were enabled/disabled. They don't any more.

image

**To Reproduce**
1. Go to the log.
2. Click on the filter buttons.
3. Observe no change in colour.

danyill:
Closes #1035

I'm not sure what was wrong with the previous approach. 

The inline style expression had an extra bracket on the end (from when I modified it to allow toggle buttons in the cleanup editor), but that was not the only cause.

It seems more general to rely on CSS here than inline styles, and that also resolves the issue, so perhaps this is an acceptable fix? (In the cleanup editor I used CSS and didn't add to the colours which had the side-effect of an "undefined" in the styles attribute.)

### DIFF
--- a/src/Logging.ts
+++ b/src/Logging.ts
@@ -320,6 +320,22 @@ export function Logging<TBase extends LitElementConstructor>(Base: TBase) {
             display: flex;
           }
 
+          #infofilter[on] {
+            color: var(--cyan);
+          }
+
+          #warningfilter[on] {
+            color: var(--yellow);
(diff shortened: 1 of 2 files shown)


JakobVogelsang (approved):
LGTM

PR #1044, opened 2022-10-16
feat(menu/exportCommunication): Allow export of communication section 

### LINKED ISSUE #1042: Export Communication Section Plugin
As a user of OpenSCD I want to export a bare Communication section without any other information included. There currently is no defined file format for this.

I'd like a plugin that exports just the Communication section from an scd as a single SCL file.

danyill:
Closes #1042 

This seemed straightforward to me and within my capability, so I thought I might do it.

I only belatedly realised that I had not carefully read the issue and noticed that @ca-d had self-assigned. Feel free to close my PR if this was not what you had in mind or if that's easier, I will not mind in the least :wink:

I'm not very sure about the icon or the tests and the name of the extension doesn't quite fit the Menu (should the menu be wider?)

Incidentally, I'm interested in this because I'd like to define (at some point) a custom file format for the communication section to function as an interface for confirming IPAM and to enable building network switch configuration with Communication section information on VLANs and with MAC address filtering for SV traffic. At some stage, I'll raise a separate issue for that. What is missing is "who needs" the GOOSE/SMV traffic.


danyill:
@ca-d suggested making this a conformant SCL file by including the SCL root note and the Header element as well. I will add these.

danyill:
> @ca-d suggested making this a conformant SCL file by including the SCL root note and the Header element as well. I will add these.

I have done these, I hope - I will be grateful for a review and happy to make improvements.
I will also want to add some documentation for this plugin for the Help function.

src/menu/ExportCommunication.ts
```diff
@@ -0,0 +1,72 @@
+import { LitElement, property } from 'lit-element';
+import { get } from 'lit-translate';
+
+import { saveXmlBlob } from './SaveProject.js';
```
danyill, line 4: Perhaps this method should be in `./src/foundation.ts`?

src/menu/SaveProject.ts
```diff
@@ -1,16 +1,38 @@
...
+  a.href = URL.createObjectURL(blob);
+  a.dataset.downloadurl = ['application/xml', a.download, a.href].join(':');
+  a.style.display = 'none';
+  document.body.appendChild(a);
+  a.click();
+  document.body.removeChild(a);
+  setTimeout(function () {
+    URL.revokeObjectURL(a.href);
+  }, 5000);
+  return blob;
```
danyill, line 35: It seems a little clunky to return the blob, which I am only doing for the tests.

public/js/plugins.js
```diff
@@ -225,4 +225,13 @@ export const officialPlugins = [
     requireDoc: false,
     position: 'bottom',
   },
+  {
+    name: 'Export Communication Section',
+    src: '/src/menu/ExportCommunication.js',
+    icon: 'share',
```
danyill, line 231: It is not my favourite icon but I have not found a better one. Certainly on Android, sharing files or communicating information between apps sometimes uses this icon but I think the metaphor is a little confused.

JakobVogelsang: Why not refuse the same icon as we use for the `Communication` editor plugin?

src/menu/ExportCommunication.ts
ca-d: In principle it would be nice to have a unified way to locally open and save XML files one day, and I'd like to include one in `open-scd-core`'s main entry point (à la `import { saveDocLocal, openDocLocal } from '@openscd/open-scd-core`), however I'd like to use a much more involved mechanism for constructing (or parsing when opening) the XML `Blob` Object into an `XMLDocument` object without running into Firefox's `1GiB` `string` size limit (or Chrome's even worse `512MiB`).

For this simpler ad-hoc solution (just `XMLSerializer().serializeToString` the whole thing and dump it in a blob in one piece to be downloaded) I'd suggest just copying the code wherever it's needed and replacing it with a more robust central solution later. What I'd try to avoid is importing code from another plugin, I'd rather see those lines duplicated here. How do you view this topic?

src/menu/SaveProject.ts
ca-d: I concur, let's try to never alter code, export functions, or expose internal data just in order to make it more testable. That's why we decided to leave this stuff untested so far - we technically can't test the browser's "Download file" functionality and must assume that the blob construction web API works as advertised, since Browsers test this using https://github.com/web-platform-tests/wpt .

public/js/plugins.js
ca-d: Apart from that, we could also re-use the `save` icon we use for regular saving (plugin name "Communication Export" makes it clear), or if it's visually too undistinguished we could use one of [these here download icons](https://fonts.google.com/icons?icon.query=download)

ca-d:
> > @ca-d suggested ...
> 
> I have done these

Thank you, @danyill - great work overall! Regarding the Help, I think it'll be useful to update the wiki, but I'd really like to fix the help viewer (internal links currently do not work) before updating the content again. Currently the built in help viewer is hardly usable to me.

My current favored solution for fixing this bug:

1. Introduce a build step which compiles the markdown to static HTML files into the distribution directory.
2. Get rid of the markdown renderer runtime dependency.
3. Render the static html help statically into an `iframe` using `import.meta.url` to construct the `iframe`'s original `src` attribute.
4. ???
5. Profit!

ca-d (commented):
Not all of the comments need to be attended to in this pull request (especially not help viewer bugfixes and a more robust foundation save blob function), but let's discuss them and make extra issues for the ones we want to implement before merging this!

danyill:
:+1: Thank you both for these review comments, I hope to look at these over the next few days

public/js/plugins.js
danyill: Great suggestions, thank you.

I'm going to go with the one from ca-d. I prefer the icon to be associated with the verb (Save/Export) than the noun (Communication) because I think users mostly think of the "action word" and this is the best association.

src/menu/ExportCommunication.ts
danyill: I agree that we must not pollute extensions and think that is probably the overriding concern here.

src/menu/SaveProject.ts
danyill: Quite right, thank you. Junior developer recognises error 🤦

danyill:
> Regarding the Help, I think it'll be useful to update the wiki, but I'd really like to fix 

I have logged #1057

src/menu/ExportCommunication.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import { formatXml, newLogEvent } from '../foundation.js';
+
+function cloneAttributes(destElement: Element, sourceElement: Element) {
+  let attr;
+  const attributes = Array.prototype.slice.call(sourceElement.attributes);
+  while ((attr = attributes.pop())) {
+    destElement.setAttribute(attr.nodeName, attr.nodeValue);
+  }
+}
```
danyill, line 12: I am not sure if this is a safe way to transfer namespace declarations.

test/integration/menu/ExportCommunication.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    element = <ExportCommunication>(
+      parent.querySelector('export-communication')!
+    );
+
+    element.doc = doc;
+    element.docName = 'CommunicationTest.scd';
+    element.run();
+    await parent.requestUpdate();
+  });
```
ca-d, line 35: I don't get this test file, did you have some tests in here before, or did you still intend to add some? If you want help coming up with tests, I'd be happy to think of a couple together.

danyill:
I did have before but removed as per review comments.

I thought this would at least test the plugin functions without an
exception.

I would be pleased to receive some help.

ca-d (approved):
LGTM

danyill:
Thank you ca-d for reviewing this and for providing the tests

PR #1048, opened 2022-10-20
bug(editor/subscriber): Filtering on FCDA Value as displayed made possible

### LINKED ISSUE #1047: Filtering for Data/Later Binding not working correctly for Headers of FCDA
**Describe the bug**
When using the filter on the left side list (FCDA List) the value of DOName and DAName are displayed seperated by a '.' character.
This value is not added to the Header (SMV/GOOSE) in this way, the value of DoName and DAName are in there but, not separated by a '.' character.

**To Reproduce**
1. Open a Test File for SMV/GOOSE Data Binding or Later Binding
2. Go to one of the Data or Later Binding Editors
3. Filter on something like 'AmpSv.instMag.i', where 'AmpSv' is the DOName value and the other part the DAName
4. Header isn't displayed anymore

**Expected behavior**
Header should still be visible.

**Screenshots**
Before filtering the header is there.
image

After the filter header should be there but is not displayed anymore.
image

 - OS: Ubuntu
 - Browser: Chrome

**Additional context**
The value as is displayed should also be added to the value attribute of the Header, not only the identity.

dlabordus:
closes #1047

### DIFF
--- a/src/editors/subscription/fcda-binding-list.ts
+++ b/src/editors/subscription/fcda-binding-list.ts
@@ -25,6 +25,7 @@ import { gooseIcon, smvIcon } from '../../icons/icons.js';
 import { wizards } from '../../wizards/wizard-library.js';
 
 import {
+  getFcdaSubtitleValue,
   getFcdaTitleValue,
   newFcdaSelectEvent,
   styles,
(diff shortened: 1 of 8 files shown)


juancho0202 (approved):
LGTM

PR #1061, opened 2022-10-30
feat(editors/subscription) Add supervision indication

### LINKED ISSUE #942: Add `LGOS` and `LSVS` acc. to IEC 61850-6 and IEC 61850-7-1

### LINKED ISSUE #1037: Make LGOS and LSVS accessible in subscription editors

When #942 is complete, supervision of subscriptions will be provided.
But it will not be visible to the user.

As a user, I care about this because I want to know:

* Where is this subscription supervised?
* When browsing the SCL object model online, which LSVS or LGOS logical node is associated with the communications of this FCDA or Control Element.

I would like to see the path to the LN and the LN name and a small icon to indicate supervision, perhaps `monitor_heart`

Perhaps something like one of these for the later binding plugin:

image

Perhaps only the LN information could be shown with the fully qualified path being a tooltip.

Another option might be to show only the `monitor_heart` icon with the tooltip being the path and the LGOS name which would reduce clutter. This would mean "this subscription is supervised" and the user would need to hover to see the detail. The icon could be a button which opened the Logical Node editor for the user to access additional details.

In the Subscriber(GOOSE) and Subscriber(SMV) plugins, perhaps it could be on the right.

image

image

danyill:
Closes #1037 

Possibly helpful for functional reviewing?


src/editors/subscription/foundation.ts
```diff
@@ -343,6 +344,56 @@ export function findOrCreateAvailableLNInst(
...
+    ?.querySelectorAll('LN0 > Inputs, LN > Inputs')
+    .forEach(inputs => {
+      Array.from(dataSet!.querySelectorAll('FCDA')).some(fcda => {
+        const anExtRef = getExtRef(inputs, fcda, publishedControlBlock);
+        if (anExtRef) {
+          extRef = anExtRef
+          return true;
+        }
+      });
+    });
```
danyill, line 366: This is not very efficient, I must improve the way the iteration is done.

danyill:
OK, I think it's possible that if the tests are working then this will pass. I've got issues running the tests locally at the moments (a lot of test failures with timeouts and I do not know why).

1. I've augmented the message binding test files with LGOS and LSVS supervision nodes.
2. I've done what I can with the unit tests.
3. The logic of the `getFirstSubscribedExtRef` function is not very clean and I will be interested in any suggestions for improvement.

I would be most grateful for a review :smile:

danyill:
While watching a [:rocket: launch](https://www.rocketlabusa.com/live-stream/) I realised I was returning the identity of Val not the LN, so I have corrected.

JakobVogelsang:
It looks all good to me @danyill. Nice job:) I would wait for #942 to go in? Is that ok with you. I think there would be many merge conflicts otherwise.

danyill:
Sure, thanks for taking a look

I targeted [openscd:942-subscriber-add-lgos-and-lsvs](https://github.com/openscd/open-scd/tree/942-subscriber-add-lgos-and-lsvs) instead of main but don't mind which order things are done in, whatever is easy.

danyill:
:cry: did you mean to close this @juancho0202?

juancho0202:
@danyill not at all I'm so sorry, I thought it would automatically change to point to `main` instead of closing it 😔

danyill:
OK, thanks @juancho0202.

I think it cannot repoint because you have force-pushed your branch and broken the dependency chain (and deleted the evidence!). I often force push too when my work is "private" but once there is a branch depending on it then to force push shared commits is to cause damage.

So I think Github is not providing any options on this PR because there is no way to reconnect it.

I'll do something when time allows.

PR #1065, opened 2022-11-02
fix(Logging): proper event usage

danyill:
See [comment](https://github.com/openscd/open-scd/pull/1034#issuecomment-1299623520).
This should fix the broken tests which should not be broken.


JakobVogelsang (approved):
Should not have happened in the first place...sorry, my review discipline was terrible. I promise to improve

PR #1073, opened 2022-11-07
 Add read-only transformer winding editor to substation plugin 

juancho0202:


src/editors/substation/transformer-winding-editor.ts
```diff
@@ -0,0 +1,95 @@
...
+
+@customElement('transformer-winding-editor')
+export class TransformerWindingEditor extends LitElement {
+  /** The document being edited as provided to editor by [[`Zeroline`]]. */
+  @property({ attribute: false })
+  doc!: XMLDocument;
+  /** SCL element TransformerWinding */
+  @property({ attribute: false })
+  element!: Element;
+
```
dlabordus, line 29: I think you need a showfunctions boolean property that needs to be passed to eq-function-editor.
It need to be set from powertransformer-editor which gets it from it parent editor.

Maybe it also needs to be passed to l-node-editor, but you need to check this.

This is something that slowly is being added to all the editors of the substation plugin.

src/foundation.ts
```diff
@@ -1703,6 +1703,7 @@ export const tags: Record<
       ...tAbstractConductingEquipmentSequence,
       'EqFunction',
       'SubEquipment',
+      'TransformerWinding',
```
dlabordus, line 1706: Why is this added? Does ConductingEquipment has TransformerWindign Elements?

test/testfiles/TransformerWinding.scd
```diff
@@ -0,0 +1,252 @@
...
+			<Bay name="Q03" desc="">
+				<Function name="F1">
+					<SubFunction name="Sub1">
+						<LNode iedName="None" lnClass="CSWI" lnInst="1" lnType="OpenSCD_CSWI"/>
+					</SubFunction>
+				</Function>
+			</Bay>
+		</VoltageLevel>
+	</Substation>
+	<DataTypeTemplates>
```
dlabordus, line 45: Is DataTypeTemplates needed to test the transformer-winding-editor?

Makes the file even smaller if not needed.

test/unit/editors/substation/transformer-winding-editor.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  let element: TransformerWindingEditor;
+  let doc: XMLDocument;
+
+  beforeEach(async () => {
+    doc = await fetch('/test/testfiles/TransformerWinding.scd')
+      .then(response => response.text())
+      .then(str => new DOMParser().parseFromString(str, 'application/xml'));
+  });
+
+  it('looks like the latest snapshot', async () => {
```
dlabordus, line 16: Why do we have this snapshot test? It looks the same as the one without children.

dlabordus (changes requested):

src/editors/substation/transformer-winding-editor.ts
juancho0202: This is already checked in the only possible parent PowerTransformer 
image

test/testfiles/TransformerWinding.scd
juancho0202: Removed ☑️

test/unit/editors/substation/transformer-winding-editor.test.ts
juancho0202: i don't think they look the same
image

src/foundation.ts
juancho0202: Removed ☑️

src/editors/substation/transformer-winding-editor.ts
dlabordus: But don't you want to send 'true' to the editor 'transformer-winding-editor'.
So that 'transformer-winding-editor' can send it to 'eq-function-editor', because the default will be false.

test/unit/editors/substation/transformer-winding-editor.test.ts
dlabordus: Look good at the snapshot

compare 
  snapshots["transformer-winding-editor looks like the latest snapshot"] =
and 
  snapshots["transformer-winding-editor without children looks like the latest snapshot"] =

juancho0202: Oh yes that one actually does, I will remove snapshots["transformer-winding-editor looks like the latest snapshot"] ☑️

src/editors/substation/transformer-winding-editor.ts
juancho0202: Passed `showFunctions` to <transformer-winding-editor> and subsequently to <eq-function-editor> ☑️

test/unit/editors/substation/transformer-winding-editor.test.ts
```diff
@@ -0,0 +1,41 @@
+import { fixture, html, expect } from '@open-wc/testing';
```
dlabordus, line 1: Sorry, for the next comment.

Because there is now a showfunctions boolean, I think there is a extra snapshot needed.
The with children now can be shown in 2 modes, showfunctions true and false.
You now only have the false situation, because the eq-function-editor is missing.

juancho0202: added missing test and updated the snapshot file ☑️

dlabordus (approved):

PR #1075, opened 2022-11-10
fix(menu/importieds) Import IEDs with Communication section if IED named TEMPLATE

### LINKED ISSUE #1074: No import of Communication section with IEDs named TEMPLATE
**Describe the bug**

Currently, importing an ICD file with an iedName of TEMPLATE results in the Communication section not being imported.

I don't think this is what was intended, an ICD can and should include Communication sections which should be imported.

**To Reproduce**

1. Import from an ICD file with an IED named "TEMPLATE" with a configured Communication section.
2. Examine the Communication section
3. Find it is unchanged.

**Expected behavior**

The communication section should be imported irrespective of the IED name.

**Additional context**

I think this occurred as a side-effect of avoiding duplicated device names:

What we have at present is the following adjustment of the IED name if there is already devices named 'TEMPLATE'

https://github.com/openscd/open-scd/blob/42fddb87b83733447293e1879b531a7e16429911/src/menu/ImportIEDs.ts#L371-L385

But this doesn't update the ConnectedAP iedName attributes and so the "new de-duplicated IED name" is used in this part of the code which, unsurprisingly doesn't locate any ConnectedAPs

https://github.com/openscd/open-scd/blob/42fddb87b83733447293e1879b531a7e16429911/src/menu/ImportIEDs.ts#L82-L108

I'll be happy to work on this if it is agreed as a bug.

danyill:
Closes #1074

### DIFF
--- a/src/menu/ImportIEDs.ts
+++ b/src/menu/ImportIEDs.ts
@@ -373,15 +373,22 @@ export async function importIED(
   doc: XMLDocument,
   dispatchObject: HTMLElement
 ): Promise<void> {
-  if (ied.getAttribute('name') === 'TEMPLATE')
-    ied.setAttribute(
-      'name',
+  if (ied.getAttribute('name') === 'TEMPLATE') {
+    const newIedName =
       'TEMPLATE_IED' +
-        (Array.from(doc.querySelectorAll('IED')).filter(ied =>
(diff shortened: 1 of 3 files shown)


JakobVogelsang (approved):
LGTM

PR #1077, opened 2022-11-13
fix(filtered-list) Filter when document changes

### LINKED ISSUE #1018: GOOSE Subscription filtering and options after removal don't appear correct
**Describe the bug**

In the Subscriber Later Binding (GOOSE) plugin, I get some behaviour that is unexpected with the attached file:

[XAT_Prot1.scd.zip](https://github.com/openscd/open-scd/files/9641674/XAT_Prot1.scd.zip)

1. I filter on the left with `232_P1 OpCls.ge` and on the right with `232_M1`. This works fine.
    image

2. Then I click on the subscribed value. I expect it to be removed from "Subscribed" and added to "Available to subscribe"

3. Instead, what occurs is:

    image

4. What is odd:

    * After the removal the filter shows `242_M1`, not `232_M1`
    * `232_M1` is missing
    * In fact, neither of the "available to subscribe" appear to match the filter.

**To Reproduce**

1. Open the attached scd file.
2. Go to Subscriber Later Binding (GOOSE) plugin
3. Enter `232_P1 OpCls.ge` on the left and on the right with `232_M1`.
4. Remove the existing subscription
5. See error described above.

**Expected behavior**

The filter should refresh and show the one available option and the one new option due to removing the subscription.

danyill:
Closes #1018 

When a document is changed the filtered list doesn't receive a new filtering action, and its internal state appears to become inconsistent.

Only a couple of lines change for the new updated method. This seems like the right place to do this but I am interested and would be grateful for a review.

I noticed an incorrect import which seems to have been added by me and there is a few formatting changes.

I think this is an issue in the Cleanup plugin and probably other places (everywhere where a document update on a filtered list occurs and the filtered list has filter text).

I can't see how to add tests for this so I have not.

### DIFF
--- a/src/filtered-list.ts
+++ b/src/filtered-list.ts
@@ -18,7 +18,6 @@ import { List } from '@material/mwc-list';
 import { ListBase } from '@material/mwc-list/mwc-list-base';
 import { ListItemBase } from '@material/mwc-list/mwc-list-item-base';
 import { TextField } from '@material/mwc-textfield';
-import { ReportControlElementEditor } from './editors/publisher/report-control-element-editor';
 
 function slotItem(item: Element): Element {
   if (!item.closest('filtered-list') || !item.parentElement) return item;
@@ -46,14 +45,14 @@ function hideFiltered(item: ListItemBase, searchText: string): void {
     .split(/\s+/g);
 
   (terms.length === 1 && terms[0] === '') ||
-  terms.every(term => {
-    // regexp escape
-    const reTerm = new RegExp(
-      `*${term}*`.replace(/\*/g, '.*').replace(/\?/g, '.{1}'),
-      'i'
-    );
-    return reTerm.test(filterTarget);
-  })
+    terms.every(term => {
+      // regexp escape
+      const reTerm = new RegExp(
+        `*${term}*`.replace(/\*/g, '.*').replace(/\?/g, '.{1}'),
+        'i'
+      );
+      return reTerm.test(filterTarget);
+    })
     ? slotItem(item).classList.remove('hidden')
     : slotItem(item).classList.add('hidden');
 }
@@ -115,6 +114,12 @@ export class FilteredList extends ListBase {
     this.requestUpdate();
   }
 
+  public layout(): void {
+    super.layout();
+    // regenerate filtering of text
+    this.onFilterInput();
+  }
+
   constructor() {
     super();
     this.addEventListener('selected', () => {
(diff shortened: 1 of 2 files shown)


danyill:
The snapshot test failures appear unrelated to this PR as far as I can tell, they all involve list items in the publisher editor with "activatable" not being included in the latest snapshot.

JakobVogelsang (commented):
The formatting issues seem odd to me. But the fix itself is straight forward. Thanks for this.
And thx for removing the strange import!!

JakobVogelsang:
> The snapshot test failures appear unrelated to this PR as far as I can tell, they all involve list items in the publisher editor with "activatable" not being included in the latest snapshot.

After a second look, the tests work perfectly fine. The publisher editor has activatable list. The updated function remove the activated attribute from the list item. This happens in all subscriber plugins, by the way. There are just simply not snapshots taken of the left hand side list after it has been clicked. 

I try to see how this can be fixed.

danyill:
So the present state of play is that the moment even a blank `updated` function causes `activatable` to be lost with this PR.

```js
  public updated(): void {
    // nothing to see here
  }
```

What I have done:

* a rebase on main.
* added a test to ensure we see the incorrect behaviour rather than through an indirect snapshot.
* tried updating mwc-list to 0.27.0 to see if the problem is caused/resolved upstream. It is not.
* spent quite some time getting a little lost in the debugger

I have noticed that I can fix this for any particular filtered list by doing the following:

Setting a breakpoint on L#216 in `@material.mwc-list.mwc-list-base.v0.22.1.js` in sources in the browser debugger (at this URL: `http://localhost:8080/_snowpack/pkg/@material.mwc-list.mwc-list-base.v0.22.1.js`)

When I break at  this line, I execute the following console command:

```js
this.mdcFoundation.setUseActivatedClass(true);
```

This can be also applied on L#171 after `firstUpdated` is called.

I think this implies that the cause is related to construction/initialisation rather than reactive updates.

What is happening seems to be that the existence of the `updated` method is causing `activatable` to be lost.

We consider our problem a little more deeply.

I think the problem is caused by not calling the superclass update method, the following is a solution.

```js
  protected update(
    changedProperties: Map<string | number | symbol, unknown>
  ): void {
    super.update(changedProperties);
    // regenerate filtering of text
    this.onFilterInput();
  }
```

I have verified that this resolves the original issue. I'm not sure why this idea didn't occur earlier :wink:

JakobVogelsang (approved):
LGMT

JakobVogelsang:
Well done, thank you. It totally makes sense, especially looking at the Lit documentation. I was not able to see it :)

danyill:
Thank you for this review. :+1: 
Yes, I think the clue in the lit docs was probably the words in bold, "All lifecycle methods need to call the super method."
:wink: 

I've rebased this, but somehow I have also dismissed your review and no longer have approval, if you could click through I'll merge it.

JakobVogelsang (approved):
LGTM

I have added a new rule, that forces us to review if anything about the PR changes. The merge of main into this PR triggered it in this case.

PR #1078, opened 2022-11-14
fix(subscription) Coalesce lack of prefix to empty string, allow no prefix in FCDA description

### LINKED ISSUE #1076: FCDA secondary text includes the word null
**Describe the bug**

I noticed in the attached file the word "null" appears and probably shouldn't.

**To Reproduce**

[XAT_Prot1.scd.zip](https://github.com/openscd/open-scd/files/9996724/XAT_Prot1.scd.zip)

1. Open the file.
2. Look at the B30_Coupler_Commands message and notice the word null
    image
3. I suspect this is the prefix which may be optional.

**Expected behavior**

If the prefix isn't present it should not be shown.

On openscd.github.io

danyill:
Closes #1076

The code I am touching was only added recently so the reason may be nearby

I think `ldInst` is almost always part of an FCDA (GSSE being mostly dead/unused) but prefix maybe not, it is optional in Table 22 of IEC 61850-6 Ed 2.1

So I don't fully understand this code path but this should get rid of the word null.

### DIFF
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -1,4 +1,6 @@
-import { css, LitElement, query } from 'lit-element';
+import { css, html, LitElement, query } from 'lit-element';
+import { nothing } from 'lit-html';
+
 import {
   cloneElement,
   compareNames,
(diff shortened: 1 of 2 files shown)


src/editors/subscription/foundation.ts
```diff
@@ -99,10 +99,10 @@ export function getFcdaTitleValue(fcdaElement: Element): string {
 
 export function getFcdaSubtitleValue(fcdaElement: Element): string {
   return `${fcdaElement.getAttribute('ldInst')} ${
-    fcdaElement.hasAttribute('ldInst') && fcdaElement.hasAttribute('prefix')
+    fcdaElement.hasAttribute('ldInst')
       ? `/`
       : ''
-  } ${fcdaElement.getAttribute('prefix')} ${fcdaElement.getAttribute(
+  } ${fcdaElement.getAttribute('prefix') ?? ''} ${fcdaElement.getAttribute(
```
dlabordus, line 105: To make it even better to get ride of the double spaces the code could be 
```
  fcdaElement.hasAttribute('ldInst')
      ? `/`
      : ''
  }${fcdaElement.getAttribute('prefix') ? html ` ${fcdaElement.getAttribute('prefix')` ? nothing} ${fcdaElement.getAttribute( 
    'lnClass'
  )} ${fcdaElement.getAttribute('lnInst')}`;
}
```

danyill: I have done similar to what you have suggested, thanks for the review.

dlabordus (approved):

danyill:
Thank you for this review :clap:

PR #1079, opened 2022-11-14
feat(editors/substation): redirect LNode's on clone

### LINKED ISSUE #1053: Redirect `LNode` on cloning `Bay`, `VoltageLevel` and `Substation`
As a user of OpenSCD I have to design the same bay with little differences over and over again and want OpenSCD to give me an efficient process to do so.

**Requirements**
1. extend the current clone behavior to incorporate copying of `LNode` elements
>NOTE:  as `LNode` reference is unique within the whole SCL a plane copy is not possible instead we can replace with `LNode` to the same data in another `IED`
- take over all `LNode` elements within the to be copied entity (`Bay`, `VoltageLevel` or `Substation`)
- do no change any `LNode` element that has `iedName=None`
- for each `LNode` where `iedName` is not `None` get the IED to be replaced with from the UI (see mock)
   - If `none` is selected (see Mock), remove all `LNode`
   - If not `none` is selected e.g.`OhterIED2` check whether the reference is valid in the new IED and replace the `iedName` attribute in the `LNode` in case the reference is valid. Otherwise, remove the `LNode`. Valid means that the reference as a combination of IED LDevice and logical node is pointing to a valid reference
- for each IED print on select all per default `none`
- if there are no `LNode` elements in the to be copied entity or all `LNode` are `iedName=None` to no show the UI

Mockup:
grafik

JakobVogelsang:
Close #1053


test/unit/editors/substation/redirectionUI.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  redirectUI
+    .querySelector<HTMLElement>('mwc-button[icon="content_copy"]')
+    ?.click();
+  await redirectUI.updateComplete;
+}
+
+describe('Clone Redirection UI', () => {
+  let element: BayEditor;
```
dlabordus, line 55: Don't we want these snapshot tests in the different editors and not only with the BayEditor?

test/unit/editors/substation/__snapshots__/redirectionUI.test.snap.js
```diff
@@ -0,0 +1,195 @@
+/* @web/test-runner snapshot v1 */
```
dlabordus, line 1: It contains 2 snapshots, but I can only find one case having a snapshot.

JakobVogelsang: Thx. That was a test rename that added the second snapshot. In those cases, a new snapshot must be generated. I fixed it

test/unit/editors/substation/redirectionUI.test.ts
JakobVogelsang: Nice seen. I did trigger the redirect UI from all the three editors. Thx

dlabordus (approved):

PR #1080, opened 2022-11-14
Copy namespace and all attributes between documents before copying IED

danyill:
Please never merge this! For discussion only.

### DIFF
--- a/src/menu/ImportIEDs.ts
+++ b/src/menu/ImportIEDs.ts
@@ -368,6 +368,14 @@ function isIedNameUnique(ied: Element, doc: Document): boolean {
   return true;
 }
 
+function cloneAttributes(destElement: Element, sourceElement: Element) {
+  let attr;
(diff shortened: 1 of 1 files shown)


danyill:
Very hacky but I tried to use `importNode` in 7820235a8e523242640e0caad732d50303fc6f57 but without success.

danyill:
OK, a little progress, it may be that to get the DOM to recognise the namespaces we need to be more explicit, something like this seems to help:

```js
this.doc.documentElement.setAttributeNS('http://www.w3.org/2000/xmlns/', 'xmlns:esel', 'http://www.selinc.com/2006/61850');
```

 I think it is insufficient to copy the attributes associated with the namespace to the document element without explicitly handling the prefixes.

danyill:
This has served it's purpose and I have a way forward thanks to discussion with ca-d

PR #1101, opened 2022-12-05
feat(wizard/generalequipment): Add_create_wizard_GeneralEquipment

marcvanraalte:


JakobVogelsang (changes requested):
You are removing test files. That should not happen

PR #1102, opened 2022-12-06
feat(wizards/generalequipment): Add_create_wizard_GeneralEquipment_and_test

marcvanraalte:


test/unit/wizards/generalequipment.test.ts
```diff
@@ -108,25 +213,6 @@ describe('Wizards for SCL GeneralEquipment element', () => {
       expect(actionEvent).to.not.have.been.called;
     });
 
-    it('does not accept invalid type attribute', async () => {
```
JakobVogelsang, line 111: Why do you touch these lines of code?

JakobVogelsang (changes requested):
Some naming is inconsistent and need to be changed

test/unit/wizards/generalequipment.test.ts
JakobVogelsang: This removal is still in the code. Why are those tests removed. They seem to be reasonable?

JakobVogelsang: It is always an alarm for me that tests are removed for no reason. This is such an alarm. Why are two tests that make perfect sense being removed?

JakobVogelsang (approved):
LGTM

Issue #1052, opened 2022-10-26
Import IEDs menu item fails to import multiple selected IEDs


danyill:
**Describe the bug**

I try to import multiple merging units on top of an existing file.
Only the first one is imported, possibly not even correctly but there are no obvious errors or warnings.

[Prot1_SCD.zip](https://github.com/openscd/open-scd/files/9868382/Prot1_SCD.zip)

**To Reproduce**

With reference to the files in the zip file.

1. Open the scd file, 'first.scd'
2. Click on 'Import IEDs'.
3. Enter M1 in the filter field, select all IEDs and go import.
    ![image](https://user-images.githubusercontent.com/674783/197997382-17117d82-ca3e-43ab-beb7-58034adf0702.png)
4. Notice only one IED is added (`XAT_T1_M1`).

Note the file is initially schema invalid because the SubNetwork is empty and is schema valid after an IED is imported.

**Expected behavior**

All IEDs should be added.

**System**

 - OS: Ubuntu (also tested on Windows Server 2016)
 - Browser [Firefox (also tested on Chrome)
 - Using openscd.github.io

cfyi @TPChenDNZ

JakobVogelsang:
That's on my sorry. I did a fix here some time ago and obviously the tests were not sufficient

PR #1103, opened 2022-12-06
fix(ieds-import): multiple IEDs import

JakobVogelsang:
This PR is primarily fixing #1052. However, I invested also so time to remove the dependency from wizard-dialog. 
@danyill would you be so kind and test it with some files. I would not want to break, repair it again. :)

### DIFF
--- a/src/menu/ImportIEDs.ts
+++ b/src/menu/ImportIEDs.ts
@@ -1,75 +1,75 @@
-import { css, html, LitElement, query, TemplateResult } from 'lit-element';
-import { get } from 'lit-translate';
+import {
+  css,
+  html,
+  LitElement,
+  property,
+  query,
+  state,
+  TemplateResult,
+} from 'lit-element';
+import { get, translate } from 'lit-translate';
 
 import '@material/mwc-list/mwc-check-list-item';
+import '@material/dialog';
+import '@material/mwc-button';
+import { Dialog } from '@material/mwc-dialog';
 import { List } from '@material/mwc-list';
 import { ListItemBase } from '@material/mwc-list/mwc-list-item-base';
 
 import '../filtered-list.js';
 import {
   createElement,
-  EditorAction,
   identity,
+  isPublic,
   newActionEvent,
   newLogEvent,
   newPendingStateEvent,
-  newWizardEvent,
   selector,
   SimpleAction,
-  Wizard,
-  WizardActor,
 } from '../foundation.js';
 
-function importIedsAction(
-  importDoc: XMLDocument,
-  doc: XMLDocument
-): WizardActor {
-  return (_, wizard: Element): EditorAction[] => {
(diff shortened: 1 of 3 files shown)


src/menu/ImportIEDs.ts
```diff
@@ -512,28 +497,74 @@ export default class ImportingIedPlugin extends LitElement {
...
+    const mergedPromise = new Promise<void>((resolve, reject) =>
+      Promise.allSettled(promises).then(
+        () => resolve(),
+        () => reject()
+      )
+    );
 
-  firstUpdated(): void {
-    this.parent = this.parentElement!;
+    this.dispatchEvent(newPendingStateEvent(mergedPromise));
```
danyill, line 529: This line appears to be the cause of the console error. I seem to be able to import IEDs without it quite cheerfully.

danyill (commented):
L(A)GTM, a typo and something in the events I cannot see does not appear entirely right.

src/menu/ImportIEDs.ts
JakobVogelsang: Thank you very much. This comment is exactly why I love the bazaar style of open source projects. :D

JakobVogelsang: I forgot that the `run` method is async. Which means that I did not touch this line of code. It is working for me right now. @danyill would you be so kind and try it out on your machine?

src/menu/ImportIEDs.ts
```diff
@@ -368,120 +368,105 @@ function isIedNameUnique(ied: Element, doc: Document): boolean {
...
-      'TEMPLATE_IED' +
-      (Array.from(doc.querySelectorAll('IED')).filter(ied =>
-        ied.getAttribute('name')?.includes('TEMPLATE')
-      ).length +
-        1);
-
-    ied.setAttribute('name', newIedName);
+export default class ImportingIedPlugin extends LitElement {
+  @property({ attribute: false })
+  doc!: XMLDocument;
```
danyill, line 379: Maybe we don't want to pass in `doc` here?

If I try and import  from a large scd file with 10-20 IEDs I see my dialog updating as each is added to the project. This seems a little new. This may be one of the rare use cases where we don't want to update the UI as the project changes.

danyill:
It's looking really good, thank you for this improvement :love_letter:. One more comment from me about the handling of `doc`, and we're all good.

Thank you also for changing the naming convention for importing TEMPLATE IEDs :+1:

danyill (approved):

src/menu/ImportIEDs.ts
JakobVogelsang: Hmm true I will change that. It feels weird as is at the moment.

JakobVogelsang: Fixed. Thank you for the hint

PR #1107, opened 2022-12-12
feat(general-equipment-editor): add remove button

marcvanraalte:


src/editors/substation/general-equipment-editor.ts
```diff
@@ -94,11 +116,21 @@ export class GeneralEquipmentEditor extends LitElement {
           @click=${() => this.openEditWizard()}
         ></mwc-icon-button>
       </abbr>
+      <abbr slot="action" title="${translate('remove')}">
+        <mwc-icon-button
```
JakobVogelsang, line 120: Using `action-icon` `mwc-icon-button` looks strange. That is why all other editor use `mwc-fab` instead.
Same is true for the edit button. Sorry that slipped through the review

JakobVogelsang (changes requested):

src/editors/substation/general-equipment-editor.ts
pascalwilbrink: It looks like they're being mixed. sometimes mwc-icon-button, sometimes mwc-fab.
If you look at the function-editor (for example), that one is using mwc-icon button.
the powertransformer-editor (for example) is using the mwc-fab.

@JakobVogelsang, we can change it to mwc-fab for now, but maybe it's better to change it everywhere?

pascalwilbrink: We found out what you meant with action-icon > mwc-icon-button :) changed it in the code. should look good now

JakobVogelsang: Nice. Thx

JakobVogelsang (approved):
LGTM

PR #1110, opened 2022-12-15
feat(substation/general-equipment-editor):add_button_general-equipment_container

marcvanraalte:


JakobVogelsang (changes requested):
You forgot to set the anchor for the menu. This is why is pushed far to the left and is outside of the panes. Like so

```ts
firstUpdated(): void {
    if (this.addMenu && this.addButton)
      this.addMenu.anchor = <HTMLElement>this.addButton;
  }
```

JakobVogelsang (approved):
LGTM

PR #1111, opened 2023-01-03
fix: close menu after open project

Stef3st:


Stef3st:
@juancho0202 That approach is better! I implemented it in the last commit. Could you review it?

juancho0202 (approved):

PR #1124, opened 2023-01-10
fix(wizards/connectedap) Scope Address for ConnectedAP more narrowly, closes #1123

### LINKED ISSUE #1123: ConnectedAP wizard incorrectly populates pTypes from descendant <Address> nodes
**Describe the bug**
When editing a ConnectedAP using the Communication editor, the `VLAN-ID` and `VLAN-PRIORITY` pTypes are populated with values from a <GSE> node, even when no pTypes are under the ConnectedAP `<Address>` node.

**To Reproduce**
1. Open a project with an IED that has a ConnectedAP containing a GSE/SMV that has `VLAN-ID` and `VLAN-PRIORITY` fields populated, but not within the ConnectedAP <Address>.
2. Click on `Communication`
3. Click on the ConnectedAP within the IED
4. Click edit
5. Scroll down to VLAN-ID and VLAN-PRIORITY
6. See those fields populated with values from the first GSE <Address>
7. Open the SCL xml editor and confirm the ConnectedAP `<Address>` doesn't contain any `VLAN-ID` or `VLAN-PRIORITY` pTypes.

**Expected behavior**
When a ConnectedAP is edited that doesn't contain pTypes, it should not use populated pTypes from descendant GSE/SMV `<Address>` nodes.

**Screenshots**
image

image

 - OS: Windows 10
 - Browser: Google Chrome
 - Version: 108.0.5359.125 (Official Build) (64-bit)

**Additional context**
I suspect the issue is to do with the `createPTextField` function within the connectedAP wizard. The function at present grabs the first pType under an `<Address>` child from all descendants under the ConnectedAP.

https://github.com/openscd/open-scd/blob/2d1464fa3446d459aa411f82080241ebf0ff379a/src/wizards/connectedap.ts#L325-L338

One way to avoid this is the following:

```ts
.maybeValue=${Array.from(element.querySelectorAll(`Address > P[type="${pType}"]`)).filter(e => e.parentElement.parentElement === element)[0]?.innerHTML ?? null}
```

However this probably doesn't work for some other cases...

Here is a test scd file which you can use to reproduce the issue.

[connectedap_test.zip](https://github.com/openscd/open-scd/files/10378593/connectedap_test.zip)

danyill:


JakobVogelsang (approved):
LGTM

danyill:
Thank you for the review.

PR #1130, opened 2023-01-13
fix: styling issue with SubEquipment editor

### LINKED ISSUE #1129: Substation Editor doesn't handle SubEquipment Functions and SubFunctions element overflow
**Describe the bug**
When SubFunctions are contained within SubEquipment elements within the `<Substation>` section, OpenSCD does not manage the overflow of content within the element when there are a large number of bays within a voltage level.
This results in icons appear outside of the elements boundary, and obstructs editor menus.

**To Reproduce**
1. Open a scl file that contains SubEquipment with Function and SubFunction elements, and at least one LNode.
2. Click on Substation
3. Scroll down to a SubEquipment with a Function and a SubFunction.
4. Observe malformed elements.

**Expected behavior**
The Substation editor should be able to present all legal elements within a `<Substation>` section without obscuring content, buttons or menus.

**Screenshots**

image

 - OS: Windows 10
 - Browser Google Chrome
 - Version 108.0.5359.126 (Official Build) (64-bit)

**Additional context**

Attached is a ssd which reproduces the issue.

[XAT.zip](https://github.com/openscd/open-scd/files/10407268/XAT.zip)

JakobVogelsang:
Closes #1129 

Two things that have been fixed
1. by mistake, the container for sub-equipment-editors was declared min size 64px. That is why it looked too strange. This declaration is only necessary for icon type components. `SubEquipment` is never rendered as icon.
2. sub-equipment-editor do not need to be within a container. That is only necessary when grid type display is used. E.g. Bay within VoltageLevel or ConductingEquipment as icon within Bay. In all other cases, just adding it as child editor is enough. `action-pane` will do all the styling.

There is no regression test as I only touched CSS. Just updated snapshots due to the changes html

### DIFF
--- a/src/editors/substation/conducting-equipment-editor.ts
+++ b/src/editors/substation/conducting-equipment-editor.ts
@@ -21,6 +21,7 @@ import '../../action-icon.js';
 import '../../action-pane.js';
 import './eq-function-editor.js';
 import './l-node-editor.js';
+import './sub-equipment-editor.js';
 import { startMove, getIcon, styles } from './foundation.js';
 import {
   getChildElementsByTagName,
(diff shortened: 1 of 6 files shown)


danyill (approved):
LGTM

PR #1136, opened 2023-01-14
fix: add missing code editor to edit wizards

### LINKED ISSUE #1133: Provide edit button for EqFunction and EqSubFunction

In Pro mode, we have an ace editor edit button for most elements in the Substation editor.
Notably, we don't have this for EqFunction and EqSubFunctions

I'd just like that button because it delights me to quickly jump into the SCL without having to go to a parent container which has the button.

So a button in the top right corner:

image

Like for other containers:

image

JakobVogelsang:
Closes #1133 

No test adoption needed as we do not screenshot test wizards. This is once more am argument to do that when we move plungins.


danyill (approved):
LGTM, thanks!

PR #1137, opened 2023-01-16
feat(wizards/transformerwinding): add edit wizard

marcvanraalte:
edit wizard and tests added


src/wizards/transformerWinding.ts
```diff
@@ -0,0 +1,114 @@
...
+  cloneElement,
+  getChildElementsByTagName,
+  getValue,
+  SimpleAction,
+  Wizard,
+  WizardActor,
+  WizardInputElement,
+} from '../foundation';
+
+export function editTransformerWindingWizard(element: Element): Wizard {
```
JakobVogelsang, line 13: Functions within OpenSCD are sorted so that the function with no dependencies is further down below and the functions that it depends on are above. In this case, `editTransfomerWinding` uses `updateTransformerWindingAction` and `contentTransformerWindingWizard`. They therefore should be allocated above `editTransformerWinding`

src/wizards/transformerWinding.ts
```diff
@@ -0,0 +1,114 @@
...
+import { get, translate } from 'lit-translate';
+import {
+  cloneElement,
+  getChildElementsByTagName,
+  getValue,
+  SimpleAction,
+  Wizard,
+  WizardActor,
+  WizardInputElement,
+} from '../foundation';
```
JakobVogelsang, line 11: it should be `foundation.js`

src/wizards/transformerWinding.ts
```diff
@@ -0,0 +1,114 @@
+import { html, TemplateResult } from 'lit-element';
+import { get, translate } from 'lit-translate';
+import {
```
JakobVogelsang, line 3: Please add an empty line between two different kinds of import. It makes import statements more readable

src/wizards/transformerWinding.ts
```diff
@@ -0,0 +1,114 @@
...
+export function contentTransformerWindingWizard(
+  content: ContentOptions
+): TemplateResult[] {
+  return [
+    html`<wizard-textfield
+      label="name"
+      .maybeValue=${content.name}
+      helper="${translate('scl.name')}"
+      required
+      minLength="${1}"
```
JakobVogelsang, line 90: `minLenght` is not necessary. This is already covered with `required`. This property does not allow an empty string as inoput

test/testfiles/editors/substation/VS869-TransformerWinding.scd
```diff
@@ -0,0 +1,252 @@
+<SCL xmlns="http://www.iec.ch/61850/2003/SCL" version="2007" revision="B" release="4">
```
JakobVogelsang, line 1: This test file has already been included. I am in favor to move it if you want to, but please delete the other one and correct the other tests then.

JakobVogelsang (changes requested):

JakobVogelsang (approved):
LGTM

PR #1141, opened 2023-01-24
feat(editors/later-binding): Improve supervision visibility and remove clutter

### LINKED ISSUE #1024: Reduce clutter in later binding editor

As a user of OpenSCD I find the later binding information display a bit cluttered or confusing. Also, when I subscribe to an item the item I clicked on changes and the secondary field is quite different.

I'd like to propose that we remove the information about the subscribing IED from the later binding view.

Currently, it repeats the information available on the left hand pane, tends to be cut off by an ellipsis on a large form factor device and feels visually cluttered. If it is retained I think it must be on a second line or have some iconography to help make it clearer.

See the screenshot:

image

`MUSV I01 TCTR1 AmpSv instMag.i` repeats the left hand pane.
`@1-C:MUGO/GOINGGIO1.SPC` repeats the major name.

Additionally when I look at an element before it is subscribed:

image

And after it is subscribed:

image

I think these should remain the same and that the first is clearer. The user is moving an "available to subscribe" into "subscribed", ideally the display wouldn't change too much.

I can do a PR for this for the later binding SMV and GOOSE plugins if this is acceptable after refinement. Perhaps I don't understand or it is there for another reason.

The other reason I want to reduce the clutter is that I'd like to add some in :stuck_out_tongue_winking_eye: I think we need to show the LN for LGOS and LSVS here at some stage.

danyill:
Closes #1024, #1037


test/unit/editors/subscription/later-binding/__snapshots__/ext-ref-later-binding-list.test.snap.js
```diff
@@ -76,7 +76,7 @@ snapshots["extref-later-binding-list for Sampled Value Control when SVC has no s
                  (MeasPoint.CT2)
       </span>
       <span slot="secondary">
-        SMV_Subscriber>>Overvoltage> PTRC 1>AmpSv;TCTR2/AmpSv/instMag.i[0]
+        SMV_Subscriber>>Overvoltage> PTRC 1
```
JakobVogelsang, line 79: I do not see a snapshot showing subscription supervision. Or am I missing something?

JakobVogelsang (approved):
LGTM

test/unit/editors/subscription/later-binding/__snapshots__/ext-ref-later-binding-list.test.snap.js
danyill: Thanks for spotting, that is intriguing... I will investigate. I don't see it in #1082 either (except for message binding).

danyill: I have added some snapshots

danyill:
After some faffing I think the snapshots are correct.

JakobVogelsang (approved):
LGTM

PR #1149, opened 2023-01-30
feat(editors/binding): Allow filtering of subscribed/unsubscribed data in binding editors

### LINKED ISSUE #1062: Allow filtering in later binding to "only configured" data

As a user of OpenSCD when looking at a fully configured system I would like the ability to only show configured FCDA for GOOSE and SV. This reduces the clutter and makes the navigation simpler.

This is useful for designers and technicians to look at a configured system when the main question asked is "How is this system configured?", not "How do I create an additional mapping.

The ability for the user to select whether or not unconfigured FCDAs are shown.

Perhaps provide in the later binding plugins a "3 dots" menu with more options to allow "Configured" and "Unconfigured" checkboxes to be shown. If one is only adding to an existing scheme, seeing existing information may be unnecessary (?).

Perhaps not this shade of purple, but something like this:

image

danyill:
Closes #1062

Draft until we agree if _any_ of this is a good idea, and I will add some tests.

I'm not comfortable with the filter button being somewhat shaded when the section is selected, and the background colour darkens.  I have tried to deal with it as best I can, but perhaps there is a better way.

It also doesn't entirely play nicely when "Not Subscribed" is selected and the user does subscriptions (the activated item becomes hidden). Maybe we don't need this option?


src/editors/subscription/fcda-binding-list.ts
```diff
@@ -32,6 +37,7 @@ import {
   SubscriptionChangedEvent,
 } from './foundation.js';
 import { getSubscribedExtRefElements } from './later-binding/foundation.js';
+import { classMap } from 'lit-html/directives/class-map.js';
```
JakobVogelsang, line 40: This statement is usually below the one importing from lit

src/editors/subscription/fcda-binding-list.ts
```diff
@@ -59,6 +65,14 @@ export class FcdaBindingList extends LitElement {
   @state()
   private extRefCounters = new Map();
 
+  @query('.actions-menu') actionsMenu!: Menu;
+  @query('.actions-menu-icon') actionsMenuIcon!: Icon;
+
+  @property({ attribute: false })
```
JakobVogelsang, line 71: It is common to put `query` directives below `properties`

JakobVogelsang: Does it really have to be a `property` or is that rather a `state` used only within the component?

src/editors/subscription/fcda-binding-list.ts
```diff
@@ -186,14 +208,76 @@ export class FcdaBindingList extends LitElement {
...
   }
 
+  protected firstUpdated(): void {
+    this.actionsMenu.anchor = <HTMLElement>this.actionsMenuIcon;
+
+    this.actionsMenu.addEventListener('closed', () => {
+      this.filterSubscribed = (<Set<number>>this.actionsMenu.index).has(0);
+      this.filterNotSubscribed = (<Set<number>>this.actionsMenu.index).has(1);
+
+      // reset filters if all selected = no filtering
```
JakobVogelsang, line 218: This is a bit strange to me. It is primarily so, because I see as a user that the selection is dropped. I would rather expect something like this:
-  both are enabled by default
- you can disable something you do not want to see 
- if the user disables both, that is predictable behavior

danyill: I agree, I suggest having a single check box for "Show only configured" which is what I tried to do in the subscriber-oriented view.

I'd also write this state to local storage.

If you're happy that the basic approach is acceptable I shall improve this PR a bit and move it out of draft in the next few  days.

JakobVogelsang (dismissed):
The code looks good. There is only one thing that I would want to be changed:
- On my Mac when I select the first item the button has a different color compared to selecting the second choice. 

Another one thing is debatable and I leave it up to you:
- when no FCDA is viable also do not show the control block. With many control blocks in the file this effect of this filter is lost. This could be possible with simple CSS rule I feel

src/editors/subscription/fcda-binding-list.ts
JakobVogelsang: I am in general. The feature is definatly a very very useful one.

JakobVogelsang:
Thank you @danyill for your wonderful contribution to that thing that we are all working on.

src/editors/subscription/fcda-binding-list.ts
danyill: > This is a bit strange to me. It is primarily so, because I see as a user that the selection is dropped. I would rather expect something like this:

It is a bit strange, I am sorry. Here is why it happened:

- The filter turns things off that you would normally see. 
- But I don't want the checkboxes to be on by default because the filter should not be "on by default" to keep the UI simple.
- I change the styling to show that a filter checkbox has been checked and I want this not to be highlighted initially.
- This then causes the difficulty of allowing the disable of both which means the user has intentionally selected "nothing" which is not so much a valid use case.

I will implement as you suggest and see how it feels.

danyill:
> * On my Mac when I select the first item the button has a different color compared to selecting the second choice.

I don't see that - can you provide a screenshot?

Here is what I see: 

image

> * when no FCDA is viable also do not show the control block. With many control blocks in the file this effect of this filter is lost. This could be possible with simple CSS rule I feel

Yes, good suggestion - let's do that.

src/editors/subscription/fcda-binding-list.ts
danyill: It could be a property -- if the plugin were embedded or used elsewhere, a user might want to set this on the plugin?
I am happy to change if you wish.

src/editors/subscription/fcda-binding-list.ts
danyill: It feels good.

danyill:
> when no FCDA is viable also do not show the control block. With many control blocks in the file this effect of this filter is lost. This could be possible with simple CSS rule I feel

 The CSS is kind of simple but for my weak mind it was not so simple...

I take a moment to pity my reviewer. :pray: 

This table may help my reviewer! It helped me...

image

OTOH although it seems complex to me it is blindingly fast which I like. :heart_on_fire:

danyill:
OK, after making the review changes I am much happier with the outcome.

I have implemented local storage so the filter selection isn't ephemeral.
Part of using a boolean attribute means I must default to false so the states must be inverted which can be confusing to think about.

I use accessors for holding the filter state so I can use other properties to create a unique identifier for local storage.

danyill:
Thank you for these review comments, the product is greatly improved. :tada: 

This is ready for some more testing and review. 

If we are happy, then I can add some brief integration tests for the four plugins where it is used.

I'll also do something similar to this in #1187 for the subscriber-oriented view.

danyill:
I have tried to add some initial tests, but they are not working - and so far I cannot see why. Could I ask you to take a look at these please?

danyill:
> I have tried to add some initial tests, but they are not working - and so far I cannot see why. Could I ask you to take a look at these please?

I have figured out the reason for this -- the integration test has to include the ExtRef section as well otherwise the counting is never present. I will do some more work. Please do nothing! :wink:

danyill:
I have finally resolved the issues with tests and would be grateful for a review of this.
Sorry for the messy commit history. There were a few mis-steps made amid some confusion but eventually I straightened myself out :wink:

I will be happy to take another review and make further improvements as required.

danyill:
I think Jakob is unavailable for some time. I'd be very pleased to receive a review on this so I can incorporate it into my subscriber view which will be similar.

I've nominated a few people and I would be happy to discuss/go through it if anyone is interested or available to look at this.

test/integration/editors/SMVSubscriberLaterBinding.test.ts
juancho0202, line 1: why are we getting rid of these integration tests? not 100% against it, but I would like to see your reasoning here.

test/integration/editors/GooseSubscriberLaterBinding.test.ts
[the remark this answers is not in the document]
danyill: Looks like an error, my bad. I will revert.

test/integration/editors/SMVSubscriberLaterBinding.test.ts
danyill: Looks like an error, my bad. I will revert.

src/editors/subscription/fcda-binding-list.ts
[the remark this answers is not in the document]
danyill: Thank you for these helpful comments.

Wouldn't this force a recalculation of the classes for all items, and wouldn't we lose speed?
Is this a compile-time vs a run-time consideration?

Currently, we have many classes; as soon as a filter change occurs, the CSS is just applied -- the only recalculation is whatever the CSS engine does in the browser.

With the above change, there may be some additional delays -- each list item needs to be recalculated. I will test the change and see how it works. I suspect the calculation time is quite short. I think it would be helpful if it was a little more straightforward.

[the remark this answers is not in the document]
juancho0202: If you debug in the browser and set breakpoints in your render functions for the `fcda-binding-list.ts` you will realize both approaches need to recalculate each list item everytime you click on the filter buttons. If you're looking for an approach actually purely CSS your renderFCDA and renderControls functions should not depend on `this.hideSubscribed` or `this.hideNotSubscribed`. 

I believe a potential approach for avoiding recalculation should be rendering the elements with two classes that indicate that an item has or doesn't have a subscription (`has-subscription`, `does-not-have-subscription` maybe?) and then at the end of your component just have 
```js
static styles = css`
... /* actually define or not define the two classes depending on the values of this.hideSubscribed or this.hideNotSubscribed */
`;
``` 
That way we only recalculate the CSS and not any of the list-items.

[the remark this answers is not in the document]
danyill: >I believe a potential approach for avoiding recalculation should be rendering the elements with two classes that indicate that an item has or doesn't have a subscription (has-subscription, does-not-have-subscription maybe?) 

That is what I intended to do, the only dependency on `hideSubscribed` and `hideNotSubscribed` is to set the classes on the filtered list. This is already done outside of the render function (by adding an event in `firstUpdated`) and I should not have included it within the render function.

Once I correct this and make sure the property doesn't force a component update, `render` is not called and the filter becomes pure CSS.

The price is the complexity of the CSS to manage the combinations. Is the price too high? 

Your solution also works but requires a re-render of all items which I am now successfully avoiding.

I will be happy to take direction.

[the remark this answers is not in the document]
danyill: Here is an updated table on the 5 CSS rules. I think 3 of them are quite simple, and the other two are not very intuitive:

image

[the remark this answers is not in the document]
juancho0202: @danyill  I believe my approach is not suitable since it would require re-render. If you fix the re-rendering in your approach I'm ok with the classes not being "readable"

juancho0202 (approved):
LGTM!

juancho0202:
hey @danyill please merge main back into your branch 😬

danyill:
Thank you for the assistance on this PR and the detailed review. :clap:

Issue #967, opened 2022-08-30
Add create wizard for `TransformerWinding` element into the wizard-library


JakobVogelsang:
As a user of OpenSCD I want to create a `TransformerWinding` element within a `PowerTransformer` element

**Requirements**
- add create wizard for the `TransformerWinding` to the wizard-library
- make sure that this wizard can be triggered from all potential parents in the substation plugin that is `PowerTransformer`

**Restrictions**
| name | required | type | pattern or restriction | default | helper text |
| --- | --- | --- | --- | --- | --- |
| name |  yes |  xs:normalizedString | min. lenght 1 | no |scl.name |
| desc | no | xs:normalizedString | no | no | scl.desc |
| type | no | enumeration | fixed to `PTW` | `PTW` | scl.type | 
| virtual | no | boolean | no | false | no | 

- make input for `type` attribute disabled please

PR #1154, opened 2023-01-31
feat(substation/transformerwinding): add create wizard

marcvanraalte:


marcvanraalte:
this closes #967

test/unit/editors/substation/__snapshots__/powertransformer-editor.test.snap.js
```diff
@@ -113,6 +113,17 @@ snapshots["powertransformer-editor rendered as action pane looks like the latest
           LNode
         </span>
       </mwc-list-item>
+      <mwc-list-item
```
JakobVogelsang, line 116: Why has this snapshot changed. There is not change in the test file not the test SCD file. Can you explain that?

JakobVogelsang (commented):
I cannot see why there is a change in one of the snapshots.

test/unit/editors/substation/__snapshots__/powertransformer-editor.test.snap.js
JakobVogelsang: Ahh, now I get it. This must be there because `PowerTransformer` is the parent, indeed the only parent, of `TransformerWinding`. In fact, these changes must be on there. Thx

JakobVogelsang (approved):
LGTM

PR #1155, opened 2023-01-31
feat(editors/ied): show all instantiated setting group values

### LINKED ISSUE #1138: Show all `Val` inside a `DAI` element
As a user of OpenSCD I want to see all setting group (`sGroup`) values of given data attribute (`DAI` or `DA` or `BDA`) within the IED section and `DataTypeTemplate` section respectively.
At the moment, only the first `Val` is displayed and can be manipulated. See for a given example, `setVal` only showing the first `Val` element, although it has two.

```xml
<DOI name="ActionTmms" desc="Action time">
	<DAI name="setVal">
		<Val sGroup="1">300</Val>
	        <Val sGroup="2">300</Val>
	</DAI>
	<DAI name="minVal">
		<Val>10</Val>
	</DAI>
	<DAI name="maxVal">
		<Val>60000</Val>
	</DAI>
</DOI>
```

grafik

**Requirement**
- show per given `Val` element, the value of the element
- allow per given `Val` element to change the element
- distinguish the setting groups `sGroup` with a prefix like `SG${val.getAttribute('sGroup')}: `
- if there is no `sGroup` definition leave the UI as is

JakobVogelsang:
Closes #1138

### DIFF
--- a/src/editors/ied/da-wizard.ts
+++ b/src/editors/ied/da-wizard.ts
@@ -22,12 +22,11 @@ import {
 function getValues(element: Element): string {
   const hasValue = getValueElements(element).length !== 0;
 
-  if (hasValue)
-    return `${getValueElements(element)
-      .map(val => val.textContent ?? '')
-      .join(', ')}`;
-
-  return '-';
+  return hasValue
+    ? `${getValueElements(element)
+        .map(val => val.textContent ?? '')
+        .join(', ')}`
+    : '-';
 }
 
 function renderFields(
--- a/src/editors/ied/da-wizard.ts
+++ b/src/editors/ied/da-wizard.ts
@@ -9,7 +9,6 @@ import {
   getDescriptionAttribute,
   getInstanceAttribute,
   getNameAttribute,
-  newWizardEvent,
   Wizard,
 } from '../../foundation.js';
 import { Nsdoc } from '../../foundation/nsdoc.js';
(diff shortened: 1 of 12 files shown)


src/editors/ied/da-wizard.ts
```diff
@@ -9,17 +9,27 @@ import {
...
   findElement,
   findLogicaNodeElement,
-  getValueElement,
+  getValueElements,
 } from './foundation.js';
 
+function getValues(element: Element): string {
+  const hasValue = getValueElements(element).length !== 0;
+
+  if (hasValue)
```
pascalwilbrink, line 25:
``` 
return hasValue ? 
       `${getValueElements(element)
              .map(val => val.textContent ?? '')
              .join(', ')}` : '-'; 
```

Reduces 1 return statement

src/wizards/foundation/dai-field-type.ts
```diff
@@ -208,7 +208,10 @@ export function getCustomField(): Record<DaiFieldTypes, CustomField> {
   }
 
   function getInstanceValue(instanceElement?: Element): string {
-    return instanceElement!.querySelector('Val')?.textContent?.trim() ?? '';
+    if (instanceElement?.tagName === 'DAI')
```
pascalwilbrink, line 211:
```
return instanceElement?.tagName === 'DAI' 
    ? instanceElement!.querySelector('Val')?.textContent?.trim() ?? ''
    : instanceElement!.textContent?.trim() ?? ''
```

Same here

src/editors/ied/da-wizard.ts
JakobVogelsang: Done thx

src/wizards/foundation/dai-field-type.ts
JakobVogelsang: Thx

pascalwilbrink (approved):

PR #1160, opened 2023-02-06
feat(filtered-list) Allow quoting for exact values in filtered list

### LINKED ISSUE #1158: Support quoting of filtered-list search strings

As a user of OpenSCD I want to be able to filter based on the user interface that appears to me.
If there is a space in an identity string or a collection of "paths" then that shouldn't prevent me from carrying out my query.

For instance,  there is a space in the identity string of an LGOS LN, e.g. "LGOS 7" and I want to search all subscriptions for this.

Similarly if I am looking for the protection trip instance 1 operate value,  which we might show in OpenSCD as "PTRC 1.Op" then I should be able to do a "fully-qualified" search for this.

For the examples above I would like to type in either:

- `'LGOS 7'`
- `"LGOS 7"`
- `"PTRC 1.Op"`

And expect the filter to behave correctly.

**Requirements**

Allow quotes (either single or double quotes).

User enters string in filtered text entry field.

danyill:
To allow exact quoting of values including spaces, e.g. "This thing" or 'this thing'

What I like about the implementation is that you can go back and add the quotes without the filter clearing.

I'm somewhat optimistic that with wildcards and quotes, there is not much more I would want in the filtered list for a text entry search (unless we did a full regex search) for "narrowing of the search"

Closes #1158

### DIFF
--- a/src/filtered-list.ts
+++ b/src/filtered-list.ts
@@ -38,21 +38,22 @@ function hideFiltered(item: ListItemBase, searchText: string): void {
     value
   ).toUpperCase();
 
-  const terms: string[] = searchText
-    .toUpperCase()
-    .replace(/[.+^${}()|[\]\\]/g, '\\$&')
-    .trim()
-    .split(/\s+/g);
+  const terms: string[] =
+    searchText
+      .toUpperCase()
+      .replace(/[.+^${}()|[\]\\]/g, '\\$&')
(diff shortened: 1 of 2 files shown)


danyill:
We think a new cooler behaviour is required and we spent, like, half an hour debating it so I think we'll close this!

Issue #1162, opened 2023-02-07
LGOS/LSVS supervision - where should valKind and valImport be found?


danyill:
**Describe the bug**

OpenSCD will instantiate supervision based on the `valImport` and `valKind` attributes.

It looks in the DataTypeTemplates section to confirm these are correct.
It does not look at the first instantiated instance of the LGOS/LSVS. Perhaps it should?

(I apologise for the length of this issue)

For a device we have, the instantiated LN looks like the following:

```xml
<LN lnType="LGOS_4" lnClass="LGOS" inst="1" prefix="">
    <Private type="GE_Digital_Energy_UR LN_UUID">Master_LGOS1</Private>
    <DOI name="NamPlt">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>Ctl</Val>
        </DAI>
    </DOI>
    <DOI name="Beh">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>This logical node's behaviour</Val>
        </DAI>
        <DAI name="stVal" valKind="RO"/>
    </DOI>
    <DOI name="St">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>RxGOOSE1 On operand</Val>
        </DAI>
        <DAI name="stVal" valKind="RO" sAddr="0:1441793,0"/>
    </DOI>
    <DOI name="Alm">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>RxGOOSE1 Off operand</Val>
        </DAI>
        <DAI name="stVal" valKind="RO" sAddr="0:1507329,0"/>
    </DOI>
    <DOI name="SimSt">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>Status showing that Sim messages are being received and accepted for RxGOOSE1</Val>
        </DAI>
        <DAI name="stVal" valKind="RO" sAddr="2:288538,0"/>
    </DOI>
    <DOI name="LastStNum">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>Actual value RxGOOSE1 StNum</Val>
        </DAI>
        <DAI name="stVal" valKind="RO" sAddr="1:316928,0"/>
        <SDI name="units">
            <DAI name="SIUnit" valKind="RO">
                <Val></Val>
            </DAI>
            <DAI name="multiplier" valKind="RO">
                <Val></Val>
            </DAI>
        </SDI>
    </DOI>
    <DOI name="LastSqNum">
[...]

PR #1165, opened 2023-02-08
fix(editors/subscriber): Count supervision instances correctly

### LINKED ISSUE #1161: LGOS/LSVS supervision instantiation counts existing instances incorrectly
**Describe the bug**

Currently we check several items as part of instantiating supervisions:

https://github.com/openscd/open-scd/blob/8ebac53bbf86b86e6b8cdb9d79b28957d4a7b5fc/src/editors/subscription/foundation.ts#L355-L378

One of these items is `instantiatedSupervisionsCount`:

https://github.com/openscd/open-scd/blob/8ebac53bbf86b86e6b8cdb9d79b28957d4a7b5fc/src/editors/subscription/foundation.ts#L509-L520

With the following (slightly hand modified IID file), supervisions cannot be instantiated.

[B30_valKind_valImport.zip](https://github.com/openscd/open-scd/files/10670869/B30_valKind_valImport.zip)

The reason for this is that the count function is showing a much larger number of supervisions than exist.

This is because we are counting all `DOI`s which have `DAI`s which have a `Val`. In this particular IED there are many `DOI > DAI > Val` within each `LGOS`.

```xml
<LN lnType="LGOS_4" lnClass="LGOS" inst="1" prefix="">
    <Private type="GE_Digital_Energy_UR LN_UUID">Master_LGOS1</Private>
    <DOI name="NamPlt">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>Ctl</Val>
        </DAI>
    </DOI>
    <DOI name="Beh">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>This logical node's behaviour</Val>
        </DAI>
        <DAI name="stVal" valKind="RO"/>
    </DOI>
    <DOI name="St">
        <DAI name="d" valKind="RO" valImport="true">
            <Val>RxGOOSE1 On operand</Val>
        </DAI>
        <DAI name="stVal" valKind="RO" sAddr="0:1441793,0"/>
    </DOI>
and so on...
```

There are enough supervision instances available, within the `Services` section: `<SupSubscription maxGo="64" maxSv="16"/>`

I think this query needs to be tightened up. When is a supervision instantiated? I think only when a DO `GoCBRef`has a `Val` against the DA `SetSrcRef`.

```
Array.from(subscriberIED.querySelectorAll(`LN[lnClass="${supervisionType}"]>DOI>DAI>Val`))
(396)

Array.from(subscriberIED.querySelectorAll(`LN[lnClass="${supervisionType}"]>DOI[name="GoCBRef"]>DAI[name="setSrcRef"]>Val`))
12
```

I think this change would be sufficient (with appropriate modifications for the control block type):

```diff
const instantiatedValues = Array.from( 
     subscriberIED.querySelectorAll( 
-       `LN[lnClass="${supervisionType}"]>DOI>DAI>Val` 
+      `LN[lnClass="${supervisionType}"]>DOI[name="GoCBRef"]>DAI[name="setSrcRef"]>Val`
     ) 
   ).filter(val => val.textContent !== '');
```

**To Reproduce**

1. Import an instance of the IED included in the above zip file.
2. Include another IED, e.g. the attached: [NR.zip](https://github.com/openscd/open-scd/files/10671066/NR.zip)
3. Try to carry out a GOOSE subscription.
4. No supervision occurs. 

**Expected behavior**

A supervised GOOSE instance should appear with the supervision icon.

danyill:
With regression test.
Closes #1161
#1162 will follow on from this
Still a bit repetitive.

### DIFF
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -296,6 +296,25 @@ export function instantiateSubscriptionSupervision(
 }
 
 /**
+ * Return Val elements within an IED for the control block type
+ * @param ied - IED SCL element
+ * @param cbTagName - Either GSEControl or (defaults to) SampledValueControl
+ * @returns an Element array of Val SCL elements
+ */
+export function getSupervisionCbRefs(
+  ied: Element,
+  cbTagName: string
+): Element[] {
+  const supervisionType = cbTagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  const supervisionName = supervisionType === 'LGOS' ? 'GoCBRef' : 'SvCBRef';
+  return Array.from(
+    ied.querySelectorAll(
+      `LN[lnClass="${supervisionType}"]>DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val,LN0[lnClass="${supervisionType}"]>DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val`
+    )
+  );
+}
+
+/**
  * Return an array with a single Delete action to delete the supervision element
  * for the given GOOSE/SMV message and subscriber IED.
  *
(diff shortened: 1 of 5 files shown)


src/editors/subscription/foundation.ts
```diff
@@ -407,13 +427,15 @@ export function findOrCreateAvailableLNInst(
...
+    const instantiatedSiblings = getSupervisionCbRefs(
+      subscriberIED,
+      controlBlock.tagName
+    ).map(val => val.closest('LN'));
+
+    if (!instantiatedSiblings || instantiatedSiblings.length === 0) return null;
     availableLN.setAttribute(
       'lnType',
-      instantiatedSibling.getAttribute('lnType') ?? ''
+      instantiatedSiblings[0]!.getAttribute('lnType') ?? ''
```
danyill, line 438: I think this is OK?

danyill:
Closing in favour of #1169 which includes this work

PR #1169, opened 2023-02-10
feat(editors/subscription): Support first instance valKind and valImport for supervision instances

### LINKED ISSUE #1162: LGOS/LSVS supervision - where should valKind and valImport be found?

danyill:
Closes #1162
TODO: add some detail here:

### DIFF
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -296,6 +296,25 @@ export function instantiateSubscriptionSupervision(
 }
 
 /**
+ * Return Val elements within an IED for the control block type
+ * @param ied - IED SCL element
+ * @param cbTagName - Either GSEControl or (defaults to) SampledValueControl
+ * @returns an Element array of Val SCL elements
+ */
+export function getSupervisionCbRefs(
+  ied: Element,
+  cbTagName: string
+): Element[] {
+  const supervisionType = cbTagName === 'GSEControl' ? 'LGOS' : 'LSVS';
+  const supervisionName = supervisionType === 'LGOS' ? 'GoCBRef' : 'SvCBRef';
+  return Array.from(
+    ied.querySelectorAll(
+      `LN[lnClass="${supervisionType}"]>DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val,LN0[lnClass="${supervisionType}"]>DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val`
+    )
+  );
+}
+
+/**
  * Return an array with a single Delete action to delete the supervision element
  * for the given GOOSE/SMV message and subscriber IED.
  *
@@ -310,12 +329,9 @@ export function removeSubscriptionSupervision(
   subscriberIED: Element | undefined
 ): Delete[] {
   if (!controlBlock || !subscriberIED) return [];
-  const supervisionType =
-    controlBlock?.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
-  const valElement = Array.from(
-    subscriberIED.querySelectorAll(
-      `LN[lnClass="${supervisionType}"]>DOI>DAI>Val,LN0[lnClass="${supervisionType}"]>DOI>DAI>Val`
-    )
+  const valElement = getSupervisionCbRefs(
+    subscriberIED,
+    controlBlock.tagName
   ).find(val => val.textContent == controlBlockReference(controlBlock));
   if (!valElement) return [];
   const lnElement = valElement.closest('LN0, LN');
@@ -361,16 +377,14 @@ function isSupervisionAllowed(
   if (subscriberIED.querySelector(`LN[lnClass="${supervisionType}"]`) === null)
     return false;
   if (
-    Array.from(
-      subscriberIED.querySelectorAll(
-        `LN[lnClass="${supervisionType}"]>DOI>DAI>Val`
-      )
-    ).find(val => val.textContent == controlBlockReference(controlBlock))
+    getSupervisionCbRefs(subscriberIED, controlBlock.tagName).find(
+      val => val.textContent == controlBlockReference(controlBlock)
+    )
   )
     return false;
   if (
     maxSupervisions(subscriberIED, controlBlock) <=
-    instantiatedSupervisionsCount(subscriberIED, controlBlock, supervisionType)
+    instantiatedSupervisionsCount(subscriberIED, controlBlock)
   )
     return false;
 
@@ -390,11 +404,17 @@ export function findOrCreateAvailableLNInst(
 ): Element | null {
   let availableLN = Array.from(
     subscriberIED.querySelectorAll(`LN[lnClass="${supervisionType}"]`)
-  ).find(
-    ln =>
-      ln.querySelector('DOI>DAI>Val') === null ||
-      ln.querySelector('DOI>DAI>Val')?.textContent === ''
-  );
+  ).find(ln => {
+    const supervisionName = supervisionType === 'LGOS' ? 'GoCBRef' : 'SvCBRef';
+    return (
+      ln.querySelector(
+        `DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val`
+      ) === null ||
+      ln.querySelector(
+        `DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]>Val`
+      )?.textContent === ''
+    );
+  });
   if (!availableLN) {
     availableLN = subscriberIED.ownerDocument.createElementNS(
       SCL_NAMESPACE,
@@ -407,13 +427,15 @@ export function findOrCreateAvailableLNInst(
     openScdTag.setAttribute('type', 'OpenSCD.create');
     availableLN.appendChild(openScdTag);
     availableLN.setAttribute('lnClass', supervisionType);
-    const instantiatedSibling = subscriberIED
-      .querySelector(`LN[lnClass="${supervisionType}"]>DOI>DAI>Val`)
-      ?.closest('LN');
-    if (!instantiatedSibling) return null;
+    const instantiatedSiblings = getSupervisionCbRefs(
+      subscriberIED,
+      controlBlock.tagName
+    ).map(val => val.closest('LN'));
+
+    if (!instantiatedSiblings || instantiatedSiblings.length === 0) return null;
     availableLN.setAttribute(
       'lnType',
-      instantiatedSibling.getAttribute('lnType') ?? ''
+      instantiatedSiblings[0]!.getAttribute('lnType') ?? ''
     );
   }
 
@@ -508,13 +530,11 @@ export function getExistingSupervision(extRef: Element | null): Element | null {
  */
 export function instantiatedSupervisionsCount(
   subscriberIED: Element,
-  controlBlock: Element,
-  supervisionType: string
+  controlBlock: Element
 ): number {
-  const instantiatedValues = Array.from(
-    subscriberIED.querySelectorAll(
-      `LN[lnClass="${supervisionType}"]>DOI>DAI>Val`
-    )
+  const instantiatedValues = getSupervisionCbRefs(
+    subscriberIED,
+    controlBlock.tagName
   ).filter(val => val.textContent !== '');
   return instantiatedValues.length;
 }
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -327,7 +327,9 @@ export function instantiateSubscriptionSupervision(
     });
   }
 
-  let daiElement = availableLN.querySelector(`DAI[name="setSrcRef"]`);
+  let daiElement = availableLN.querySelector(
+    `DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]`
+  );
   if (!daiElement) {
     daiElement = subscriberIED.ownerDocument.createElementNS(
       SCL_NAMESPACE,
@@ -360,14 +362,14 @@ export function instantiateSubscriptionSupervision(
       SCL_NAMESPACE,
       'Val'
     );
-    valElement.textContent = controlBlockReference(controlBlock);
-    actions.push({
-      new: {
-        parent: daiElement!,
-        element: valElement,
-      },
-    });
   }
+  valElement.textContent = controlBlockReference(controlBlock);
+  actions.push({
+    new: {
+      parent: daiElement!,
+      element: valElement,
+    },
+  });
 
   return actions;
 }
--- a/src/editors/subscription/foundation.ts
+++ b/src/editors/subscription/foundation.ts
@@ -11,11 +11,6 @@ import {
   isPublic,
   minAvailableLogicalNodeInstance,
 } from '../../foundation.js';
-import {
-  createTemplateStructure,
-  determineUninitializedStructure,
-  initializeElements,
-} from '../../foundation/dai.js';
 import { getFcdaReferences } from '../../foundation/ied.js';
 import { SCL_NAMESPACE } from '../../schemas.js';
 
@@ -233,6 +228,34 @@ function checksDataTypeTemplateConditions(lnElement: Element): boolean {
 }
 
 /**
+ * Searches for first instantiated LGOS/LSVS LN for presence of DOI>DAI[valKind=Conf/RO][valImport=true]
+ * given a supervision type and returns a boolean indicating whether it allows instantiation by an SCT.
+ * @param ied - SCL IED element.
+ * @param supervisionType - either 'LGOS' or 'LSVS' supervision LN classes.
+ * @returns boolean indicating if subscriptions are allowed based on the first instantiated instance.
+ */
+function checkInstSupervisionConditions(
+  ied: Element,
+  supervisionType: string
+): boolean {
+  const firstSupervisionLN = ied.querySelector(
+    `LN[lnClass="${supervisionType}"]`
+  );
+
+  if (firstSupervisionLN === null) return false;
+
+  const supervisionName = supervisionType === 'LGOS' ? 'GoCBRef' : 'SvCBRef';
+  const valKind = firstSupervisionLN!
+    .querySelector(`DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]`)
+    ?.getAttribute('valKind');
+  const valImport = firstSupervisionLN!
+    .querySelector(`DOI[name="${supervisionName}"]>DAI[name="setSrcRef"]`)
+    ?.getAttribute('valImport');
+
+  return (valKind === 'RO' || valKind === 'Conf') && valImport === 'true';
+}
+
+/**
  * Returns an array with a single Create action to create a new
  * supervision element for the given GOOSE/SMV message and subscriber IED.
  *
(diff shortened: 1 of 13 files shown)


src/editors/subscription/foundation.ts
```diff
@@ -407,13 +523,15 @@ export function findOrCreateAvailableLNInst(
...
+    const instantiatedSiblings = getSupervisionCbRefs(
+      subscriberIED,
+      controlBlock.tagName
+    ).map(val => val.closest('LN'));
+
+    if (!instantiatedSiblings || instantiatedSiblings.length === 0) return null;
     availableLN.setAttribute(
       'lnType',
-      instantiatedSibling.getAttribute('lnType') ?? ''
+      instantiatedSiblings[0]!.getAttribute('lnType') ?? ''
```
danyill, line 534: Is this call for `instantiatedSiblings[0]` safe?

danyill:
We did some testing of this PR's functionality with the GE B30 bus protection relay and found it worked correctly, and could be loaded to the device and functioned correctly online.

What worked was:

* updating the GoCBRef for an existing LGOS instance in OpenSCD
* creating a new LGOS instance in OpenSCD and importing it in the Enervista UR Setup tool (version 8.40). Enervista instantiates attributes for the LGOS instance, removes the OpenSCD private element and this seems to me to be a reasonable implementation.
* subscribed attributes were updated
* the supervision instance worked correctly with the device online (if the sending device's ethernet port was disabled the LGOS instance showed stVal=false)

What is interesting (we think) for the GE B30 is that without supervision instantiation ExtRefs are not processed on import from an scd file. Supervision is a pre-requisite to processing of data attributes in GOOSE messages. Normally I think of supervision as a "bolt on" feature for subscriptions but for this relay that appears not to be so.

danyill:
> includes a Move action to move new supervision instances to the end of the supervision section (but new ones are always second-to-last since the action API doesn't allow them to be last cry). I think this matters, because if there is only two then I think the manufacturers one should be first in the file but I think fixing it is perhaps outside of the scope of this PR. However I think including a Move action is a good first step.

I discussed with Christian -- I'll update this PR to remove the move action, have only Create actions and set the reference to `null` which will result in the elements being appended at the end. 

Alas, I have not previously understood that  this was an option passed through to `Node.insertBefore`

src/editors/subscription/foundation.ts
```diff
@@ -10,12 +10,8 @@ import {
   getSclSchemaVersion,
   isPublic,
   minAvailableLogicalNodeInstance,
+  Move,
```
JakobVogelsang, line 13: I was thinking you removed the `Move` action and instead want to use `Remove` and `Create`?

src/editors/subscription/foundation.ts
```diff
@@ -232,6 +228,34 @@ function checksDataTypeTemplateConditions(lnElement: Element): boolean {
...
 }
 
+/**
+ * Searches for first instantiated LGOS/LSVS LN for presence of DOI>DAI[valKind=Conf/RO][valImport=true]
+ * given a supervision type and returns a boolean indicating whether it allows instantiation by an SCT.
+ * @param ied - SCL IED element.
+ * @param supervisionType - either 'LGOS' or 'LSVS' supervision LN classes.
+ * @returns boolean indicating if subscriptions are allowed based on the first instantiated instance.
+ */
+function checkInstSupervisionConditions(
```
JakobVogelsang, line 238: This function and the function above are doing a very similar thing, and I wonder whether we can unify this to one that is `checkSupervisionConditions`. This is then responsible to check whether those are met and how to do that. If we forgot one check, we can easily extend this function in the future.

src/editors/subscription/foundation.ts
```diff
@@ -257,42 +281,115 @@ export function instantiateSubscriptionSupervision(
...
-    if (parent)
-      createActions.push({
+    if (parent) {
+      // use Create Action for supervision LN
+      actions.push({
         new: {
-          parent,
+          parent: parent,
           element: availableLN,
+          reference: null,
```
JakobVogelsang, line 305: I am not sure if this is not breaking something. `reference = null` is adding it to the end of the parent. In a LN0 parent this would mean adding it after the `Inputs` child which is against the schema. My gut feeling is that this requires either an own `getReference` or to have a config option in the current `getReference` to add at the end of a set of given element say `LN` and `LN0`.

JakobVogelsang (commented):
I think there is some schema nonconforming behavior there. The rest LGTM.

src/editors/subscription/foundation.ts
danyill: Yes, that's a good idea, thanks.

src/editors/subscription/foundation.ts
danyill: I will look into this some more, thanks.

src/editors/subscription/foundation.ts
JakobVogelsang: Not nice. I would have used a query to pick the first of some type. As there is a check for the length I would say it is ok. You could also change `instantiatedSiblings.length === 0` to `!instantiatedSibling[0]`

src/editors/subscription/foundation.ts
danyill: Quite right, I have also dealt with some other loose imports.

src/editors/subscription/foundation.ts
danyill: You are insightful :dart: 

I am unsure but...

I think our current API doesn't allow the behaviour I'm looking for.

What we are are looking for is to append this at the end of existing `LN[lnCLass="SVS"]` instances or perhaps more appropriately (?) at the end of all `LN` instances. If the first LSVS instances is also the last `LN` of the `LDevice` then I think it cannot be added and we should cover this use case.

My options appear to be:

* I can specify `null` on my create action and add to the end of the parent's children - causing a schema invalid outcome (see below and as you have pointed out)
* I can specify an element in which case it allows putting it _before_ but not _after_ the element.

I think I cannot specify  "here is the element to which I wish to be the next sibling".

I think there is also a conflict for `LDevice`, because the `AccessControl` element:

```xml
  <xs:complexType name="tLDevice">
    <xs:complexContent>
      <xs:extension base="tUnNaming">
        <xs:sequence>
          <xs:element ref="LN0" />
          <xs:element ref="LN" minOccurs="0" maxOccurs="unbounded" />
          <xs:element name="AccessControl" type="tAccessControl" minOccurs="0" />
        </xs:sequence>
        <xs:attribute name="inst" type="tLDInst" use="required" />
        <xs:attribute name="ldName" type="tLDName" use="optional" />
      </xs:extension>
    </xs:complexContent>
  </xs:complexType>
```

Does this imply a short-fall in the Create API for this use case?

We could modify getReference, perhaps something along the lines of the following (not tested!) with a "lastReference" parameter, but once we have received the reference we can't use it with the Create API.

```js
export function getReference(
  parent: Element,
  tag: SCLTag,
  lastReference?: boolean
): Element | null {
  const parentTag = parent.tagName;
  const children = Array.from(parent.children);

  if (
    parentTag === 'Services' ||
    parentTag === 'SettingGroups' ||
    !isSCLTag(parentTag)
  )
    return children.find(child => child.tagName === tag) ?? null;

  const sequence = tags[parentTag]?.children ?? [];
  let index = sequence.findIndex(element => element === tag);

  if (index < 0) return null;

  let nextSibling: Element | undefined;
  while (index < sequence.length && !nextSibling) {
    // if lastReference is true, find the last element of the type
    lastReference
      ? (nextSibling = children.find(
          child => child.tagName === sequence[index]
        ))
      : (nextSibling = children.find(
          child => child.tagName === sequence[index]
        ));

    index++;
  }

  return nextSibling ?? null;
}
```

WDYT?

danyill: Maybe we really _did_ want this in a Move action after all and this is what should be augmented?

src/editors/subscription/foundation.ts
danyill: I have done something, I will be interested in your thoughts on it, it may be too complicated.

src/editors/subscription/foundation.ts
JakobVogelsang: Hmmm there are two ways we could go about this.
1. Rebuild the `getReference` and allow append not insert before. 
2. Solve user defined wherever it is needed in the code like in your situation here. 

I personally am for the second option if you ask me. The reason is when you do care about the order of children in the parent, you often do need more control than append vs insert before. In this case, you just write your own little `getReference`. In your case this would look something like this.

```ts
new: {
   parent,
   element,
   reference: parent.querySelector('LN[....]:last-child').nextElementSibling 
```

danyill: That sounds very fine to me, let's take the second option.

danyill:
Thank you for this review -- I'd be very pleased to receive another of the changes :smile: 

I've overloaded the function signature on `getSupervisionCbRefs` because it seemed like a simple solution but is not something I have done before -- if it's too complex, I am happy to revisit it.

JakobVogelsang (approved):
LGTM

danyill:
Hurrah! Thank you for this thorough review and the hints no ways to improve.

PR #1181, opened 2023-02-22
feat(openscd): Move progress indicator beneath plugin tabs

### LINKED ISSUE #1178: Move validation progress indicator

Currently when the validator is called, a circular progress indicator appears on the screen.

This doesn't  block user actions, but it may appear to -- "This thing is in the middle of the screen, maybe I shouldn't keep doing stuff", or even more "If I do something while the computer is doing something else, will they interact badly?".

I think it could be moved to be a small icon adjacent to the undo/redo buttons and this would improve the presentation to the user and also point them toward the area where we have the log for actions/edits and the undo/redo button.

image

**Requirements**

User interface is not overlaid with information which suggests a blocking action is taking place.

danyill:
Closes #1178.

Unfortunately, the buttons and the title already overlap in the mobile view, which means the progress indicator on a small mobile screen will hover in the centre of the screen. 

This is not ideal, but I can't think of a better location, and would be willing to live with myself over this...

image

Otherwise I quite like it:

image

This especially bothered me on the subscriber plugins where I was constantly carrying out an action, near the centre of the screen and found the progress indicator distracting and overpowering.

As always a review would be appreciated as CSS has been carefully designed to impart humility and there is no reason why I should be exempt.

I wonder if this is inadequate -- I'm not sure how CoMPAS downstream integrates or consumes open-scd, I know it's top bar is different. Perhaps this needs to be slotted somewhere?


danyill:
After our discussion at a recent refinement I'll look into other options to improve compatibility with CoMPAS.

danyill:
I first tried to put the progress bar under the filename. Can be done but impossible to style with mwc linear progress out of the box (it seems to me) as I can apply suitable overrides.

So I thought why don't I put it just below the tab bar and have it occupy the full width? Not perfect but better. I've decreased the opacity to make it less visually intrusive.

I wanted to avoid the progress bar being on top of the menu if a validation action is triggered from a menu action.
So I made it close the menu always. This is, I hope a general improvement and works nicely _except_ that the menu plugin, "Show SCL History" goes into an infinite update loop and never opens. I may need some help with why this occurs (how odd?!)

src/open-scd.ts
```diff
@@ -119,13 +119,15 @@ export class OpenSCD extends Hosting(
...
-      left: 50%;
-      transform: translate(-50%, -50%);
-      z-index: 99;
+      --mdc-theme-primary: var(--mdc-theme-secondary);
+      left: 0px;
+      top: 112px;
+      opacity: 50%;
+      width: 100%;
       pointer-events: none;
+      z-index: 99;
```
ca-d, line 130: Have you tried whether changing the order of mixins in `open-scd.ts` in combination with removing the `z-index` from the style sheet does something to alleviate the problem of the progress bar rendering atop the open menu?

src/open-scd.ts
```diff
@@ -119,13 +119,15 @@ export class OpenSCD extends Hosting(
...
-    mwc-circular-progress-four-color {
+    mwc-linear-progress {
       position: fixed;
-      top: 50%;
-      left: 50%;
-      transform: translate(-50%, -50%);
-      z-index: 99;
+      --mdc-theme-primary: var(--mdc-theme-secondary);
+      left: 0px;
+      top: 112px;
```
ca-d, line 126: I think this might need an alternative value set inside a media query in order to align with the lower top bar height on smaller screens. I'll try to produce a screenshot to see if this is an issue.

ca-d (dismissed):
I'd really prefer rendering the linear progress bar between the editor plugin and the menu (in terms of CSS layering, meaning arranging it in the DOM to appear after the editor plugin but still inside the drawer) and removing the `z-index` definition. Do you think that is a viable alternative?

src/open-scd.ts
danyill: I have not I guess because I didn't especially like the way the menu stayed open anyway.

However on testing the z-index (which I only kept) seems unnecessary.

I tested making `Waiting` precede `Hosting` and that seems to do the trick. I guess it's a change in the stacking context but I can't see exactly how as the progress indicator is well below the `aside` in `mwc-drawer` in the DOM.

So I've removed the automatic menu close and the z-index. I've also contemplated whether modifying the opacity is a crime against our theming and decided to remove it for now (unless you think otherwise).

src/open-scd.ts
danyill: I have added a media query, thanks!

src/open-scd.ts
```diff
@@ -23,8 +23,8 @@ import { Wizarding } from './Wizarding.js';
 /** The `<open-scd>` custom element is the main entry point of the
  * Open Substation Configuration Designer. */
 @customElement('open-scd')
-export class OpenSCD extends Hosting(
-  Setting(Wizarding(Waiting(Plugging(Editing(Logging(LitElement))))))
+export class OpenSCD extends Waiting(
+  Hosting(Setting(Wizarding(Plugging(Editing(Logging(LitElement))))))
```
danyill, line 27: I was not sure how far up the stack to move this and this was the first thing I tried. It seems like "Waiting" was somehow less important or more appropriately the last item if I had to change it.

danyill:
@ca-d, is there anything I can do to move this along?

pascalwilbrink (dismissed):
Looks good! quick scan is saying that `mwc-circular-progress-four-color` isn't being used anymore. Can it be removed from the package.json?

danyill:
Thank you for this review, I have merged to main and cleaned up the package.

pascalwilbrink (approved):

danyill:
Thanks for the review!

PR #1182, opened 2023-02-22
fix(foundation) improve ExtRef identity function

### LINKED ISSUE #1179: Improve ExtRef identity function
**Describe the bug**

Currently the ExtRef identity function is not stable for the case of:

* An ExtRef is specified for later binding with an internal address
* A subscription is made

If an IED has an ExtRef and an internal address and the user then does a subscription, then the identity is changed.

https://github.com/openscd/open-scd/blob/e1b9568a1c1d0f04f2dacea133eea04e9ee28af2/src/foundation.ts#L824-L872

The identity function should return a stable reference.

If an internal address is specified, I think this should be the "primary key". It is also true that tne `intAddr` can be mapped multiple times but we already use the instance number to take this into account.

**Expected behavior**

@ca-d and I discussed, and tentatively propose:

```diff
function extRefIdentity(e: Element): string | number {
  if (!e.parentElement) return NaN;
  const parentIdentity = identity(e.parentElement);
  const iedName = e.getAttribute('iedName');
  const intAddr = e.getAttribute('intAddr');
  const intAddrIndex = Array.from(
    e.parentElement.querySelectorAll(`ExtRef[intAddr="${intAddr}"]`)
  ).indexOf(e);
-  if (!iedName) return `${parentIdentity}>${intAddr}[${intAddrIndex}]`;
+  if (intAddr) return `${parentIdentity}>${intAddr}[${intAddrIndex}]`;
  const [
    ldInst,
    prefix,
    lnClass,
    lnInst,
    doName,
    daName,
    serviceType,
    srcLDInst,
    srcPrefix,
    srcLNClass,
    srcLNInst,
    srcCBName,
  ] = [
    'ldInst',
    'prefix',
    'lnClass',
    'lnInst',
    'doName',
    'daName',
    'serviceType',
    'srcLDInst',
    'srcPrefix',
    'srcLNClass',
    'srcLNInst',
    'srcCBName',
  ].map(name => e.getAttribute(name));

  const cbPath = srcCBName
    ? `${serviceType}:${srcCBName} ${srcLDInst ?? ''}/${srcPrefix ?? ''} ${
        srcLNClass ?? ''
      } ${srcLNInst ?? ''}`
    : '';
  const dataPath = `${iedName} ${ldInst}/${prefix ?? ''} ${lnClass} ${
    lnInst ?? ''
  } ${doName} ${daName ? daName : ''}`;
  return `${parentIdentity}>${cbPath ? cbPath + ' ' : ''}${dataPath}${
    intAddr ? '@' + `${intAddr}` : ''
  }`;
}
```

danyill:
Closes #1179 

Happy to withdraw this PR if it doesn't pass refinement.
This would be helpful for me in the subscriber view for later binding.

A review would be gratefully accepted.


ca-d (approved):
LGTM

Thank you very much for this contribution, I think it would also be fair to classify this as `fix:` rather than `feat:`, do you agree?

danyill:
> LGTM
> 
> Thank you very much for this contribution, I think it would also be fair to classify this as `fix:` rather than `feat:`, do you agree?

It shall be as you wish.

PR #1186, opened 2023-02-24
fix: added translation key for phase

### LINKED ISSUE #1184: Translation error in the `Add SubEquipment` dialog
As a user of OpenSCD I want to have clear (translated) texts in the dialog views.

**Reproduce**

- Click on the add button on, for example, a power transformer
- Choose SubEquipment
- In the phase field you see that the text is not translated correctly
Schermafbeelding_20230224_134655

**Todo**
See if the needed translation key exists, if so correct it. If not create a new translation key for this one

Stef3st:
closes #1184


ca-d (approved):
LGTM, thanks!

PR #1199, opened 2023-03-10
feat(wizards): Pretty print XML inside code editor, closes #895

### LINKED ISSUE #895: Pretty-print XML in code editor

For advanced editing and looking at an SCD file, some editors pretty-print the XML and some do not.

Either a reformat for the whole document or a reformat within the code editor.

**Requirements**

A mechanism for reformatting the document or what the user sees in the Ace editor.

We could have a reformat button in the Ace editor (the green circle).
The XML would then be reformatted (red arrow).

image

danyill:
Closes #895

We use `formatXml` to save documents.

Is it dangerous to use it here? Perhaps a little bit.

When OpenSCD modifies elements e.g. datasets, subscriptions or supervision elements, the XML it produces is difficult to view in the Ace editor without manually adding linebreaks. This resolves that.

### DIFF
--- a/src/wizard-dialog.ts
+++ b/src/wizard-dialog.ts
@@ -42,6 +42,7 @@ import {
   identity,
   WizardInput,
   WizardMenuActor,
+  formatXml,
 } from './foundation.js';
 
 function renderWizardInput(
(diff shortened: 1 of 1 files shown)


Sander3003:
Let's try it, @danyill please check if private sections are still working with multiple vendors.

ca-d (approved):
LGTM, we decided at today's refinement that it would be nice to test this pretty printing on some real world vendor specific `Private` sections to see if changing whitespace presents a problem in that scenario

danyill:
Thanks for reviewing. I will look into testing this, although it should be no less safe than our existing "Save Project".

Issue #1139, opened 2023-01-18
Add multiple `Val` element into `DAI` distinguished by `sGroup`


JakobVogelsang:
As a user of the `IED` plugin, I want to instantiate multiple setting group parameters at the same time.

**Requirement**
- if a data attribute is not instantiated, show an add icon on the header of the data attribute container
- given the data attribute `DAI` has the functional constraint `SG` or `SE`
   - change the create wizard to the mock below
   - given setting groups are available at all
     -> the parent `IED` element must have a `SettingControl` element and this an attribute `numOfSGs`
     -  allow to specify a specific setting group for the newly created value `sGroup`
       -> checked `show individual setting groups` (default) 
       - show as many input fields as there are `numOfSGs`, make them required to assure a value
     - allow to specify that the given value is for all setting groups
       -> unchecked `show individual setting groups` 
   - without existing `SettingControl`
     - leave the wizards as is at the moment
- given the data attribute `DAI` has any other functional contraint
   - leave the wizards as is at the moment

Mocks:
- when fc=SG or SE and user wants to set setting groups individualy
![grafik](https://user-images.githubusercontent.com/66802940/213170866-2edc0477-4458-4ff4-ab16-42483bd3b57f.png)

- when fc=SG or SE and user wants to have value for all setting groups
![grafik](https://user-images.githubusercontent.com/66802940/213170889-1e8353a7-67ee-49be-b819-dda2ac8b4d2b.png)

-when fc=SG or SE and not `SettingControl` is defined in parent `IED`
- wizard as is

- when fc!=SG or SE
- wizard as is

romdhanisam:
According to `SCL`, the `SettingControl` element are not allowed in `IED`, they can only be found in `LN0`
1. 2003: https://github.com/openscd/open-scd/blob/main/src/schemas.ts#L1477
2. 2007B: https://github.com/openscd/open-scd/blob/main/src/schemas.ts#L4575
3. 2007B4: https://github.com/openscd/open-scd/blob/main/src/schemas.ts#L8014

Sander3003:
![image](https://user-images.githubusercontent.com/7427358/227514413-a5723f3d-33ad-428a-90a4-742a0f597e3c.png)
 @romdhanisam that is correct, the sgroup can be used in the LN as well.

juancho0202:
How do we check if a `DAI` has a functional constraint of SE or SG?

Sander3003:
[...]

PR #1221, opened 2023-03-31
feat(openscd):Add multiple Val element into DAI distinguished by sGroup

juancho0202:
### Note to reviewer:
My approach is to mod all the custom fields in `dai-field-type.ts` so that both the **render** and **value** functions include an extra parameter called `numOfSGs` and modify their behavior accordingly when the param is present.

I tested this initially by modifying the `test\testfiles\wizards\ied.scd` file so that the `ScaledValueConfig` DA would comply with all the conditions described in #1139 and it seems to be doing the job nicely, but I would prefer if some IEC 61850 expert in the room could step up and help me build a proper file for testing this functionality.


src/wizards/foundation/dai-field-type.ts
```diff
@@ -63,32 +67,44 @@ export function getCustomField(): Record<DaiFieldTypes, CustomField> {
...
-          html`<wizard-select
-            id="Val"
-            label="Val"
+      render: (
+        element: Element,
+        instanceElement?: Element,
+        numOfSGs: number | null = null
+      ) => {
+        // If numOfSGs is -1, then it is a single value, otherwise it is treated as a group of values
+        return (numOfSGs ? [...Array(numOfSGs)] : [numOfSGs]).map((item, i) => {
```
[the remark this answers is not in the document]
juancho0202: maybe there is a better way of getting what I want instead of the `[null].map(` ? Here's what I'm trying to do: In case there is a specific `numOfSGs` execute the loop #numOfSGs times. Else just execute the loop once. In the current implementation if I do `[].map(` I don't get the "execute the loop once" it just doesn't get executed at all.

src/wizards/dai.ts
```diff
@@ -41,18 +41,41 @@ export function createValue(
...
+          i + 1
+        );
+
+        const valElement = parent.ownerDocument.createElementNS(
+          SCL_NAMESPACE,
+          'Val'
+        );
+        valElement.textContent = newValue;
+        valElement.setAttribute('sGroup', `${i + 1}`);
+        instanceElement.append(valElement);
```
pascalwilbrink, line 67: Should this go via the customEvent `Create` instead of directly appending it to the instanceElement?

src/wizards/dai.ts
```diff
@@ -41,18 +41,41 @@ export function createValue(
...
+      });
+    } else {
+      const newValue = getCustomField()[<DaiFieldTypes>bType].value(inputs);
+
+      let valElement = instanceElement.querySelector('Val');
+      if (!valElement) {
+        valElement = parent.ownerDocument.createElementNS(SCL_NAMESPACE, 'Val');
+        instanceElement.append(valElement);
```
pascalwilbrink, line 75: same here

src/wizards/dai.ts
juancho0202: it shouldn't in this case since instanceElement is not an existing element in the SCL document but rather a reference to a piece of SCL document that is about to be appended with the `Create` option of the **Action API**

https://github.com/openscd/open-scd/blob/ff047c78121a4e90c274a3874e96228755fc417b/src/wizards/dai.ts#L154-L158

src/wizards/dai.ts
juancho0202: same here 🔙 😅

pascalwilbrink (approved):

PR #1222, opened 2023-04-03
fix(menu/importIEDs): Allow importing multiple IEDs from multiple SCD files

### LINKED ISSUE #1106: Import IEDs only imports from the first document
**Describe the bug**

When importing IEDs from an SCD file, I have the ability to select multiple scd files.

When I select multiple scd files, only  the first one is processed.

**To Reproduce**

1. Go to Import IEDs menu plugin.
2. Select two files with more than one IED in them.
3. Notice the dialog allows you to select IEDs, but only IEDs from the _first_ file are allowed.

**Expected behavior**

If I am allowed to select multiple files, then I expect either:

1. The dialog to combine all my IEDs from both files in the dialog, or 
2. To receive a dialog for each SCL file, or
3. To only be allowed to select a single file.

I think any of these may be a legitimate option and my preference is in the order listed.

danyill:
Resolves #1106.

The import IEDs plugin will still be somewhat prone to failure I believe until we move from the Action API to the Edit API. I'm seeking some clarification from ca-d about this.

The behaviour that has been unhelpful here is caused by async functions in ways we don't test. It's not easy to test these so I have not added any specific tests.

I will be delighted to receive a review and comments on this. :smile_cat: 

(FWIW This PR tries to work around some of the worst behaviour of the Action API with some optional chaining but it is something of a hack and not entirely successful)


danyill:
I've made this PR draft as it is awaiting some improvements to the Action API/document substitution that Christian is working on and once this is complete I will rebase the branch and check performance.

danyill:
This PR can be removed from draft once PR #1233 is merged. However I have tested on top of this branch and the Action API seems to be stable and a large number of IEDs from different scd files can be imported without obvious error.

Although I love nullish coalescing and optional chaining, this specific piece of work has alerted me that they can mask errors which should be noticed and should be used somewhat sparingly!

ca-d (approved):
This LGTM overall and works on my machine, but I'm not sure about the bug you're referencing in your description: Can you still reproduce it? I didn't manage, and I couldn't find a regression test, so I'm assuming it's not fixed yet.

danyill:
The bug referred to in the PR description was mostly related to the exchange of the document and now that we use `editCount` is resolved. Thanks for the review :+1:

JakobVogelsang (approved):
LGTM

PR #1244, opened 2023-04-26
Improve layout for full-screen plugins 

uvulpos:
Issue: #1258

### DIFF
--- a/src/Hosting.ts
+++ b/src/Hosting.ts
@@ -271,73 +271,78 @@ export function Hosting<
     }
 
     render(): TemplateResult {
-      return html` <mwc-drawer
-          class="mdc-theme--surface"
-          hasheader
-          type="modal"
-          id="menu"
-        >
-          <span slot="title">${translate('menu.title')}</span>
-          ${this.docName
-            ? html`<span slot="subtitle">${this.docName}</span>`
-            : ''}
-          <mwc-list
-            wrapFocus
-            @action=${(ae: CustomEvent<ActionDetail>) => {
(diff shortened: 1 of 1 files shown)


src/Hosting.ts
```diff
@@ -271,73 +271,78 @@ export function Hosting<
...
-            @action=${(ae: CustomEvent<ActionDetail>) => {
-              //FIXME: dirty hack to be fixed in open-scd-core
-              //       if clause not neccassary when oscd... compenents in open-scd not list
-              if (ae.target instanceof List)
-                (<MenuItem>(
-                  this.menu.filter(item => item !== 'divider')[ae.detail.index]
-                ))?.action?.(ae);
-            }}
+      return html` 
+      <div style="min-height: 100vh; display: grid; grid-template-rows: auto 1fr">
```
juancho0202, line 275: I'm not completely sure why you would need this extra container div here for a plugin to be "fullscreenable"? I am thinking that you could just have your plugin's CSS to include a fixed positioning and having it on top of everything else on the viewport like this: http://jsfiddle.net/onigetoc/h71sLt9r/

Since I'm intermittently on the calls because of parental leave, I may have missed a refinement call where we discussed this need and its potential implementation, could you point me to the direction of the issue linked to this PR?

uvulpos: That works well in your example, I agree, but then we would have lost the menu header. The goal is to make our telemetry plugin as high as possible, without the possibility to scroll from the body or cut out content and navigate inline in another element.

PR #1252, opened 2023-05-08
fixes: Services section tooltip seems incorrect 

juancho0202:


src/translations/en.ts
```diff
@@ -222,6 +222,7 @@ export const en = {
     lnFilter: 'Logical Node Filter',
     missing: 'No IED',
     toggleChildElements: 'Toggle child elements',
+    settings: 'Show Services the IED provides',
```
Sander3003, line 225: Suggestion: change "Show Services the IED provides" to "Show Services the IED/port provides"

Services can be applied to the IED and/or specific ports
image

image

Sander3003 (changes requested):

src/translations/en.ts
Sander3003: Does this pull-request closes an issue?

src/translations/de.ts
```diff
@@ -225,6 +225,7 @@ export const de: Translations = {
     lnFilter: '???',
     missing: 'Kein IED vorhanden',
     toggleChildElements: 'Kindelemente umschalten',
+    settings: '?????',
```
juancho0202, line 228: @JakobVogelsang please suggest the appropiate german translation for _'Show Services the IED/port provides'_

JakobVogelsang:
lnFilter: 'Filter für logische Knoten'
settings: 'Services für IED or AccessPoint'

src/translations/en.ts
```diff
@@ -222,6 +222,7 @@ export const en = {
     lnFilter: 'Logical Node Filter',
     missing: 'No IED',
     toggleChildElements: 'Toggle child elements',
+    settings: 'Show Services the IED/port provides',
```
danyill, line 225: Rather than `port` should we match Jakob's translation and retain the SCL-specific name with:

```suggestion
    settings: 'Show Services the IED/AccessPoint provides',
```

danyill (dismissed):
Just one little suggestion, thank you for attending to this :+1:

Sander3003 (approved):

PR #1253, opened 2023-05-09
fix: enable IED editor plugin by default

### LINKED ISSUE #1238: Show IED editor by default

As per #1225 opening a file without a substation results in a blank display.

Change the default plugin to be the IED plugin and enable this by default

Stef3st:
closes #1238

### DIFF
--- a/public/js/plugins.js
+++ b/public/js/plugins.js
@@ -1,15 +1,15 @@
 export const officialPlugins = [
   {
-    name: 'Substation',
-    src: '/src/editors/Substation.js',
-    icon: 'margin',
+    name: 'IED',
+    src: '/src/editors/IED.js',
+    icon: 'developer_board',
     default: true,
     kind: 'editor',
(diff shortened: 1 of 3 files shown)


danyill (commented):
Thank you for looking into this. :+1: 

To resolve the logged issue, once we have enabled the plugin by default,  I think we also need to make this the default plugin which loads in the editor view (instead of the Substation plugin). Does that sound correct or should that be another PR?

Stef3st:
> Thank you for looking into this. 👍
> 
> To resolve the logged issue, once we have enabled the plugin by default, I think we also need to make this the default plugin which loads in the editor view (instead of the Substation plugin). Does that sound correct or should that be another PR?

I can do that in this PR, no problem

danyill (approved):
LGTM, just to note that this user-facing change was something we agreed by consensus at a previous standup.

PR #1254, opened 2023-05-09
fix: only later binding  subscribers by default

### LINKED ISSUE #1239: Show only later-binding subscriber plugin by default

Having six plugins for subscription is confusing and cluttering to end users.

(Splitting apart #1225)

As an end-user I'd like the most common option to be the default.
This is the later-binding subscription plugin for GOOSE and SMV as far as I can tell.
The other four subscription editors should be off by default.

Ultimately I hope to have an open-scd-core extension that does GOOSE and SMV in a single plugin which means that the subscriptions would be shown as a single tab.

Stef3st:
closes #1239

### DIFF
--- a/test/unit/editors/subscription/fcda-binding-list.test.ts
+++ b/test/unit/editors/subscription/fcda-binding-list.test.ts
@@ -168,9 +168,6 @@ describe('fcda-binding-list', () => {
       expect(displayedElements.length).to.equal(5);
     });
 
-    /* TODO: This test fails after disabling the SMV/GOOSE subscriber data-/messagebinding plugins by default
-             Check if this test is still relevant to have.
-
     it('allows filtering out of all subscribed control blocks', async () => {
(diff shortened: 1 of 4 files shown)


test/unit/editors/subscription/fcda-binding-list.test.ts
```diff
@@ -168,6 +168,9 @@ describe('fcda-binding-list', () => {
       expect(displayedElements.length).to.equal(5);
     });
 
+    /* TODO: This test fails after disabling the SMV/GOOSE subscriber data-/messagebinding plugins by default
+             Check if this test is still relevant to have.
+
```
Stef3st, line 173: Can this be checked out for relevancy?

danyill (dismissed):
Thank you for attending to this :+1: 

I believe we discussed this at refinement and there was consensus on this UI change.

PR #1255, opened 2023-05-09
fix: enable compare plugin by default

### LINKED ISSUE #1240: Enable compare plugin by default

(splitting apart of #1225)

As a user of OpenSCD I often want to carry out comparisons. Although the current Compare plugin could be enhanced, it would be useful to know it is there and have it available by default.

Stef3st:
closes #1240

### DIFF
--- a/public/js/plugins.js
+++ b/public/js/plugins.js
@@ -202,7 +202,7 @@ export const officialPlugins = [
     name: 'Compare IED',
     src: '/src/menu/CompareIED.js',
     icon: 'compare_arrows',
-    default: false,
+    default: true,
     kind: 'menu',
     requireDoc: true,
     position: 'middle',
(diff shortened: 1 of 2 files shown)


danyill (approved):
Thanks for attending to this :+1: 

Just one small thing. There seems to be something of a house rule to begin a PR description with:

`<PR type>(<plugin type>/<plugin name>): <then the name>`

e.g: `feat(editors/subscription): Make subscriptions cool again`

I am not involved in the release process but it might be worth checking what the house rules are/should be and making a note of them in the README or an ADR since I noticed on our Slack there was some recent difficulty about changelog preparation.

Stef3st:
> Thanks for attending to this 👍
> 
> Just one small thing. There seems to be something of a house rule to begin a PR description with:
> 
> `<PR type>(<plugin type>/<plugin name>): <then the name>`
> 
> e.g: `feat(editors/subscription): Make subscriptions cool again`
> 
> I am not involved in the release process but it might be worth checking what the house rules are/should be and making a note of them in the README or an ADR since I noticed on our Slack there was some recent difficulty about changelog preparation.

The most important thing regarding the release is using git conventional commits (feat/fix) then it will be taken into the release notes. The plugintype/name afterwards is for more details. But since Im now not working on a specific plugin but the entire pluginlist and UI I thought just the conventional commit naming would suffice. Do you agree?

PR #1256, opened 2023-05-09
fix: reduce items on menu if elsewhere in the UI

### LINKED ISSUE #1241: Reduce items on menu if available elsewhere in the UI

As a user of OpenSCD, I'd like the menu to be simple and not repeat functionality used elsewhere (and rarely used from the menu).

(splitting apart of #1225)

I want to suggest that we could declutter the top of the menu in OpenSCD by:

* removing undo and redo and relying on the top right of the app
* removing view log and view diagnostics menu items and instead relying on the top right of the app where we have buttons for these

Stef3st:
closes #1241

### DIFF
--- a/src/Hosting.ts
+++ b/src/Hosting.ts
@@ -237,6 +237,7 @@ export function Hosting<
     renderMenuItem(me: MenuItem | 'divider'): TemplateResult {
       if (me === 'divider')
         return html`<li divider padded role="separator"></li>`;
+      if (me.actionItem) return html``;
       return html`
         <mwc-list-item
           class="${me.kind}"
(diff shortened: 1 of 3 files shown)


danyill (changes requested):
Thank you for attending to this :+1: 

However  I can't open the "Import IEDs", "Settings" or "Extensions" menu items with this PR when I did some local testing so I think perhaps other modifications are required elsewhere otherwise the menu is not in synch with some other data source?

Stef3st:
> Thank you for attending to this 👍
> 
> However I can't open the "Import IEDs", "Settings" or "Extensions" menu items with this PR when I did some local testing so I think perhaps other modifications are required elsewhere otherwise the menu is not in synch with some other data source?

Thank you for checking it out, I'll look at it and let you know when Ive got a solution for this!

Stef3st:
@danyill fixed the problem. It still tried to render menu items that weren't created any more. Should work fine now!

danyill (approved):
LGTM, thanks!

PR #1261, opened 2023-06-07
fix: goose subscription reflects state incorrectly

### LINKED ISSUE #1223: GOOSE Subscription Plugin does not reflects state correctly
**Describe the bug**
There are multiple things but I assume it is the same root cause for all of them.
In the "Subscribe Message Binding (GOOSE)" plugin's "subscriber | publisher" view:
1. You don't seem to be able to add two messages. 
2. When you want to add the second one, it seems to replace the already added one
3. If you have two messages subscribed and want to remove one of them it seems to remove both

I have wrote "seems" everywhere because it does it right but does not displays it correctly.
If you switch to "publisher | subscriber" and then back to "subscriber | publisher" it shows everything correctly

**To Reproduce**
See gif below.

**Expected behavior**
To display the correct state without switching views

**Screenshots**

subscriber-bug

 - OS: MacOS
 - Browser Chrome
 - Version 111

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

Stef3st:
closes #1223


pascalwilbrink (approved):

danyill (changes requested):
Thanks for looking into this. :+1: 

Alas, this change doesn't seem correct to me. :crying_cat_face: 

What we're doing here is the following:

1. We start with the IED we have and we take its Inputs (which comprise ExtRefs which are "plugged" into a DataSets FCDAs)
2. We then go through all the GOOSE control blocks (`GSEControl`) in the document for the other IEDs.
3.  We then look for matching Datasets for this IED and when we find them we go through the entries (FCDAs).
4. We build two lists, that of subscribed control blocks (meaning all FCDAs in this dataset associated with a GSEControl and this IED are subscribed to ExtRefs) and available control blocks (meaning _none_ or _some_ FCDAs in this dataset associated with a GSEControl and this IED are subscribed to ExtRefs).

Within this, the `getExtRef` function is checking for that the  subscribing IEDs inputs section contains an ExtRef which matches the given FCDA and the control block GSEControl.

So this seems as it should be.

danyill:
I hope to try and have a look to see where the problem is, unfortunately the message binding subscription editor was the first subscription editor and it features:

1. not very Lit/reactive design
2. complex event handling (to make up for the first point)

As a result, I don't think it is updating correctly for a range of scenarios in the tests I ran this morning.

danyill:
I made a few updates, I think I didn't understand this clearly this morning (and also got confused by some other behaviour) so I think the change may be quite correct but I'll take another look tomorrow.

danyill (approved):
I've had another and prepared a linked pull request to fix some other separate but linked issues, in fact your solution looks exactly right. Thanks for your patience on this.

Stef3st:
Issues is fixed, but it can not be closed yet due to fast test errors. When this is fixed, this PR will be merged

PR #1263, opened 2023-06-13
fix(editors/subscriber-message) Filter controls with datSet attribute presence, closes #1262

### LINKED ISSUE #1262: Absent datSet in GSEControl causes console error in Message Binding subscriber plugin
**Describe the bug**

While looking at #1261, I realise that the message binding plugins  are not robust for a `GSEControl` (and probably a `SampledValueControl`) with a missing `datSet` which we assume to be present.

https://github.com/openscd/open-scd/blob/a238cc97c271ce6fb55116c438d8f449a5e8d54c/src/editors/subscription/goose/subscriber-list.ts#L197-L227

**To Reproduce**

1. Use the test file, `test/testfiles/editors/MessageBindingGOOSE2007B4.scd`
2. Open the Message Binding (GOOSE) plugin
3. Select the Subscriber view.
4. Select `IED 3`.
5. Try to connect `GCB2 (IED 4)`

Observe a console log:

image

**Expected behavior**

A subscription which does not succeed should not cause a console error.

An absent `datSet` is allowed according to the standard:

> The name of the data set to be sent by the GSE control block. For type=GSSE, the
> FCDA definitions in this data set shall be interpreted as DataLabels according to
> IEC 61850-7-2. The attribute datSet should only be missing within an ICD-File, or to
> indicate an unused control block. It resides in LLN0 like the control block

We should probably filter out control blocks which don't have a `datSet` attribute before building the lists.

danyill:
While looking at #1261 I noticed one reason for the buggy update performance when using our test files is the lack of the `datSet` attribute on a control block.

Closes #1262.

<s>I need to update the snapshots and it looks like some tests need some work so this may take a little while.</s>
done!


Stef3st (approved):
LGTM! Great catch aswell! Didn't realise at first that the datSet was optional 👍

PR #1266, opened 2023-06-15
fix: escaped symbols in regex patterns

### LINKED ISSUE #1265: Regex strings need to be corrected
**Describe the bug**
Chromium browser had an update, which causes the regex patterns to fail.  The solution to this bug is already known thanks to @JakobVogelsang. To solve this problem the `-` and `+` symbols need to be escaped with the `\` char. 

The patterns can be found here:
https://github.com/openscd/open-scd/blob/a238cc97c271ce6fb55116c438d8f449a5e8d54c/src/foundation.ts#L2614-L2630

**Screenshots**
image

Stef3st:
closes #1265


ca-d:
Can someone explain what the changes made are supposed to have to do with the original error? Also, why should those escapes be needed in the character classes? Doesn't regex always interepret `[+-]` as "either `+` or `-`"? Also, why do seven new errors show up in the tests after these changes?

ca-d:
Some quick research reveals https://v8.dev/features/regexp-v-flag is the source of this error, since `<input pattern="` uses the new `v` flag by default, thereby making the unescaped minus at the end of the character class problematic.

ca-d:
https://github.com/openscd/open-scd/pull/1268 is a version of this with fixed snapshots.

Stef3st:
Thank you @ca-d for pointing out that the escape character is not needed in front of the + symbol. I removed these.

Stef3st:
When creating a snapshot test prettier gets called and removes the necessary escape characters needed for the regex pattern. In order to pass the test the patterns will be ignored and test individually.

danyill (approved):
I have learned a little about the `v` regex operator and am a little surprised that upstream has gone about this in this fashion but I guess fixing it and moving on is all we can do. 

Thanks for looking into and fixing this :+1: 

I will be most pleased to see this merged.

PR #1268, opened 2023-06-16
fix(foundation): make patterns compatible with v flag

### LINKED ISSUE #1265: Regex strings need to be corrected

ca-d:
Resolves #1265 .

[The new "v" flag](https://v8.dev/features/regexp-v-flag) in V8, which is used by default for `<input pattern="`, which is in turn used inside `<mwc-textfield`, includes some capabilities for combining character classes, among other things including subtraction of character classes, which now makes it impossible to use the time honored trailing `-` at the end of a character class to mean a literal `-` character, thereby necessitating a backslash for escaping it.

Sic transit gloria regexi.


ca-d:
@Stef3st I suggest this as the solution for your error.

PR #1274, opened 2023-06-23
fix(editors/subscription): Increase timeout for failing subscriber/fcda-binding-list test

### LINKED ISSUE #1257: fcda-binding-list test fails
**Describe the bug**
fcda-binding-list test sometimes fails. It looks like a synchronization/timing issue.

**To Reproduce**
1. run `npm run test:watch` in open-scd
2. Check the results, it doesn't always happen which make it hard to reproduce

**Expected behavior**
Test will always succeed and always have the same behaviour.

**Screenshots**
image

danyill:
Closes #1257

Just a timeout increase, I have looked for a better option and not found one.


Stef3st (approved):
Lgtm, if tests are going too fast I dont see any harm in increasing the time

PR #1275, opened 2023-06-23
fix(editors/IED): Allow IEDs to be deleted after edit count change

### LINKED ISSUE #1272: Unable to delete multiple IEDs in IED editor
**Describe the bug**

On a file I have I am unable to delete IEDs in the current openscd.github.io.

I can usually delete the first one but the second one causes console errors:

image

**To Reproduce**
[XAT_Prot2_Base_Configuration.scd.zip](https://github.com/openscd/open-scd/files/11839892/XAT_Prot2_Base_Configuration.scd.zip)

1. Open the above file
2. Go to the IED editor
3. Delete the first IED successfully.
4. Try to delete another IED.
5. See error.

**Expected behavior**

I can merrily delete IEDs to my heart's content, cheerfully wreaking destruction on my SCL file.

danyill:
Closes #1272

Now that we are using the `editCount`  property we must make sure that plugins are not relying on the old behaviour, like this one. As the `doc` is no longer exchanged on each edit we also need to to look for `editCount` in the `updated` method.

For now I just fix this problem, I hope to have a look through other editors where I suspect similar issues may occur and may not be very visible in our test coverage.

### DIFF
--- a/src/editors/IED.ts
+++ b/src/editors/IED.ts
@@ -96,7 +96,11 @@ export default class IedPlugin extends LitElement {
     super.updated(_changedProperties);
 
     // When the document is updated, we reset the selected IED.
-    if (_changedProperties.has('doc') || _changedProperties.has('nsdoc')) {
+    if (
+      _changedProperties.has('doc') ||
+      _changedProperties.has('editCount') ||
+      _changedProperties.has('nsdoc')
+    ) {
       this.selectedIEDs = [];
       this.selectedLNClasses = [];
 


danyill:
> For now I just fix this problem, I hope to have a look through other editors where I suspect similar issues may occur and may not be very visible in our test coverage.

I can't see that we use the updated method in a similar manner to this in other plugins so I think we're fine.

ca-d (approved):
LGTM, thanks for the fix!

danyill:
Thanks for the review!

PR #1276, opened 2023-06-24
fix(open-scd): Make linear progress bar Github stylez, closes #1269

### LINKED ISSUE #1269: Progress indicator appears in wrong place if vertical height exceeds window height
**Describe the bug**

When I moved the progress indicator to the top bar, it never occurred to me (alas) that we would hide this on overflow, which means it is possible to have the progress indicator appear within the canvas as shown in the screenshot below.

**To Reproduce**

1. Open the clean up plugin with enough items that you need to scroll down.
2. Remove some item.
3. Watch the progress indicator turn up in the wrong place.

**Expected behavior**

The progress indicator should move to the top underneath the top navigation bar.

**Screenshots**

image

 - OS: Windows
 - Browser Chrome
 - Version:  114.0.5735.110 (Official Build) (64-bit)

danyill:
Closes #1269

I couldn't find a solution without changing the z-index without more serious DOM modifications.

I have to use `--primary` I think because I am overwriting `--mdc-theme-primary`.

I  quite like this.

### DIFF
--- a/src/open-scd.ts
+++ b/src/open-scd.ts
@@ -121,17 +121,13 @@ export class OpenSCD extends Waiting(
 
     mwc-linear-progress {
       position: fixed;
+      --mdc-linear-progress-buffer-color: var(--primary);
       --mdc-theme-primary: var(--mdc-theme-secondary);
       left: 0px;
-      top: 112px;
+      top: 0px;
       width: 100%;
       pointer-events: none;
-    }
-
-    @media (max-width: 599px) {
-      mwc-linear-progress {
-        top: 104px;
-      }
+      z-index: 10;
     }
 
     tt {
--- a/src/open-scd.ts
+++ b/src/open-scd.ts
@@ -122,12 +122,12 @@ export class OpenSCD extends Waiting(
     mwc-linear-progress {
       position: fixed;
       --mdc-linear-progress-buffer-color: var(--primary);
-      --mdc-theme-primary: var(--mdc-theme-secondary);
+      --mdc-theme-primary: var(--secondary);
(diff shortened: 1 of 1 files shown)


src/open-scd.ts
```diff
@@ -121,17 +121,13 @@ export class OpenSCD extends Waiting(
...
+      top: 0px;
       width: 100%;
       pointer-events: none;
-    }
-
-    @media (max-width: 599px) {
-      mwc-linear-progress {
-        top: 104px;
-      }
+      z-index: 10;
```
ca-d, line 130:
```suggestion
      z-index: 1000;
```

just to be safe, we want this to be visible whatever other plugins do.

src/open-scd.ts
```diff
@@ -121,17 +121,13 @@ export class OpenSCD extends Waiting(
 
     mwc-linear-progress {
       position: fixed;
+      --mdc-linear-progress-buffer-color: var(--primary);
       --mdc-theme-primary: var(--mdc-theme-secondary);
```
ca-d, line 125:
```suggestion
      --mdc-theme-primary: var(--secondary);
```

ca-d (approved):
LGTM, I especially appreciate the "z" in the pull request title indicating the use of `z-index` in the `css`.

src/open-scd.ts
danyill: Thanks

src/open-scd.ts
danyill: Sure, thanks.

ca-d (approved):
Still LGTM

PR #1278, opened 2023-06-26
Create CODE_OF_CONDUCT.md

benvantende:


pascalwilbrink (approved):

danyill (commented):
It's dangerous waters :skull_and_crossbones:  to contribute to a discussion on ethical standards of any kind on the internet! So I want to just say my piece and then I'm happy to talk to anyone who would like to explore further.

- I want to take seriously the point recently made that "offence is always taken, not given", in the sense that in 2023 it's easy to take offence and that outrage and umbrage are just about a language of their own (thank you social media).

- At the same time offence is sometimes intended but not so often (more just a misunderstanding or cultural context or personal style).

This morning I took a look at the Contributor Covenant and compared to the [Debian Code of Conduct](https://www.debian.org/code_of_conduct) and also the [Go Code of Conduct](https://go.dev/conduct) which is based on the Contributor Covenant (v1.4) but interestingly different.

A few scattered thoughts:

* I do like to have something written - so that we can be transparent and welcoming without opaque backroom conversations or unwritten procedures.
* I think the contributor covenant is heavy on enforcement (most of the document). I know it's shepherded by Microsoft and all...
* I'm not sure it is helpful to have a lengthy set of predefined enforcement steps or approaches - most conflict seems to require a response as tailored and unique as the individuals involved.
* I like to focus on the constructive behaviours we would like to see and in this regard I like the Debian CoC.

So I kind of like L12-L26 of this PR the best but even then with a few reservations.

Similarly in the Go CoC I like the the statement on values (like in the Debian CoC), I like to aspire to something. The Contributor Covenant CoC seems an expression of toleration+nice positive things whereas Debian/Go say, "These are the things that we want to be").

There's some interesting background behind the Go COC here: https://github.com/golang/proposal/blob/master/design/13073-code-of-conduct.md#open-issues

And here's a "trimming" of the Go CoC for a smaller community project which I find quite appealing:

https://github.com/gravitystorm/openstreetmap-carto/pull/2289
https://github.com/gravitystorm/openstreetmap-carto/blob/master/CODE_OF_CONDUCT.md

Specifically, I like in the Open Streetmap version the following:

* the values section
* the statement of goals in "Reporting issues" and the focus on harmony, and forceful measures being a last resort
* the summary section
* the statement about where the CoC applies (I think this is important, overreach in online spaces is real).

All of these seem a bit less present in the Contributor Covenant.
Finally, the version of the Contributor Covenant posted here is version 2.0 but 2.1 is the present version and can be found [here](https://www.contributor-covenant.org/).

Finally, I can live wherever the consensus is on this. I'd encourage people to spend some time reading and getting the flavour of the different CoCs out there to help the thinking process.

Peace out :dove: :olive:

trusz:
Could we approach this iteratively? Or would that make the CoC (Code of Conduct) unserious?

Other than that I think I would rather have a shorter more positively formulated CoC. What are we want to achieve? How we imagine an optimal environment?

Writing a lot about enforcement would personally discourage me from contributing. Also I think we can reserve the right to handle situations how we deem fit, can't we? 

Values are great, if we follow less-is-more, maybe we don't need a summary?
Could the examples of bad behavior lead to people looking for wholes in the CoC and taking it as a challenge to see what is ok and what is not? (not in this regard, but I do have usually the urge to test the limits if any given 😄 )

benvantende:
Agree that this CoC breathes enforcement and that is not we should be looking for . Like in with other stuff we are working on we do not need to reinvent the wheel or work iteratively on it from my perspective. There is for sure something that fits our case. Short is my preference too. The CoC should mainly encourage collaboration and state that we strive for a safe environment.



On 22 August 2023 14:09:23 CEST, "Tamás Russ" ***@***.***> wrote:
>Could we approach this iteratively? Or would that make the CoC (Code of Conduct) unserious?
>
>Other than that I think I would rather have a shorter more positively formulated CoC. What are we want to achieve? How we imagine an optimal environment?
>
>Writing a lot about enforcement would personally discourage me from contributing. Also I think we can reserve the right to handle situations how we deem fit, can't we? 
>
>Values are great, if we follow less-is-more, maybe we don't need a summary?
>Could the examples of bad behavior lead to people looking for wholes in the CoC and taking it as a challenge to see what is ok and what is not? (not in this regard, but I do have usually the urge to test the limits if any given 😄 ) 
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/openscd/open-scd/pull/1278#issuecomment-1688061907
>You are receiving this because you were assigned.
>
>Message ID: ***@***.***>
-- 
ben van 't ende

T: +31 (0)6 47074151

Age of Peers — Open Source Communication

benvantende:
Current recommendation to use https://typo3.org/community/values/code-of-conduct, as this is nicer and has a short version: _**Be constructive and considerate, respect other people's boundaries.
Consider your actions—are they OK?**_

Sander3003:
I'm fine with the Typo3 approach

PR #1281, opened 2023-07-25
New Wizarding API

trusz:
#### Noteworthy changes

1. `package.json`: I've updated lit. I am going to check if everything is ok, but so far so good
2. `Hosting.ts`: Refactored for readability and added the `oscd-wizard-host` element to the DOM
3. `Plugging.ts`: Added the new wizard plugin type also a bit of readability improvements
4. `oscd-card.ts`: There were now card component so I've copied the structure and style from the dialog component
5. `wizard-host.ts`: The wizarding itself.
6. Documentations:
   - ADR: [Wizarding API](https://github.com/openscd/open-scd/blob/772ee6582220394774c114891e2efee605b1d6d1/docs/adr/0002-wizarding-api.md)
   - API: [Wizarding](https://github.com/openscd/open-scd/blob/772ee6582220394774c114891e2efee605b1d6d1/docs/apis/wizarding.md)
   - How Tos: [How To Create a Wizard Plugin](https://github.com/openscd/open-scd/blob/772ee6582220394774c114891e2efee605b1d6d1/docs/how-to/create-a-wizard-plugin.md), [How To Develop For OpenSCD](https://github.com/openscd/open-scd/blob/772ee6582220394774c114891e2efee605b1d6d1/docs/how-to/develop-for-openscd.md)

#### Why not a mixin?

7. It is general philosophy to try to stay close to the DOM. 
8. It creates a more loosely coupled structure. 
9. It does not depend too much on Lit or JavaScript/TypeScript.
10. Its use case is isolated enough not to need too much from the other mixins

#### What is not included and will be in the next step

11. A built-in XML Wizard that is just and XML editor to ensure that there is always at least one wizard available for every element
12. A way of choosing between multiple wizards is available


docs/adr/0002-wizarding-api.md
```diff
@@ -0,0 +1,226 @@
...
+- inspection: view and/or edit an existing element
+- creation: create a new element under a parent element
+
+Wizards must have two public static functions called `canInspect` and `canCreate`.
+`Core` will call these functions when it tries to figure out which wizard supports
+the given element and action.
+```ts
+type CanInspect: (tagName: string) => boolean
+type CanCreate: (tagName: string) => boolean
+```
```
ca-d, line 103: These should probably at least be

```suggestion
```ts
type CanInspect: (element: Element) => boolean
type CanCreate: (parent: Element, tagName: string) => boolean
```
```

because we want to know whether a particular request can be handled.

I personally am still in favor of

```ts
  canHandle(request: WizardRequest): boolean {
```

ca-d:
Does this pull request actually contain any wizard libraries to be loaded? It doesn't seem like it actually needs to do that.

If we want to adopt the new wizard API in open-scd, I would suggest we
1. initially implement the new `WizardRequest` event API (the one facing plugin authors) using the current built-in wizards, without the ability to modularily load wizard libraries,
2. do the other work that needs to be done to allow individual plugins to be factored out
   * pass a `locale` property to plugins so they can translate themselves
   * pass support documents (schema, nsd, nsdoc) to plugins in a unified way using `docs`
   * add `oscd-` prefixed `css` variables for theming colors and fonts
3. start factoring out individual plugins into separate repositories, and finally
4. once we have to factor out a plugin which relies on an adapted wizard (e.g. for extension namespace support or the like), add the ability to load wizard libraries and defined the wizard library facing API

That would make this initial pull request significantly simpler and smaller, and would avoid adding any unused code.

What do you think about this procedure?

pascalwilbrink:
Couple of questions:

* How do you deal with different SCL schemas?
* How can you switch dialogs? For example, instead of having the mwc-dialog, i want to have a dialog from a different component library. (This was one of the requirements)

* Why is wizard a plugin? 
* This solution is creating a (soft) dependency from one plugin to another plugin. My thought was that OpenSCD next would have a microservice architecture, not a distributed monolith.

trusz:
> How can you switch dialogs? For example, instead of having the mwc-dialog, i want to have a dialog from a different component library. (This was one of the requirements)

Added point `12.` to "answer" this question :)

***  
 
> * Why is wizard a plugin?

In order allow different types of wizard. You might want to edit a specific element in a different way and you can provide your own wizard for this element.

***

> This solution is creating a (soft) dependency from one plugin to another plugin. My thought was that OpenSCD next would have a microservice architecture, not a distributed monolith.

I have added point `11.` to the description. Does that make it a bit clearer? Do you think that would remove this dependency?

src/wizard-host.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (!wizardsAvailable) {
+      console.warn({
+        level: 'warn',
+        msg: 'no wizards available for creation, stopping',
+        tagName,
+      });
+      return;
+    }
+
+    const newActiveWizard = availableWizards?.[availableWizards.length - 1];
```
trusz, line 151: NitPick:
`availableWizards.at(-1)` or
`availableWizards.pop()`

trusz:
We will wait with this until Milestone 2 of #1300 is done to see if it still fits into our ideas.

trusz:
I'll close this PR because it has too much conflicts and it only takes up space in the list.
If we need it later we can re-open it.

PR #1282, opened 2023-07-26
Initalize ADR for the repository

trusz:
Rendered version: [1. Record Architecture Decision](https://github.com/openscd/open-scd/blob/caabedfdee3c48820a8e3b4160c4c5b21a4d7b59/docs/adr/0001-record-architecture-decisions.md)

### DIFF
--- a/docs/adr/0001-record-architecture-decisions.md
+++ b/docs/adr/0001-record-architecture-decisions.md
@@ -1,0 +1,25 @@
+# 1. Record architecture decisions
+
+Date: 2023-07-26
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will follow the decisions recorded in the central organizational 
+repository ([.github](https://github.com/openscd/.github)), 
+and record new repo-specific decisions in this repository.
+
+We write ADRs in the `docs` folder instead of a standard `doc`, 
+because `doc` is used for the generated documentation.
+
+## Consequences
+
+- It will be harder to track which decisions have to be taken into consideration
+- Local decisions will be easier to find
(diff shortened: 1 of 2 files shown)


danyill (commented):
LGTM. 

I only didn't approve because I feel we need a "maintainer's approval" and that cannot reasonably be "everyone with admin rights" on this repo which is insufficient to give a architecturally harmonious outcome (recently referred to as the "wild west" :horse_racing: :cowboy_hat_face:)

docs/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,25 @@
...
+We will follow the decisions recorded in the central organizational 
+repository ([.github](https://github.com/openscd/.github)), 
+and record new repo-specific decisions in this repository.
+
+We write ADRs in the `docs` folder instead of a standard `doc`, 
+because `doc` is used for the generated documentation.
+
+## Consequences
+
+- It will be harder to track which decisions have to be taken into consideration
```
juancho0202, line 24: Wouldn't it be easier if everyone adheres to creating ADRs before making big changes to the project and we track those taken into consideration by looking at active PRs?

juancho0202 (changes requested):

docs/adr/0001-record-architecture-decisions.md
trusz: I'm not sure if I understand your point.
Yes big, architecture relevant changes should be a PR with an ADR
The problem is if we have a global ADR in the .github repo that applies to all the repo and projects in this organisation.

pascalwilbrink (approved):
LGTM!

docs/adr/0001-record-architecture-decisions.md
trusz: I agree with your statement generally, but I think that is for another ADR or some other form of agreement, so I think there is no change needed in this PR in that regard.

PR #1289, opened 2023-08-05
Extract wizarding into its own "module"

trusz:
I've extracted the wizarding into its own "module" to make it easier for later to change/replace it. 
No functionality has changed and I re-exported everything from foundation to make sure the PR stays small. We could clean it up in another PR if needed.


trusz:
We will wait with this until Milestone 2 of #1300 is done to see if this still fits into our ideas

trusz:
I close this PR because it has too much conflict and only takes up space.
We can re-open if necessary.

PR #1299, opened 2023-08-18
feat(editing): add open-scd-core Edit API

ca-d:
This adds the Edit API from https://github.com/ca-d/open-scd-core to the Editing and Logging mixins, enabling editing through both APIs.

This change is needed in order to be able to handle namespaced updates at all.

> [!WARNING]
> Do not factor out plugins using this version of the Edit API!
> For compatibility reasons with existing open-scd-core plugins, this pull request just adds the Edit API in its current version, which - while working as intended - still handles namespaced attribute updates quite poorly, as I described in https://github.com/openscd/open-scd-core/issues/132 . I therefore strongly suggest adapting the Edit API before considering it stable. That change could either be made in open-scd-core, with the dependency being updated afterwards, or open-scd-core could be deprecated outright and the change made here, in which case all tests for the Edit API would have to move here, as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "@material/mwc-textarea": "0.22.1",
         "@material/mwc-textfield": "0.22.1",
         "@material/mwc-top-app-bar-fixed": "0.22.1",
+        "@openscd/open-scd-core": "^0.0.2",
         "ace-custom-element": "^1.6.5",
         "lit-element": "2.5.1",
         "lit-html": "1.4.1",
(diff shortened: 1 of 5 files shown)


trusz:
As we discussed, I would like to wait with this PR until we finish Milestone 2 in #1300.
With that we would import `core` from the monorepo directly.
I understand this is needed for other plugins, so we try to hurry.
Thank you for your understanding! ☀️

trusz:
I'll convert this PR to draft for the time being, until the dependencies are done

ca-d:
Thank you for attending to this issue. If I can be of help in preparing the dependencies, I'm always happy to join a pair programming or sparring session!

pascalwilbrink:
Hi @ca-d, We want to split up the current Logging Mixin first into Logging and `Historing`.
Logging will contain `info`, `warning` and `error` logs. Historing will contain the `edit` kinds.
After this split, we can more easily migrate to the new Editing API

trusz:
We would like to include the new edit event so both version of edit API works

Sander3003:
Inspiration: https://github.com/OMICRONEnergyOSS/open-scd-core/commit/4e36a575e468aa9867050ddc0b4402494e4a6205

PR #1308, opened 2023-08-22
infra: mark stale issues automatically

trusz:
This PR sets up a GitHub action to automatically mark stale issues.
It also add a manual trigger so we can run it manually if we want to test it.

---

please ignore: the `docs/decisions/0001-record-architecture-decisions.md` it shows up because of rebase

### DIFF
--- a/.github/workflows/stale-issues-cron.yml
+++ b/.github/workflows/stale-issues-cron.yml
@@ -8,4 +8,4 @@ jobs:
   run-stale:
     uses: "./.github/workflows/stale-issues.yml"
     secrets:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      TOKEN_FOR_STALE_ISSUES: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/stale-issues-cron.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on:
+  schedule:
+    - cron: '0 19 * * *'
+
+jobs:
+  run-stale:
+    uses: "./.github/workflows/stale-issues.yml"
+    secrets:
+      ORG_GITHUB_ACTION_SECRET: ${{ secrets.ORG_GITHUB_ACTION_SECRET }}
```
pascalwilbrink, line 11: there is a default secret in every workflow called `GITHUB_TOKEN`. can you use this one instead of the `ORG_GITHUB_ACTION_SECRET`?

trusz: oh sure

pascalwilbrink (approved):

PR #1309, opened 2023-08-23
fix(ci): stale issue action

trusz:
- removed unneeded permission check


trusz:
too many changes because of missing commits from main branch, will open a new one

PR #1310, opened 2023-08-23
fix(ci): remove unneeded permission check

trusz:
The permissions check does not seem to work in sub-workflows nor is it needed at all.


danyill (approved):
LGTM

PR #1316, opened 2023-08-29
fix: stale issue action

trusz:
See comments for explanations.


.github/workflows/stale-issues.yml
```diff
@@ -5,23 +5,27 @@
 name: Mark stale issues
 
 on:
-  workflow_call:
-    secrets:
-      TOKEN_FOR_STALE_ISSUES:
-        required: true
+  workflow_dispatch:
+  schedule:
+    - cron: '0 19 * * *'
```
trusz, line 10: We don't need multiple files because we can have multiple triggers for an action

.github/workflows/stale-issues.yml
```diff
@@ -5,23 +5,27 @@
...
 
 jobs:
   stale:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
trusz, line 16: This is needed to use `GITHUB_TOKEN` and to be able to add labels and comments to issues

.github/workflows/stale-issues.yml
```diff
@@ -5,23 +5,27 @@
...
     - uses: actions/stale@v5
       with:
-        debug-only: false
-        repo-token: ${{ secrets.TOKEN_FOR_STALE_ISSUES }}
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+
         days-before-stale: 60
+        days-before-close: -1
+        days-before-pr-stale: -1
+        days-before-pr-close: -1
```
trusz, line 26: We don't require permissions for the pull requests, but just to make sure we don't close them automatically

pascalwilbrink (approved):
LGTM!

PR #1320, opened 2023-09-10
fix(wizards/foundation): Added missing DAI bTypes

### LINKED ISSUE #1297: IED editor: Allow editing and instantiation of all values
Some end users have expressed the desire to change IED configuration settings in the IED editor, but were unable to do so, because some of the "edit" and "add" icon buttons appear to be grayed out. With the discussion below I want to find out whether this is a bug or working as intended.

If it's a bug I propose to fix it. If it's working as intended I propose to enhance the IED editor by letting the user choose to be able to edit and instantiate all DAs, or at least to tell the user why the uneditable ones remain uneditable.

### Discussed in https://github.com/openscd/open-scd/discussions/1283

<div type='discussions-op-text'>

<sup>Originally posted by **ca-d** July 28, 2023</sup>
When trying to edit a `LLN0.GrRef.getScrRef` instantiated value which refers to an LDevice which is no longer existent, the "edit" icon is disabled and cannot be clicked.

Why are some "edit" and "add" icon buttons disabled in the IED editor? Is there a way we could enable them all?</div>

jarradraumati:
Closes #1297 

* Added missing DAI bTypes from IEC 61850-6 ed. 2.1
* Updated test snapshots to include added bTypes
* Octet assumed to be no zero padding, no space hexadecimal string value

I think the treatment of Octet bTypes needs some consideration. However, this PR does allow for DAI of bType `ObjectReference` to be edited as requested.

I would appreciate a review, and further refinement if required.

### DIFF
--- a/src/wizards/foundation/dai-field-type.ts
+++ b/src/wizards/foundation/dai-field-type.ts
@@ -38,6 +38,11 @@ const daiFieldTypes = [
   'VisString65',
   'VisString129',
   'VisString255',
+  'ObjRef',
+  'Currency',
+  'Octet64',
+  'Octet6',
+  'Octet16',
 ] as const;
 export type DaiFieldTypes = typeof daiFieldTypes[number];
 const emptyIfNull = <T>(item: T | null, value: string): string => {
@@ -65,6 +70,11 @@ export function getCustomField(): Record<DaiFieldTypes, CustomField> {
     VisString65: stringField('VisString65', 65),
     VisString129: stringField('VisString129', 129),
     VisString255: stringField('VisString255', 255),
+    ObjRef: stringField('VisString129', 129),
+    Currency: stringField('Currency', 3),
+    Octet64: stringField('Octet64', 64 * 2),
+    Octet6: stringField('Octet6', 6 * 2),
+    Octet16: stringField('Octet16', 16 * 2),
   };
 
   function booleanField(): CustomField {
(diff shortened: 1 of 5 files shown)


pascalwilbrink:
@JakobVogelsang could you take a look at this?

JakobVogelsang (commented):
In general LvGTM. Thank you especially for the well written issue linked to the PR. 

I have just a little comment. I would have expected to see a regression test showing grRef data attributes are not disabled and have an edit button next to it as this was the starting point for this PR in the first place. But this also is just a comment that does not need to be resolved as we have not done so for other bType's as well.

jarradraumati:
> I have just a little comment. I would have expected to see a regression test showing grRef data attributes are not disabled and have an edit button next to it as this was the starting point for this PR in the first place. But this also is just a comment that does not need to be resolved as we have not done so for other bType's as well

Ah yes I had thought about doing that. There's also a case for ensuring that bTypes we don't want to allow editing are disabled. But I was struggling to get the tests working correctly in my environment (something I need to look into), and the day was dragging on... Perhaps for the next PR.

pascalwilbrink:
@jarradraumati, Thank you very much for the contribution :)
We'll merge it for now (there is a big PR coming up where the folder structure is changed). I'll update that branch with main

pascalwilbrink (approved):

PR #1322, opened 2023-09-13
Draft

pascalwilbrink:


danyill:
Is this accidental? Why is this here? It seems like a mistaken push.

pascalwilbrink:
It was done to see the differences between CoMPAS-OpenSCD and OpenSCD.
ideally, CoMPAS OpenSCD would only have additions, but this wasn't the case :(

danyill:
Ah. Thanks 👍

PR #1324, opened 2023-09-25
fix(action): correct source folder

trusz:
The deploy action does not use the `working-directory` default parameter so the full path is needed.


Stef3st (approved):
lgtm

PR #1331, opened 2023-10-18
fix: allow address update for new entries

### LINKED ISSUE #1327: Can no longer save VLAN-ID and VLAN-Priority in Wizard
**Describe the bug**

To my surprise, today I discovered that I can no longer save either VLAN-ID or VLAN-Priority for `GSE` or `SMV` elements.

**To Reproduce**

[testSCL.scd.zip](https://github.com/openscd/open-scd/files/12832357/testSCL.scd.zip)

1. Go to the Communications screen.
    1a.  Either through the Substation Editor clicking on the (G) or (S) and then clicking More (three dots) and Communication
    1b.  Or by going to the Communications editor.
2. Open a GSE. 
3. Enable the Priority and VLAN-ID and enter valid values (e.g. `7` and `3E8`)
4. Click Save. 
5. Re-open and see no changes (or alternatively download and examine the SCL).

**Expected behavior**

When I click Save, it saves...

Stef3st:
closes #1327


Stef3st:
in GSE the MAC-Address/APPID/VLAN-ID/VLAN-PRIORITY are put in the addressContent array. When saving the old and the new addressContent gets compared but only check the already existing P types. It doesn't take into account newly added P types.

JakobVogelsang (dismissed):
I am missing a regression test here. We need to make sure this error will not re-appear in the future once again.

JakobVogelsang (approved):
LGTM

PR #1344, opened 2023-10-26
feat: add open-scd-core to the repo

trusz:
It is an as-is copy from `https://github.com/openscd/open-scd-core`


trusz:
We will merge another, update PR: https://github.com/openscd/open-scd/pull/1352

PR #1352, opened 2023-11-02
Added core

pascalwilbrink:


trusz (changes requested):
We need the git logs too

pascalwilbrink:
> We need the git logs too

Fixed in #1363, i'll close and delete this pr

PR #1365, opened 2023-11-14
Release please config

pascalwilbrink:
Added Release Please Config.
After this is merged to main, we can use that GIT SHA to set the bootstrap sha


release-please-config.json
```diff
@@ -0,0 +1,69 @@
...
+      "hidden": true
+    },
+    {
+      "type": "ci",
+      "section": "Continuous Integration",
+      "hidden": true
+    }
+  ],
+  "packages": {
+    "packages/open-scd": {
```
trusz, line 52: aren't we publishing `packages/core`?

trusz (changes requested):

release-please-config.json
trusz: This is for `open-scd`, the one for `core` will follow on later

trusz (approved):

PR #1370, opened 2023-11-20
chore(104): validate IOA number in address-wizard

### LINKED ISSUE #1348: Assigning the same IOA addrees for Mapping DA of a CDC to a 104 Point
**Description of the bug**
When mapping an IEC 61850 Common Data Class (CDC) to an IEC 104 data point in the IEC 104 module, the standard procedure involves clicking on "Add 104 Addresses" and then selecting a Data Object (DO), as illustrated in the image below.

image

Following this step, it is essential to select the appropriate monitoring Type Identifier (Ti) for the chosen DO.

image

After adding the point, a new dialog box containing various parameters will be displayed. At this stage, it is necessary to enter the CASDU number and IOA number. In certain instances, specific DOs within particular CDCs may have more than one Data Attribute (DA). As depicted in the image below, there are five distinct data attributes for which CASDU numbers and IOA addresses must be defined.

image

The key point to note here is that while the CASDU number can remain the same for all the points, it's crucial to assign a unique IOA for each point. Therefore, within the CoMPAS system, it's essential to implement a validation mechanism to prevent users from selecting the same IOA address for different DAs. This ensures that each point is uniquely identified by its IOA address, which is a vital aspect of the mapping process.

It's important to note that the Common ASDU (CASDU) number serves to address a specific entity within the communication network, which can range from a substation to a bay within a substation or even a single Remote Terminal Unit (RTU). Therefore, the combination of CASDU number and Information Object Address (IOA) must be handled with care. 

In cases where there's only one CASDU number designated for an entire substation (e.g., CASDU=1), each IEC 104 point should have a unique and distinct IOA address (e.g., IOA=1, 2, 3, 4, 5, 6, and so on).

On the other hand, when multiple CASDU numbers are allocated within a substation, particularly for each bay (e.g., CASDU=1 and CASDU=2), it is permissible to utilize the same IOA address for both CASDU numbers. For example, CASDU=1 may correspond to IOA=1 from Bay 1, and CASDU=2 could also be linked to an IOA=1 from Bay 2. This flexibility allows for efficient communication and addressing within the substation environment.

**Expected behavior**
CoMPAS should not allow the user to chose the same IOA address for different point as it is shown in the picture below. As it can be seen it the picture below, we have IOA=1 for all the DAs from this DO.

Indeed, it's imperative that CoMPAS enforces a restriction preventing users from selecting the same IOA for different points, as illustrated in the image below. In this instance, all DAs from the same DO share the IOA value of 1, which is a situation that should be avoided to maintain proper data mapping and differentiation.

In scenarios where the same IOA address is employed for different Data Attributes (DAs) within the same CASDU number, the system should ideally generate an error message to alert the user. Examples of such error messages could be:

1. "IOA Address Already in Use"
2. "Duplicate IOA Detected"
3. "Cannot Use the Same IOA for Multiple DAs"
4. "IOA Address Conflict"

These messages would effectively communicate the issue and prompt the user to select a unique IOA address to ensure proper data mapping and avoid potential conflicts.

**Screenshots**

The statement effectively conveys that CoMPAS accepted the usage of the same CASDU and IOA addresses for different Data Attributes (DAs).

image

juancho0202:
Closes  #1348

### DIFF
--- a/packages/open-scd/src/editors/protocol104/wizards/address.ts
+++ b/packages/open-scd/src/editors/protocol104/wizards/address.ts
@@ -105,6 +105,21 @@ export function editAddressWizard(
     const cdc = getCdcValueFromDOIElement(doiElement) ?? '';
     const ti = addressElement.getAttribute('ti') ?? '';
 
+    let casdu = addressElement.getAttribute('casdu') ?? '';
+
+    function validateIOA(value: string, nativeValidity: ValidityState): Partial<ValidityState> {
+      const existingAddress = iedElement.querySelector(`Address[casdu="${casdu}"][ioa="${value}"]`);
+      if(existingAddress){
+        // @ts-ignore
+        (this as WizardInputElement).validationMessage = "IOA Address Conflict" 
+        return {
+          valid: false,
+          customError: true,
+        }
+      }
+      return {}
+    }
+
     // Add the basic fields to the list.
     const fields: TemplateResult[] = [
       html`<wizard-textfield
@@ -136,12 +151,16 @@ export function editAddressWizard(
       </mwc-textarea>`,
       html`<wizard-textfield
         label="casdu"
+        @change="${(evt: Event) => {
+          casdu = (<WizardInputElement>evt.target).value ?? '';
+        }}}"
         .maybeValue="${live(addressElement.getAttribute('casdu') ?? '')}"
         helper="${translate('protocol104.wizard.casduHelper')}"
         required
       >
       </wizard-textfield>`,
       html`<wizard-textfield
+        .validityTransform="${validateIOA}"
         label="ioa"
         .maybeValue="${live(addressElement.getAttribute('ioa') ?? '')}"
         helper="${translate('protocol104.wizard.ioaHelper')}"
--- a/packages/open-scd/src/editors/protocol104/wizards/address.ts
+++ b/packages/open-scd/src/editors/protocol104/wizards/address.ts
@@ -111,7 +111,7 @@ export function editAddressWizard(
       const existingAddress = iedElement.querySelector(`Address[casdu="${casdu}"][ioa="${value}"]`);
       if(existingAddress){
         // @ts-ignore
-        (this as WizardInputElement).validationMessage = "IOA Address Conflict" 
+        (this as WizardInputElement).validationMessage = get('protocol104.wizard.error.ioaConflict');
         return {
           valid: false,
           customError: true,
--- a/packages/open-scd/src/editors/protocol104/wizards/address.ts
+++ b/packages/open-scd/src/editors/protocol104/wizards/address.ts
@@ -107,11 +107,10 @@ export function editAddressWizard(
 
     let casdu = addressElement.getAttribute('casdu') ?? '';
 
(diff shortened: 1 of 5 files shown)


packages/open-scd/src/editors/protocol104/wizards/address.ts
```diff
@@ -105,6 +105,21 @@ export function editAddressWizard(
     const cdc = getCdcValueFromDOIElement(doiElement) ?? '';
     const ti = addressElement.getAttribute('ti') ?? '';
 
+    let casdu = addressElement.getAttribute('casdu') ?? '';
+
+    function validateIOA(value: string, nativeValidity: ValidityState): Partial<ValidityState> {
```
Stef3st, line 110: Why is the nativeValidity parameter in here? I dont see it being used

packages/open-scd/src/editors/protocol104/wizards/address.ts
```diff
@@ -105,6 +105,21 @@ export function editAddressWizard(
...
+      const existingAddress = iedElement.querySelector(`Address[casdu="${casdu}"][ioa="${value}"]`);
+      if(existingAddress){
+        // @ts-ignore
```
Stef3st, line 113: Why is the ts-ignore needed?

packages/open-scd/src/foundation.ts
```diff
@@ -176,6 +176,11 @@ export type WizardInputElement =
   | TextField
   | (AceEditor & {
       checkValidity: () => boolean;
+      setCustomValidity: (message: string) => void;
```
Stef3st, line 179: When will this be used?

packages/open-scd/src/foundation.ts
```diff
@@ -176,6 +176,11 @@ export type WizardInputElement =
...
+      validityTransform: (newValue: string, nativeValidity: ValidityState) => ValidityState;
+      validationMessage: string;
+      validity: ValidityState;
+      reportValidity: () => boolean;
```
Stef3st, line 183: When will this be used?

Stef3st (commented):
Looks good and works proper! I just have a few question here. Could you maybe elaborate?

packages/open-scd/src/editors/protocol104/wizards/address.ts
juancho0202: I thought it was required to bind the .validityTransform param to the validateIOA fn but just double checking it can be omitted 👍

packages/open-scd/src/editors/protocol104/wizards/address.ts
juancho0202: Sorry that was my dirty hack to bypass a valid concern by TS saying that `this` was implicitly of the awful **any** type. I found the proper way to tell the function the type of the `this` object within.

packages/open-scd/src/foundation.ts
juancho0202: can be removed 👍

packages/open-scd/src/foundation.ts
juancho0202: can be removed 👍

Stef3st (approved):
LGTM

PR #1378, opened 2023-11-30
feat(104): added descriptions to ti numbers

### LINKED ISSUE #1373: Adding IEC 104 signal name alongside the Ti number

As a 104 Engineer, I aim to incorporate Address Elements into an existing DOI/DAI that currently lacks these elements. Currently, CoMPAS lacks the display of ASDU/Ti number names during the selection of the appropriate Ti number. Below is a proposed method to effectively showcase the signal name for each Ti number.

The "Add 104 Address" button will appear below/right of the screen. If the user wants to add a 104 address, the user should simply click on the "Add 104 Address" button.

A new window will open, allowing you to select the DOI. This window is designed similarly to the dataset selection window. Below, you can see an example of how to select a Data Object (DO) with its Logical Node (LN), Access Point (AP), and Intelligent Electronic Device (IED). In this stage, any DO can be chosen. For instance, DO=TotVA has been selected to highlight the new feature.

image

After this step, it's crucial to choose the suitable Monitoring Type Identifier (Ti) for the selected Data Object (DO). The IEC 104 standard encompasses numerous Ti numbers, which can be challenging for many individuals to memorize, making it difficult to understand the signal's name solely from its Ti number. To streamline this process, it would greatly benefit from including the signal name alongside the Ti number. This enhancement would simplify the task of selecting the correct Ti number for our specific requirements. 

For this particular CDC and DO pairing, we encounter the option of selecting between two Ti numbers: 35 and 36. As demonstrated below, placing the signal name alongside the Ti number significantly enhances user convenience, making it effortless to choose the correct signal.

image

Upon adding the point, a new dialog box with various parameters will be presented. In this step, displaying the signal name next to the Ti number further aids the user. This final step allows users to confirm their selection of the correct Ti based on the signal name, ensuring accuracy and ease of use. 

image

Below a complete list of the IEC 104 signal name can be found:

image
image
image

Stef3st:
closes #1373

### DIFF
--- a/packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
+++ b/packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
@@ -65,7 +65,8 @@ export function createAddressesAction(
 
     // Create all Monitor Addresses
     const selectedMonitorTi =
-      getValue(inputs.find(i => i.label === 'monitorTi')!) ?? '';
+      getValue(inputs.find(i => i.label === 'monitorTi')!)?.split(' (')[0] ??
+      '';
     const monitorInverted = getSwitchValue(wizard, 'monitorInverted');
     const tiInformation = cdcProcessing.monitor[selectedMonitorTi];
     if (tiInformation) {
@@ -241,7 +242,9 @@ export function createAddressesWizard(
         fields.push(
           html`<wizard-textfield
             label="monitorTi"
-            .maybeValue=${monitorTis[0] ? monitorTis[0] : ''}
+            .maybeValue=${monitorTis[0]
+              ? monitorTis[0] + ' (' + getSignalName(monitorTis[0]) + ')'
+              : ''}
             disabled
           >
           </wizard-textfield>`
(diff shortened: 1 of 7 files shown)


Stef3st:
Thanks for testing it out @DavoodSooran! I'll take a look at the requested changes and retrigger you when it's done!

packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
```diff
@@ -64,7 +65,8 @@ export function createAddressesAction(
 
     // Create all Monitor Addresses
     const selectedMonitorTi =
-      getValue(inputs.find(i => i.label === 'monitorTi')!) ?? '';
+      getValue(inputs.find(i => i.label === 'monitorTi')!)?.split(' (')[0] ??
```
Stef3st, line 68: I'm not a 100% happy with this approach. It works perfectly but feels a bit dirty. If a reviewer has another suggestion I'd love to hear it.

juancho0202: I was thinking go for the first two characters in the string so instead of .split use .substring(0,2). Hacky as well so it's a matter of taste.

juancho0202 (approved):
LGTM!

DavoodSooran:
@Stef3st

I tested it again. It works well 👍

PR #1380, opened 2023-12-04
fix(104): change options between different tis when selecting a doi

### LINKED ISSUE #1375: Protocol 104: Change options between different TI's when selecting a DOI
It is possible to be able to select multiple Ti's when adding a DOI. But all Ti's are bound to the same options.

When a DOI is a chosen a wizard pops up where a Ti number can be chosen. Below that are two switches that can be enabled and therefore used, or can be disabled, depending on if a Ti number allows that option. Right now it doesn't matter which Ti number you choose. If there's only 1 Ti number that allows an option, this option will be available for all Ti's in that DOI.

I would like to see a solution that the options are set for the chosen Ti number

Stef3st:
closes #1375

### DIFF
--- a/packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
+++ b/packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
@@ -1,5 +1,7 @@
 import { get, translate } from 'lit-translate';
 import { html, TemplateResult } from 'lit-element';
+import { Select } from '@material/mwc-select';
+import { SelectedEvent } from '@material/mwc-list/mwc-list-foundation';
 
 import { Switch } from '@material/mwc-switch';
 
@@ -167,6 +169,18 @@ export function disableInvertedSwitch(
   return disableSwitch;
 }
 
+export function disableMonitorInvertedSwitch(
+  tiInfo: Record<string, TiInformation>,
+  tiNumberInfo: string
+): boolean {
+  let disableSwitch = true;
+  const tiNumber = tiNumberInfo.split(' (')[0];
+
+  if (!isNaN(+tiNumber)) disableSwitch = !tiInfo[tiNumber].inverted;
+
+  return disableSwitch;
+}
+
 export function createAddressesWizard(
   lnElement: Element,
   doElement: Element
@@ -218,6 +232,7 @@ export function createAddressesWizard(
 
     if (monitorTis.length > 0) {
       fields.push(html`<wizard-divider></wizard-divider>`);
+      let disabledSwitchByDefault = true;
       if (monitorTis.length > 1) {
         fields.push(
           html`<wizard-select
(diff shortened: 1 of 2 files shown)


packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
```diff
@@ -218,13 +232,25 @@ export function createAddressesWizard(
...
+              const selectedTi = (<Select>e.target).selected!.value;
+              (<Switch>(
+                (<Select>e.target).parentElement!.querySelector(
+                  'mwc-switch[id="monitorInverted"]'
+                )
+              )).disabled = disableMonitorInvertedSwitch(
+                cdcProcessing.monitor,
+                selectedTi
+              );
+            }}
```
juancho0202, line 253: I would prefer if you move this chunk into a separate function for code readability, also verify that your querySelector has found something before trying to modify its disabled property to avoid a "cannot read property of undefined" scenario 👍

packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
```diff
@@ -167,6 +169,18 @@ export function disableInvertedSwitch(
   return disableSwitch;
 }
 
+export function disableMonitorInvertedSwitch(
+  tiInfo: Record<string, TiInformation>,
+  tiNumberInfo: string
+): boolean {
+  let disableSwitch = true;
+  const tiNumber = tiNumberInfo.split(' (')[0];
```
juancho0202, line 177: is this necessary? I thought the value in the dropdown was still set to only the tiNumber and that the split in your other PR was only to account for the maybeValue of the wizardtextfield

packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
```diff
@@ -167,6 +169,18 @@ export function disableInvertedSwitch(
   return disableSwitch;
 }
 
+export function disableMonitorInvertedSwitch(
```
juancho0202, line 172: Why not fixing this faulty behavior for all cases using the disableInvertedSwitch? This seems like code duplication

juancho0202 (changes requested):

packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
Stef3st: Good catch! Will add that in

packages/open-scd/src/editors/protocol104/wizards/createAddresses.ts
Stef3st: you're right, this really isn't necessary. Will remove that part!

juancho0202 (approved):
LGTM!

PR #1384, opened 2023-12-07
feat(104): added ens as a supported cdc type

### LINKED ISSUE #1353: Add 104 Addresses to existing DOI/DAI (CDC = ENS)

As a 104 Engineer, I aim to incorporate Address Elements into an existing DOI/DAI that currently lacks these elements. Presently, CoMPAS does not support some critical CDCs, and we must establish their configurations within the tool. Below, the procedure for configuring CDC=ENS has been outlined.

The "Add 104 Address" button will appear below/right of the screen. If the user wants to add a 104 address, the user should simply click on the "Add 104 Address" button.

image

A new window will open, allowing you to select the DOI. This window is designed similarly to the dataset selection window. Below, you can see an example of how to select a Data Object (DO) with its Logical Node (LN), Access Point (AP), and Intelligent Electronic Device (IED). In the case of CDC=ENS, it's necessary to choose the DO=Beh. However, currently, this specific option is not visible or selectable within IEC 104 module of CoMPAS.

image

After this step, it's crucial to choose the suitable Monitoring Type Identifier (Ti) for the selected Data Object (DO). It's important to note that certain CDCs can be mapped to different Ti values within the IEC 104 standard. As depicted below, the dialog box should provide options for selecting both 30 and 35 in the "Monitor Ti" section, reflecting the flexibility of the protocol. Additionally, given that this CDC can be mapped to multiple 104 addresses, the "Created Inverted Addresses (Monitor)" feature should be enabled. This ensures that the configuration aligns with the specific requirements of the CDC and facilitates comprehensive data mapping.

image

image

image

Upon adding the point, a new dialog box with various parameters will be presented. At this stage, it is essential to input both the CASDU number and  IOA number. In some cases, specific DOs within particular CDCs may encompass more than one DA. For the specific CDC=ENS, there is only one Data Attribute **DA=stVal** that require the assignment of CASDU number and IOA addresse.

It's worth noting that the CASDU number serves to address a specific entity within the communication network, which could range from an entire substation down to a bay within a substation or even a single Remote Terminal Unit (RTU). Consequently, the combination of the CASDU number and IOA necessitates careful management to ensure accurate and conflict-free addressing.

For additional details on defining CASDU and IOA addresses, please refer to the following issue:
https://github.com/openscd/open-scd/issues/1348

image

In scenarios where the same IOA address is employed for different DAs within the same CASDU number, the system should ideally generate an error message to alert the user. Examples of such error messages could be:

"IOA Address Already in Use"
"Duplicate IOA Detected"
"Cannot Use the Same IOA for Multiple DAs"
"IOA Address Conflict"

These messages would effectively communicate the issue and prompt the user to select a unique IOA address to ensure proper data mapping and avoid potential conflicts.

Stef3st:
closes #1353

### DIFF
--- a/packages/open-scd/src/editors/protocol104/foundation/cdc.ts
+++ b/packages/open-scd/src/editors/protocol104/foundation/cdc.ts
@@ -40,6 +40,7 @@ export const supportedCdcTypes = [
   'DPC',
   'DPS',
   'ENG',
+  'ENS',
   'INC',
   'ING',
   'INS',
@@ -257,6 +258,21 @@ export const cdcProcessings: Record<
       '62': {
         daPaths: [{ path: ['setVal'] }],
         create: createAddressAction,
+      },
+    },
+    control: {},
+  },
+  ENS: {
+    monitor: {
+      '30': {
+        daPaths: [{ path: ['stVal'] }],
+        create: createAddressAction,
+        inverted: true,
+      },
+      '35': {
+        daPaths: [{ path: ['stVal'] }],
+        create: createAddressAction,
+        inverted: true,
       },
     },
     control: {},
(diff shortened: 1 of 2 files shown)


DavoodSooran:
@Stef3st I tested the CDC=ENC. It works very well.

pascalwilbrink (approved):

PR #1394, opened 2023-12-19
With oscd open

pascalwilbrink:


packages/core/package.json
```diff
@@ -35,14 +35,6 @@
...
     "dependencies": {
         "@lit/localize": "^0.11.4",
-        "@material/mwc-button": "^0.27.0",
-        "@material/mwc-dialog": "^0.27.0",
-        "@material/mwc-drawer": "^0.27.0",
-        "@material/mwc-icon": "^0.27.0",
-        "@material/mwc-icon-button": "^0.27.0",
-        "@material/mwc-list": "^0.27.0",
-        "@material/mwc-tab-bar": "^0.27.0",
-        "@material/mwc-top-app-bar-fixed": "^0.27.0",
```
trusz, line 45: Oh, this is nice :)

packages/core/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "@openscd/open-scd-core",
```
trusz, line 2: Ah, nice catch

packages/open-scd/src/foundation.ts
```diff
@@ -437,6 +440,10 @@ export interface PendingStateDetail {
   promise: Promise<void>;
 }
 export type PendingStateEvent = CustomEvent<PendingStateDetail>;
+
+/**
+ * @deprecated
```
trusz, line 445: in the future maybe we can give a reason why it is deprecated, or what to use instead of it if anything at all.
I like how you did it by `onOpenDoc` and `handleOpenDoc`

trusz (approved):
just a remark for the future :)

PR #1439, opened 2024-02-12
Feat: Added Waiter addon

pascalwilbrink:


juancho0202:
@pascalwilbrink is there any reason you use `override` in your render method in `packages/open-scd/src/addons/Waiter.ts` ?? It seems to be causing a building error

pascalwilbrink:
> @pascalwilbrink is there any reason you use `override` in your render method in `packages/open-scd/src/addons/Waiter.ts` ?? It seems to be causing a building error

Not really, it failed because it's added in TS 3.4 (i think) and building uses an older version. Removed it

juancho0202 (approved):
lgtm!

PR #1444, opened 2024-02-15
fix: do and da wizard lnclass textfield correction

Stef3st:
@DavoodSooran noticed an error in the do-wizard and da-wizard. In these wizards there's a `logical node description` field with the lnclass value. This PR changed the field title from  `Logical node description` to  `Logical node class`


DavoodSooran:
@Stef3st I tested the changes. It works well.

Sander3003 (approved):
Looks good to me

PR #1449, opened 2024-02-19
Editor addon

pascalwilbrink:


Stef3st (changes requested):
Looks good. I do have a few remarks on files that aren't altered but now can be, because of this solution:

- mock-wizard-editor and mock-editor should no longer have to rely on the Editing mixin and can be altered to the Editor addon. I understand that the mock-editor-logger and mock-setter-logger still require the Editing mixin because the Historing depends on that.

- The Wizarding mixin can be deleted. It's no longer being used.

Finally, unfortunatly the Plugin tests keep failing. I have the same issue locally on my machine at first try, when I retry the tests pass. It looks like a timing/sync issue somewhere.

packages/open-scd/test/integration/editors/SMVSubscriberMessageBinding.test.ts
```diff
@@ -1,25 +1,47 @@
...
+@customElement('smv-mock-open-scd')
+class SmvMockOpenSCD extends MockOpenSCD {
+  @query('smv-plugin')
+  plugin!: SMVSubscriberMessageBindingPlugin;
+
+  renderHosting(): TemplateResult {
+    return html`<smv-plugin
+      .doc=${this.doc}
+      .editCount=${this.editCount}
+      .nsdoc=${this.nsdoc}
```
Stef3st, line 21:
```suggestion
```
Message binding doesn't use nsdoc

Stef3st (commented):
Found a few more things that could be removed

packages/open-scd/test/integration/editors/SMVSubscriberMessageBinding.test.ts
Stef3st: This applies to all message and later binding tests. For some reason only this comment went through in the review

Stef3st (approved):

PR #1462, opened 2024-02-26
feat: make use of lerna nx

### LINKED ISSUE #1420: Make use of Lerna / Nx for Monorepository
If we want to make the most out of a monorepository, we can take a look at [Lerna](https://lerna.js.org) and [Nx](https://nx.dev/).
These 2 (Open Source) build tools are maintained by the same company and work together seamlessly.

Lerna:
Can solve the mapping between (sub)projects

Nx:
Can solve the CI maintenance

**Acceptance criteria**:
- [ ] No more cd'ing into subprojects in GH Actions
- [ ] `npm start` works on root level
- [ ] Npm scripts still work

Stef3st:
closes #1420


Stef3st:
The test, start and doc scripts were not being used by both projects in the previous workflows. Since we want to get rid of cd'ing I excluded the unneeded package in the lerna scripts at root level. Please check this out as well and let me know if this is an acceptable solution or if we should take another approach.

pascalwilbrink (dismissed):
Missing individual `project.json` files (Nx)
`npm run build` should be Nx, not lerna
`npm test` should not exclude core
`npm start` should not exclude core
`npm run doc` should not exclude core

Build should be depending on `clean` and `build`

pascalwilbrink (approved):

PR #1471, opened 2024-03-12
Fix: Moved towards get function from lit-translate

### LINKED ISSUE #1470: Update Lit translate to make use of `get` instead of `translate`
The package lit-translate hasn't been updated in quite some time.
This results that we cannot use the `translate` directive anymore, since it outputs a `TemplateResult` from an older version of lit (lit-html).

We have to use the `get` function from `lit-translate` to get the translation.

pascalwilbrink:
Fixes #1470


packages/open-scd/src/editors/cleanup/datatypes-container.ts
```diff
@@ -470,12 +470,12 @@ export class CleanupDataTypes extends LitElement {
     return html`
       <div>
         <h1>
-          ${translate('cleanup.unreferencedDataTypes.title')}
+          ${this.getAttribute('cleanup.unreferencedDataTypes.title')}
```
Stef3st, line 473:
```suggestion
          ${get('cleanup.unreferencedDataTypes.title')}
```
Should be changed to get. It would solve 1 error from the snapshot tests that go wrong, but unfortunatly not all of them

Stef3st (changes requested):
Looks good, it seems pretty straight forward. I checked the files and found 1 where the translate got changed incorrectly. There are still some other tests that fail, but I havent found any errors in the code changes other then the 1 below.

Stef3st (approved):
LGTM!

PR #1473, opened 2024-03-14
ref(core): clean setup of core

trusz:
The PR includes:
- build steps
- testing setup
- eslint setup and config
- changed spaces to tabs
- events that other projects use
- tests for these files


trusz:
We won't do this this way.

PR #1487, opened 2024-04-03
chore: forcing lerna to not run test scripts in parallel

juancho0202:


pascalwilbrink (changes requested):
not using nx

Stef3st (dismissed):
LGTM

juancho0202:
> not using nx

@pascalwilbrink do you care to elaborate why are you demanding nx for this instead of lerna?

pascalwilbrink:
SO:
Lerna is focused on linking multiple packages from the same project and managing npm publishing, and that's about it. Nx is more focused on managing development workflow for multiple packages.

TLDR;
If some packages require other packages to be built before testing, lerna will fail. With nx, dependencies are checked before targeting the script

pascalwilbrink (changes requested):
The parallel flag is a string, not a boolean
https://nx.dev/nx-api/nx/documents/run-many#parallel

So, `--parallel=1`

juancho0202:
> The parallel flag is a string, not a boolean
> 
> https://nx.dev/nx-api/nx/documents/run-many#parallel
> 
> 
> 
> So, `--parallel=1`

I followed the 4th run-many example in the docs https://nx.dev/nx-api/nx/documents/run-many#examples

pascalwilbrink:
> moreoever with flag `parallel=1` the test script doesn't run successfully only with `parallel=false` it works @pascalwilbrink

That's weird, because it works here:
https://github.com/openscd/open-scd/blob/ab8a6147f09fd6444d4481224fc2777f49b72e17/package.json#L13

Also, if you look at the source code of nx, it parses a `false` to `1`:
https://github.com/nrwl/nx/blob/18.2.0/packages/nx/src/utils/command-line-utils.ts#L180

Stef3st (approved):
Either parallel=false and parallel=1 are acceptable solutions here. This change does solve the parallel testing issue we now have. It LGTM.

PR #1499, opened 2024-04-10
Feat: Added components package

pascalwilbrink:


trusz:
I'll close this for now. Because the direction is not clear enough.

PR #1507, opened 2024-04-19
chore: moving events, interfaces and types to @openscd/core

### LINKED ISSUE #1476: Extract types and interfaces to core
As a distribution developer I would like to count on a reference set of types and interfaces that I can use to create my own wizard, plugin, addon. For that it is proposed that the types and interfaces are moved out of the /packages/open-scd project and into packages/core.

Acceptance Criteria:

- All non-wizard related types and interfaces currently residing in `/packages/open-scd` must be successfully moved to the packages/core project.
- Different foundation files are creating depending on the concern (plugins, editors, etc...) and should go into a folder called /foundation/deprecated/name-of-the-concern.ts
- Any existing dependencies or imports related to the extracted types and interfaces must be updated to point to the packages/core project.

juancho0202:
closes #1476


trusz:
not a small PR :D
Have you done it manually? How big is the risk that something won't work?

juancho0202:
> not a small PR :D Have you done it manually? How big is the risk that something won't work?

I've done it manually, since it is not a change of the code itself but it's just some lines being moved from one file to the other, I expect the current tests with the updated imports to be enough. Fun fact: 176 files only changed their import statements.

juancho0202:
Maybe interesting to discuss this in the review as well https://github.com/openscd/open-scd/issues/1476#issuecomment-2068696959

pascalwilbrink (approved):
LGTM

PR #1510, opened 2024-04-24
chore: create oscd-layout addon

juancho0202:
Some decisions taken here that I think award a discussion:

- former Plugging mixin rendering functionality was separated into the oscd-layout addon
  - snapshot test which kept track of the hosting UI and plugins is now under a separate snapshot test for the layout addon
- a "position: absolute; top: 0" style was added to the mwc-drawer to stick it to the top
- the private get plugins array was awarded a @state annotation and a setter in order to use Lit's reactivity (plus I added a requestUpdate on the open-scd tag whenever someone adds a new external plugin)


packages/open-scd/test/mock-open-scd.ts
```diff
@@ -22,10 +23,19 @@ export class MockOpenSCD extends OpenSCD {
...
+  @query('oscd-layout')
+  layout!: OscdLayout;
+
   renderHosting(): TemplateResult {
     return html`<slot></slot>`;
   }
 
+  render(): TemplateResult {
+    return html`
+    ${this.renderHosting()}
```
pascalwilbrink, line 35: RenderHosting can be removed, already being called by `super.render()`

packages/open-scd/src/open-scd.ts
```diff
@@ -359,415 +329,22 @@ export class OpenSCD extends LitElement {
               .host=${this}
               .editCount=${this.editCount}
             >
-              ${this.renderMain()}
+              <oscd-layout
+                .host=${this}
+                .doc=${this.doc}
+                .docName=${this.docName}
```
pascalwilbrink, line 335: .editCount should also be added if doc is being passed down

pascalwilbrink (changes requested):

packages/open-scd/test/mock-open-scd.ts
juancho0202: renderHosting does not exist in the OpenSCD class anymore only in mock-open-scd, that's why it's needed in this case. Maybe we can do without `renderHosting` and just have:

```ts
render(): TemplateResult {
    return html`
    <slot></slot>
    ${super.render()}`;
}
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+	TemplateResult,
+	query,
+	css
+} from 'lit-element';
+import { get } from 'lit-translate';
+import { newPendingStateEvent } from '@openscd/core/foundation/deprecated/waiter.js';
+import { newSettingsUIEvent } from '@openscd/core/foundation/deprecated/settings.js';
+import { MenuItem, Plugin, Validator, PluginKind, MenuPosition, MenuPlugin, menuPosition, pluginIcons, newResetPluginsEvent, newAddExternalPluginEvent, newSetPluginsEvent } from '../open-scd.js';
+import { HistoryUIKind, newEmptyIssuesEvent, newHistoryUIEvent, newRedoEvent, newUndoEvent, UndoRedoChangedEvent } from './History.js';
+import { Drawer } from '@material/mwc-drawer';
```
pascalwilbrink, line 16:
```suggestion
import type { Drawer } from '@material/mwc-drawer';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { ActionDetail, List } from '@material/mwc-list';
```
pascalwilbrink, line 17:
```suggestion
import type { ActionDetail, List } from '@material/mwc-list';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { ListItem } from '@material/mwc-list/mwc-list-item';
```
pascalwilbrink, line 18:
```suggestion
import type { ListItem } from '@material/mwc-list/mwc-list-item';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { Dialog } from '@material/mwc-dialog';
```
pascalwilbrink, line 19:
```suggestion
import type { Dialog } from '@material/mwc-dialog';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation.js';
```
pascalwilbrink, line 20:
```suggestion
import type { MultiSelectedEvent } from '@material/mwc-list/mwc-list-foundation.js';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { Select } from '@material/mwc-select';
```
pascalwilbrink, line 21:
```suggestion
import type { Select } from '@material/mwc-select';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { Switch } from '@material/mwc-switch';
```
pascalwilbrink, line 22:
```suggestion
import type { Switch } from '@material/mwc-switch';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+import { TextField } from '@material/mwc-textfield';
```
pascalwilbrink, line 23:
```suggestion
import type { TextField } from '@material/mwc-textfield';
```

packages/open-scd/src/addons/Layout.ts
```diff
@@ -0,0 +1,799 @@
...
+
```
pascalwilbrink, line 24:
```suggestion

import '@material/mwc-drawer';
import '@material/mwc-list';
import '@material/mwc-dialog';
import '@material/mwc-switch';
import '@material/mwc-select';
import '@material/mwc-textfield';

```

pascalwilbrink (approved):
LGTM!

PR #1529, opened 2024-05-16
Core tests

pascalwilbrink:


trusz:
I cannot figure much out without a description. I understand that there are new tests, but it seems like there is a lot more. I'll wait for feedback for a while. Maybe we can cherry-pick a few things.

trusz:
We will keep the cleanup and the two new test files

PR #1533, opened 2024-05-23
Chore: Added oscd-edit-completed Event

### LINKED ISSUE #1538: Remove undo-redo-changed-event
Currently, the chain to undo/redo is too long.
This chain should be shortened by introducing an `EditCompletedEvent` that is emitted once an `Edit` has been made.

This event should increase/decrease the EditCount on the OpenSCD class

pascalwilbrink:
Fixes #1538


packages/openscd/src/addons/Layout.ts
```diff
@@ -361,14 +378,16 @@ export class OscdLayout extends LitElement {
...
+    this.handleKeyPress = this.handleKeyPress.bind(this);
     document.onkeydown = this.handleKeyPress;
+    /*
 		this.host.addEventListener('undo-redo-changed', (e: UndoRedoChangedEvent) => {
       this.canRedo = e.detail.canRedo;
       this.canUndo = e.detail.canUndo;
     });
-	}
-	
+    */
```
juancho0202, line 388: I tried out adding a listener in the Layout addon for the `oscd-edit-completed` event and then incrementing or decrementing a redoCount here:

```ts
this.host.addEventListener('oscd-edit-completed', (evt: EditCompletedEvent) => {
  const initiator = evt.detail.initiator;

  if (initiator === 'undo') {
    this.redoCount += 1;
  } else if (initiator === 'redo') {
    this.redoCount -= 1;
  }

  this.requestUpdate();
});
```

Of course I had to turn the canUndo, canRedo state variables into computed properties in this way:
```ts
@state()
redoCount = 0;

get canUndo(): boolean {
  return this.editCount >= 0;
}

get canRedo(): boolean {
  return this.redoCount > 0;
}
```

It could be a way to fix the broken undo redo buttons in the top app bar

juancho0202 (changes requested):
Looks good. It's only missing refactoring the Layout addon a little bit to make the undo/redo buttons in the top-app-bar work again 👍

juancho0202 (approved):

Issue #1490, opened 2024-04-03
Plugin development scaffolding tool


juancho0202:
**Description**
As a developer, I want to easily create new plugins for my project, for that I want to count on a CLI scaffolding script that guides me through the process. When I run the script, it should prompt me to select the type of plugin I want to develop (editor or menu).

Based on my selection, the script should generate a new Lit project in Typescript that features:
- A plugin file with the rough structure of the type of the chosen plugin. This should include the necessary Lit imports and basic structure (run method for menu plugins, render method for editor plugins and the properties `doc` and `editCount`) , leaving placeholders for me to complete the specific behavior of the plugin.
- A test file should also be provided with the rough structure of how to add tests for the newly created plugin.

**Requirements**
- This script should live in a project called /tools/generators 
- The solution must be built with [Nx](nx.dev)

**Acceptance criteria** 

- The CLI scaffolding script should provide a menu for selecting the type of plugin (editor or menu) and it should run from the packages/plugins project, something like npm run create-new-plugin
- CLI should display the following prompts:
  1. Select type (multiselect: editor, menu)
  2. Enter the name of the plugin: (text input) (validated to only have [a-zA-Z] (Pascal Case))
- The script should generate a new TypeScript file (and folder to hold the file) with the appropriate structure and placeholders for plugin-specific behavior (name of the folder and file should be in Kebab Case)
- Tests should be added to ensure this working properly

PR #1534, opened 2024-05-23
Feat: Added OSCD Nx Plugin

pascalwilbrink:


trusz:
As far as I can tell this PR tries to establish templates for addons so developers can generate a new one with existing setup. It does not have a high priority, but nevertheless it is a nice idea and would like to keep it around.

Could be this issue the source of the idea:
- https://github.com/openscd/open-scd/issues/1490

trusz:
I will close the issue, because we think a template repo would be a better solution right now. People may not want to create plugins in the open-scd repo directly, but in their own ones.

PR #1536, opened 2024-05-24
Feat: Added XML Package for OpenSCD

pascalwilbrink:


juancho0202 (changes requested):
Everything looks well executed in the extraction, my only concern is that the migrated tests from `packages/open-scd` to `packages/xml` are not being executed because there is no test script and no devDependency on "@web/test-runner" at the `@openscd/xml` project.
If you add that missing bit this is complete imo.

packages/xml/CONTRIBUTING.md
juancho0202, line 1: I don't think this CONTRIBUTING guide goes in `packages/xml`

juancho0202 (approved):
I am approving because I think it's done, just needs the removal of the duplicated CONTRIBUTING guide but functional and code wise it's done.

packages/xml/CONTRIBUTING.md
pascalwilbrink: Deleted

PR #1542, opened 2024-05-30
Open scd components

pascalwilbrink:


juancho0202 (changes requested):

trusz:
We will close the PR because it is too big to understand and go through all the changes even if they just move files around

Issue #1532, opened 2024-05-22
Subscribing on minimal ExtRef definition


JakobVogelsang:
As a user of OpenSCD I want to be able to see already existing subscriptions in OpenSCD in all circumstances. At the moment when an `ExtRef` element has missing `srcLNClass` and `srcLNInst` is interpreted as unsubscribe. That however is incorrect as 
described here.

<img width="635" alt="grafik" src="https://github.com/openscd/open-scd/assets/66802940/75c7524c-bc0e-44d4-b5db-d002bcae9ecc">

**Acceptance Criteria**
On checking existing subscriptions, the existing rules must be extended so that:
- have `srcLNClass` defaulting to `LLN0` when missing
- have `srcLNInst` defaulting to empty string when missing

**Example**
`<ExtRef iedName="AA1E1Q01BCU" ldInst="MV" ... srcLDInst="LD0" srcCBName="gcb"  serviceType="GOOSE" />` must be interpreted as
`<ExtRef iedName="AA1E1Q01BCU" ldInst="MV" ... srcLDInst="LD0" srcLNClass="LLN0" srcLNInst="" srcCBName="gcb"  serviceType="GOOSE" />`

Sander3003:
What plug-ins are affected in this issue?

pascalwilbrink:
On which `plug-ins` does this occure?
- Later binding / message binding
- Goose / SMV

JakobVogelsang:
It could appear in all the six subscriber plugins:

- Later binding (GOOSE)
- Later binding (SMV)
- Message binding (GOOSE)
- Message binding (SMV)
- Data binding (GOOSE)
- Data binding (SMV)

JakobVogelsang:
The problem is likely to be found here:

<img width="648" alt="grafik" src="https://github.com/openscd/open-scd/assets/66802940/9cca61b0-2702-40a7-b3b0-c361a7e495f7">

Here, a query is constructed with the assumption that `srcLNClass` and `srcLNInst` are not empty.

clepski:
The part in `packages/plugins/src/editors/subscription/foundation#getExtRef` is being used in the Message binding plugins

JakobVogelsang:
@clepski this would be the file for the regression test. It has SMV and GOOSE services linked to the ExtRef and both would fail to show correct subscription at the moment.

[...]

PR #1551, opened 2024-07-09
bugfix(1532) Subscribing on minimal ExtRef definition

clepski:
#1532  Use default value for ExtRef with missing srcLNClass

Take note there is a subtle difference how it is handled in `editors/subscription/foundation.ts` and `editors/subscription/later-binding/foundation.ts`:

- The first ignores the srcLNClass attribute completely for `LLN0` (coming from Control -> closest LN.lnClass)
- The second checks that the attribute in `ExtRef` is either missing or is `LLN0`

### DIFF
--- a/packages/plugins/src/editors/subscription/foundation.ts
+++ b/packages/plugins/src/editors/subscription/foundation.ts
@@ -132,8 +132,10 @@ export function getExtRef(
   control: Element | undefined
 ): Element | undefined {
   function createCriteria(attributeName: string, value: string | null): string {
-    // If ExtRef is missing 'srcLNClass', it defaults to 'LLN0' as specified in the standard
-    if (attributeName === 'srcLNClass' && value === 'LLN0') {
+    // For ExtRef the attribute 'srcLNClass' is optional and defaults to 'LLN0', here we ignore 'srcLNClass' completely for 'LLN0'
+    // because otherwise we would have to extend the querySelector to multiple selector groups checking for 'LLN0' or missing 'srcLNClass'
+    const shouldIgnoreCriteria = attributeName === 'srcLNClass' && value === 'LLN0';
+    if (shouldIgnoreCriteria) {
       return '';
     }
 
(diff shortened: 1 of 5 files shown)


packages/plugins/test/unit/editors/subscription/foundation.test.ts
```diff
@@ -285,4 +286,52 @@ describe('foundation', () => {
       expect(count).to.equal(0);
     });
   });
+
+  describe('use default value for ExtRef without srcLNClass attribute', () => {
+    beforeEach(async () => {
+      doc = await fetch('/test/testfiles/editors/ExtRefWithoutSrcLNClass.scd')
+        .then(response => response.text())
```
trusz, line 293: Not so important, but does async-await does not work here?

trusz:
Just a bit nitpicking :)

packages/plugins/src/editors/subscription/foundation.ts
```diff
@@ -132,6 +132,11 @@ export function getExtRef(
   control: Element | undefined
 ): Element | undefined {
   function createCriteria(attributeName: string, value: string | null): string {
+    // If ExtRef is missing 'srcLNClass', it defaults to 'LLN0' as specified in the standard
```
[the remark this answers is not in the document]
clepski: This one is a bit tricky as we are actually not using the default value for the search query, we ignore the attribute `srcLNClass`, because a long query string is being built here and if we wanted to search for either `srcLNClass = 'LLN0'` or no attribute `srcLNClass` we would have to use multiple selector groups which might be too much effort.

But I can use a hopefully meaningful name for the boolean

packages/plugins/test/unit/editors/subscription/foundation.test.ts
clepski: It should work: We are waiting for the doc to be set, it could be done by explicitly returning the promise as well, but I just copied the code from the other tests and dont see any issue with it

trusz (approved):

Issue #1512, opened 2024-04-30
Editor type plug-ins should be able to be rendered without a doc being present


pascalwilbrink:
Currently only `menu` type plug-ins are supported to work without a `doc` being present.
This should also be the case for `editor` type plug-ins.

*Acceptance criteria*
* `requiredDoc` attribute should not only be present in menu type plugins but also editor type ones
* Editor type plugins should be able to be rendered without a `doc` being present.
  * From a given list of editor type plugins, only the ones with `requiredDoc: false` are visible when no doc is loaded and then if a doc is loaded all of them are visible.
* If no `editor` type plugins are rendered, don't show the `mwc-tab-bar` (CSS `:empty`)

juancho0202:
After implementing this change in https://github.com/openscd/open-scd/pull/1530, the tests for `packages/openscd/test/unit/Plugging.test.ts` stop working specifically at "_add custom plugin dialog_". The tests neither pass or fail just become unresponsive. 

I noticed that this goes away after undoing the adding of **requireDoc: true** to all editor type plugins in `packages/openscd/src/plugins.ts` but this change is now required for all editor type plugins to tell apart the ones that should appear initially at the landing screen.

Sander3003:
See the DRAFT pull-request; it can be fixed.

PR #1575, opened 2024-09-25
feat: allow editor plugins to render without an opened doc

trusz:
Issue:
- https://github.com/openscd/open-scd/issues/1512


trusz:
I have trouble with the `adds a plugin with a name and a valid URL` test in `packages/openscd/test/unit/Plugging.test.ts`

The problem seems to be that I've added `requireDoc: true` or `requireDoc: false` to all plugins configuration in `packages/openscd/src/plugins.ts`

trusz:
It got weirder. If I have a default editor plugin that does not require a doc so `requireDoc: false`, then the test goes through

trusz:
The infinite re-rendering is caused by this line https://github.com/openscd/open-scd/blob/main/packages/openscd/src/addons/Layout.ts#L635

image

not yet sure why

trusz:
I've tried to catch the event if it is not by the user
```ts
<mwc-check-list-item
            class="${plugin.official ? 'official' : 'external'}"
            value="${plugin.src}"
            ?selected=${plugin.installed}
            @request-selected=${(e: Event) => {
              console.log("request-selected",plugin.name, e.detail.source)
              if(e.detail.source !== 'interaction'){
                e.preventDefault();
                e.stopPropagation();
                e.stopImmediatePropagation();
                return false;
              }

            }}
            hasMeta
            left
          >
            <mwc-icon slot="meta">
              ${plugin.icon || pluginIcons[plugin.kind]}
            </mwc-icon>
            ${plugin.name}
          </mwc-check-list-item>
```

But the tests do not run through at all.

trusz:
so that was a missing 
```ts
 await pluginInstaller?.updateComplete
```

in the test code

packages/openscd/package.json
```diff
@@ -85,6 +85,7 @@
     "@typescript-eslint/parser": "^4.29.2",
     "@web/dev-server-esbuild": "^0.2.16",
     "@web/test-runner": "^0.13.22",
+    "@web/test-runner-playwright": "^0.11.0",
```
trusz, line 88: added playwright for besser test debuggability

packages/openscd/src/addons/Layout.ts
```diff
@@ -633,6 +642,14 @@ export class OscdLayout extends LitElement {
             class="${plugin.official ? 'official' : 'external'}"
             value="${plugin.src}"
             ?selected=${plugin.installed}
+            @request-selected=${(e: CustomEvent<{source: string}>) => {
+              if(e.detail.source !== 'interaction'){
+                e.preventDefault();
+                e.stopPropagation();
+                e.stopImmediatePropagation();
+                return false;
```
trusz, line 650: This bit was needed to stop the infinite re-rendering cycle.
Unfortunately I could not find the root cause but for the time being this work around should suffice.

packages/openscd/test/unit/Plugging.test.ts
```diff
@@ -125,26 +126,42 @@ describe('OpenSCD-Plugin', () => {
       );
     });
 
-    it('requires a name and a valid URL to add a plugin', async () => {
```
trusz, line 128: This test was testing multiple thins so I've split it

packages/openscd/test/unit/foundation.test.ts
```diff
@@ -172,20 +172,22 @@ describe('foundation', () => {
     });
   });
 
-  describe('ifImplemented', () => {
-    let nonEmpty: HTMLElement;
-    let empty: HTMLElement;
+// skipped becase of flakiness
+  describe.skip('ifImplemented', () => {
```
trusz, line 176: Everything works if run with Playwright.
I've skipped it until we have playwright running:
- https://github.com/openscd/open-scd/issues/1577

packages/openscd/web-test-runner.config.mjs
```diff
@@ -26,8 +36,13 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
       name: 'integration',
       files: 'test/integration/**/*.test.ts',
     },
+    {
+      name: 'single',
+      files: 'test/unit/foundation.test.ts',
+    },
```
trusz, line 42: This is so you can run single files with `---group single`.
Change to the file you are debugging

packages/openscd/web-test-runner.config.mjs
```diff
@@ -13,8 +14,17 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
...
   browserLogs: false,
+  // browsers: [
+  //   playwrightLauncher({
+  //       product: 'chromium',
+  //       launchOptions:{
+  //         headless:false,
+  //         devtools:true,
+  //       }
+  //   }),
+  // ],
```
trusz, line 25: comment this in if you want to debug a test in the browser

trusz:
I'll close this PR in favour of:
- https://github.com/openscd/open-scd/pull/1579

PR #1579, opened 2024-10-20
Feat: Editor plugins can be rendered without an active document

trusz:
Resolves:
- https://github.com/openscd/open-scd/issues/1512

Pre-requisite:
- https://github.com/openscd/open-scd/pull/1578

Replaces:
- https://github.com/openscd/open-scd/pull/1575


packages/openscd/src/plugins.ts
```diff
@@ -9,97 +9,111 @@ export const officialPlugins = [
     icon: 'developer_board',
     default: true,
     kind: 'editor',
+    requireDoc: true,
```
clepski, line 12: Requiring all old plugins to add this property would be a pretty large breaking change, can we default this is property to `true` for editor plugins and only use it if it is explicitly set to `false` instead?

packages/openscd/src/addons/Layout.ts
```diff
@@ -469,17 +469,26 @@ export class OscdLayout extends LitElement {
...
       </mwc-tab-bar>
-      ${renderEditorContent(this.editors, this.activeTab)}
+      ${renderEditorContent(this.editors, this.activeTab, this.doc)}
     `;
 
-    function renderEditorContent(editors: Plugin[], activeTab: number){
-      const content = editors[activeTab]?.content;
+    function renderEditorContent(editors: Plugin[], activeTab: number, doc: XMLDocument | null){
+      const editor = editors[activeTab];
+      const requireDoc = editor?.requireDoc
```
clepski, line 488: We could assume the default for requireDoc is `true` if missing, e.g. do 
```
const requireDoc = editor?.requireDoc !== false
```

packages/openscd/src/addons/Layout.ts
```diff
@@ -633,6 +642,14 @@ export class OscdLayout extends LitElement {
...
             value="${plugin.src}"
             ?selected=${plugin.installed}
+            @request-selected=${(e: CustomEvent<{source: string}>) => {
+              if(e.detail.source !== 'interaction'){
+                e.preventDefault();
+                e.stopPropagation();
+                e.stopImmediatePropagation();
+                return false;
+              }
+            }}
```
clepski, line 652: This is required to prevent rerendering?

packages/openscd/src/plugins.ts
trusz: done in `Layout.ts`

packages/openscd/src/addons/Layout.ts
trusz: We need the default value also for the tabs so I check earlier.

packages/openscd/src/addons/Layout.ts
trusz: yes

clepski (approved):

PR #1584, opened 2024-10-24
Handle Config Plugin Events

trusz:


packages/openscd/src/open-scd.ts
```diff
@@ -278,15 +259,51 @@ export class OpenSCD extends LitElement {
...
+      this.changePlugin(config)
+
+    }else if( isRemoveEvent ){
+      this.removePlugin(name, kind)
+
+    }else if(isAddEvent && config){
+      this.addPlugin(config)
+
+    }else{
+      console.error("Invalid plugin configuration event", {name, kind, config})
```
clepski, line 297: For logging we should use the log events, so the user can the error message

packages/openscd/src/open-scd.ts
```diff
@@ -341,6 +360,51 @@ export class OpenSCD extends LitElement {
...
+    this.storePlugins(newPlugins);
+  }
+
+  private changePlugin(plugin: Plugin) {
+    const storedPlugins = this.storedPlugins
+    const {name, kind} = plugin
+    const pluginIndex = this.findPluginIndex(name, kind)
+
+    if(pluginIndex < 0) {
+      console.warn("Plugin not found, stopping.", {name, kind})
```
clepski, line 396: Use log event

packages/openscd/src/open-scd.ts
```diff
@@ -278,15 +259,51 @@ export class OpenSCD extends LitElement {
...
+    const isChangeEvent = hasPlugin && hasConfig
+    const isRemoveEvent = hasPlugin && !hasConfig
+    const isAddEvent = !hasPlugin && hasConfig
+
+    // the `&& config`is only because typescript
+    // cannot infer that `isChangeEvent` and `isAddEvent` implies `config !== null`
+    if(isChangeEvent && config){
+      this.changePlugin(config)
+
+    }else if( isRemoveEvent ){
```
clepski, line 290: Code formatting: whitespaces, missing `;` etc

```
if (...) {
  doStuff;
}
```

packages/openscd/web-test-runner.config.mjs
```diff
@@ -31,10 +37,10 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
...
       files: 'test/integration/**/*.test.ts',
     },
-    {
-      name: 'single',
-      files: 'test/unit/foundation.test.ts',
-    },
+    // {
+    //   name: 'single',
+    //   files: 'test/unit/Plugging.test.ts',
+    // },
```
clepski, line 43: FYI you can start the tests like `npm run test:unit` and then select a single test file via the CLI with `F` and select the testfile

clepski (changes requested):

packages/openscd/src/open-scd.ts
trusz: This is meant for developers, should the users see this?

clepski: If this is something that should never happen unless something is wrong then I would throw an error

clepski (approved):

PR #1586, opened 2024-10-29
Feat: Allow .fsd file creation

trusz:


packages/plugins/src/menu/NewProject.ts
```diff
@@ -20,13 +20,23 @@ import {
...
+
+    const acceptedFileExtension = [".ssd", ".scd",".fsd"];
+    const isValidFileFormat = acceptedFileExtension.some((extension) => {
+      return inputs[0].value?.endsWith(extension);
+    })
+
+    if(!isValidFileFormat) {
+      docName = docName + ".scd";
+    }
+
```
clepski, line 39: Linting: `'` instead of `"`, we really should fix the linter config some time

clepski (changes requested):

clepski (approved):

PR #1598, opened 2024-11-18
feat: 1596 Enable transpower publisher plugin

### LINKED ISSUE #1596: Enable Publisher plugin from transpower in open scd
As a user I want to able to use the publisher plugin from the [transpower distribution](https://danyill.github.io/scl-editor/).

Pluging source: https://danyill.github.io/scl-editor/plugins/oscd-publisher/oscd-publisher.js

For the plugin to work we need to adjust open scd to be closer to the agreed upen core specs (https://github.com/OpenEnergyTools/open-scd-core/blob/main/API.md).

**Requirements**

- The publisher plugin can be imported and is working as intended

**Necessary steps**
- Add global styles accoring to the core API specs
- Handle openenergytools edit events v2 (which are slighty different from open scd edit events v2)
- Fix or add icons (material icons vs material symbols)
- Check for bugs resulting from conflicting mwc versions

clepski:
### Changes
- Add global styles accoring to the core API specs
- Handle openenergytools edit events v2 (which are slighty different from open scd edit events v2)
-  Add material symbols

### Demo

Add custom edtior plugin in https://openscd.github.io/open-scd/pr-preview/pr-1598/
with https://danyill.github.io/scl-editor/plugins/oscd-publisher/oscd-publisher.js

### Up for discussion
- Should both material icons and material symbols (3 mb file) be loaded? Can we stick with just one?
- ~~Do we want to add the publisher plugin to the default list of plugins?~~

closes #1596


packages/openscd/src/addons/Editor.ts
```diff
@@ -119,6 +119,15 @@ export class OscdEditor extends LitElement {
...
   async handleEditEvent(event: EditEvent) {
+    /**
+     * This is a compatibility fix for plugins based on open enegery tools edit events
+     * beause their edit event look slightly different,
+     * see https://github.com/OpenEnergyTools/open-scd-core/blob/main/foundation/edit-event-v1.ts for details
+     */
+    if (isOpenEnergyEditEvent(event)) {
+      event = convertOpenEnergyEditEventToEditEvent(event);
+    }
+
```
trusz, line 130: Shouldn't we rather adjust the events in open energy tools' distribution so that they adhere to the API definition? This really does feel good :/

clepski: Well, for the edit event previous to v3 there is no API definition. Once we switch to the edit API v3, the events will be aligned again, so this is only a temporary fix.

packages/openscd/src/addons/Editor.ts
```diff
@@ -244,4 +253,14 @@ function handleRemove({ node }: Remove): Insert | [] {
...
\ No newline at end of file
+}
+
+function isOpenEnergyEditEvent(event: CustomEvent<unknown>): boolean {
+  const eventDetail = event.detail as Edit;
+  return isComplex(eventDetail) || isInsert(eventDetail) || isUpdate(eventDetail) || isRemove(eventDetail);
+}
+
+function convertOpenEnergyEditEventToEditEvent(event: CustomEvent<unknown>): EditEvent {
+  const eventDetail = event.detail as Edit;
```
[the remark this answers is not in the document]
clepski: In open scd the event look like this
```
{
  ...
  detail: {
    edit: edit,
    initiator: initiator
  }
}
```
and in openenergytools like this
```
{
  ...
  detail: edit
}
```

So we wrap the edit in an object and add the initiator

trusz (approved):

PR #1603, opened 2024-11-21
bugfix: Refactor filters in the S-IED creation

illia-solovei-s1:


illia-solovei-s1:
## Notes for takeover

Done
- The desc attribute is now a requirement for every S-IED.
- Added a warning banner that appears if S-IEDs without description are found.
- A click on the banner opens a special wizard to edit the description of the first no-desc S-IED in the list. After editing the user can click on the banner again to edit the next problematic element. If there are no such items left - the banner disappears.

Problems and next steps
- Include desc attribute in the filtering logic

PR #1611, opened 2024-12-15
Feat: Programatic Plugin Activation

trusz:
The goal of the pull request is to allow plugins to switch to specific plugins.
For this an event was created that accepts the `name` and the `src` of a plugin, as this is how we identify them.

Furthermore, an additional event was created to allow plugins to "run" menu plugins.
Because menu plugins render inside the menu sidebar and they are only visible when the sidebar is open we open it too before running the menu plugin to make sure the users see the content. This is somewhat hacky and a better solution should be implemented in the future.


packages/openscd/src/open-scd.ts
```diff
@@ -311,57 +181,22 @@ export class OpenSCD extends LitElement {
 
   connectedCallback(): void {
     super.connectedCallback();
+    this.loadPlugins()
```
trusz, line 184: initial loading of plugins has been revamped

packages/openscd/src/open-scd.ts
```diff
@@ -563,19 +441,22 @@ export class OpenSCD extends LitElement {
...
+      this.loadedPlugins.add(tag);
+      import(plugin.src).then((mod) => {
+        customElements.define(tag, mod.default)
+      })
     }
-
     return {
       ...plugin,
-      content: staticTagHtml`<${tag}
+      content: () => {
```
trusz, line 454: the content is now a function so we regenerate the content if params change

packages/openscd/src/open-scd.ts
```diff
@@ -563,19 +441,22 @@ export class OpenSCD extends LitElement {
...
+        return staticTagHtml`<${tag}
             .doc=${this.doc}
             .docName=${this.docName}
             .editCount=${this.historyState.editCount}
+            .plugins=${this.storedPlugins}
```
trusz, line 459: we pass in the whole plugins config, so plugins can access their own and others plugins' config

packages/openscd/src/open-scd.ts
```diff
@@ -52,176 +53,45 @@ import { InstalledOfficialPlugin, MenuPosition, PluginKind, Plugin } from "./plu
 import { ConfigurePluginEvent, ConfigurePluginDetail, newConfigurePluginEvent } from './plugin.events.js';
 import { newLogEvent } from '@openscd/core/foundation/deprecated/history';
 
-// HOSTING INTERFACES
```
trusz, line 55: This has been moved to the end of the file and other stuff got moved to up to promote quicker understanding of the structure and purpose of the file

packages/core/foundation/scl.ts
```diff
@@ -0,0 +1,89 @@
+import { crossProduct } from '../foundation.js';
```
trusz, line 1: file was missing.

packages/openscd/src/plugins.ts
```diff
@@ -3,6 +3,14 @@ function generatePluginPath(plugin: string): string {
...
 
 export const officialPlugins = [
+  {
+    name: 'Launcher',
+    src: 'http://localhost:50714/index.js',
+    icon: 'launch',
+    default: true,
+    kind: 'editor',
+    requireDoc: true,
+  },
```
trusz, line 13: remove this

packages/core/foundation/scl.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function getDataModelChildren(parent: Element): Element[] {
+  if (['LDevice', 'Server'].includes(parent.tagName))
+    return Array.from(parent.children).filter(
+      child =>
```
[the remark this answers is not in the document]
trusz: It wasn't building without it. It has a reference to it, but it was missing

packages/openscd/src/addons/Layout.ts
```diff
@@ -155,6 +158,7 @@ export class OscdLayout extends LitElement {
         },
         disabled: (): boolean => !this.historyState.canUndo,
         kind: 'static',
+        content: () => html``,
```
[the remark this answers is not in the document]
trusz: no, content is required

clepski (approved):

PR #1615, opened 2025-01-08
feat: Edit api v3

### LINKED ISSUE #1593: Support Edit API v3
As a plugin developer I want to use the edit API v3 in my plugins. As a user I want to use plugins which are using the edit API v3 in open scd.

Implement edit API V3 according to the [API definition](https://github.com/ca-d/open-scd-core/blob/be0be6e27ec9835b4edb2134724997ec8a89e401/API.md#editeventv2)

**Requirements**
* Support edit API v3
* Support edit API v2 for backwards compatibility

clepski:
Support edit event API v2, according to [API definition](https://github.com/ca-d/open-scd-core/blob/be0be6e27ec9835b4edb2134724997ec8a89e401/API.md#communicating-user-intent-to-openscd-core). Note the updated [docs](https://github.com/openscd/open-scd/blob/feat/1593-edit-api-v3/docs/core-api/edit-api.md).

* Added custom property `createHistoryEntry` to `EditEventOptions` for undo and redo
* Edit event API v1 und edit action API are still supported
* To support backwards compatibility all new types and functions have `V2` as suffix and are exported from the `core` package

closes #1593


packages/core/foundation/deprecated/edit-event.ts
michelguerin, line 1: You might add a [tsdoc](https://tsdoc.org/pages/tags/deprecated/) comment to tag all these functions as deprecated like this : 

```js
/**
 * @deprecated Use the new edit event V2 API instead.
 */
export function newEditEvent<E extends Edit>(...) {
  . . .
}
```

michelguerin (commented):
Maybe we could plan to add your work on supporting the legacy event API (ie : event to event v1) to the @openenergytools package dedicated to the event API

Besides, do you know if we are planing to integrate this package instead of patching this instance ? Having a separate maintained package for this API seems to be a good option.

docs/core-api/edit-api.md
```diff
@@ -1,13 +1,123 @@
-# Edit Event API
+# Edit Event API v2
```
trusz, line 1: Haven't we agreed to call it V3?

packages/core/foundation/deprecated/edit-event.ts
```diff
@@ -0,0 +1,78 @@
...
+  attributes: Partial<Record<string, AttributeValue>>;
+};
+
+/** Intent to remove a node from its ownerDocument */
+export type Remove = {
+  node: Node;
+};
+
+/** Represents the user's intent to change an XMLDocument */
+export type Edit = Insert | Update | Remove | Edit[];
```
trusz, line 27: Is it possible to have more then one level of nesting here?
Because of `Edit = Edit[]` I can create something like
```ts
const newEdit: Edit = [[[
{...},
]]]
```

is this handled? or can it even happen?

clepski: Yes, that is possible and it is handled. The `handleEdit` moves through the edits recursively, like so
```
handleEdit(edit) {
  if (isArray) {
    edit.map(e => handleEdit(e))
  }

  ...
}
```

For example this can happen, if a complex `editorAction` with a `Replace` is transformed, because a `Replace` gets transformed into `[ delete, insert ]`, so editor actions `[ replace, create ]` would become `[ [ delete, insert ], insert ]`

docs/core-api/edit-api.md
clepski: That was my intention at first, but for compatibility reasons we cannot renamed the already exported edit event V1 stuff and skipping from V1 to V3 felt confusing. Also this is the naming convention in the agreed upon [API](https://github.com/ca-d/open-scd-core/blob/be0be6e27ec9835b4edb2134724997ec8a89e401/API.md#editeventv2)

trusz (approved):

michelguerin (approved):

Issue #1623, opened 2025-02-12
[BUG] Click side effect on local plugin load


michelguerin:
# Description

**Describe the bug**
Loading a local plugin seems to have side effects : 
* click side effect in the plugin menu

**To Reproduce**
1. Load a local plugin from the plugins menu
2. Click multiple times on the plugin checkbox in the plugin menu
4. Error appears

**Expected behavior**
Loading a local/custom plugin should have the same effect as using a core plugin, with no side effects.

**Screenshots**

Start page issue : 

https://github.com/user-attachments/assets/771f6ca2-7811-43b4-b5f7-a3c18e1166d9

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Firefox Developer Edition
 - Version: 136

## Hints : 

This regression is caused by commit : d3b2a0a

Issue #1622, opened 2025-02-12
[BUG] Inconsistent behavior on refresh when using a local plugin


michelguerin:
# Description

**Describe the bug**
Loading a local plugin and refreshing the page seems to have side effects : 
* when refreshing the page the start page is buggy (showing one, two or zero icons)
* some plugins are deactivated for no reason
* Data from loaded file doesn't get to the plugin (saving it shows a blank file)

**To Reproduce**
1. Load a local plugin from the plugins menu
2. Refresh multiple times
4. Error should appear after one to five refreshes

**Expected behavior**
Loading a local/custom plugin should have the same effect as using a core plugin.

**Screenshots**

Start page issue : 

<img width="2512" alt="Image" src="https://github.com/user-attachments/assets/9790e30a-5489-4724-ad77-5a96b61815cd" />

<img width="2504" alt="Image" src="https://github.com/user-attachments/assets/bc94f573-fab6-4fe0-9c65-d24a1f3b3c3f" />

<img width="2500" alt="Image" src="https://github.com/user-attachments/assets/e71a1345-9795-45eb-ad6e-d7ede0db32db" />

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Firefox Developer Edition
 - Version: 136

## Hints

This regression is caused by 101aaae

michelguerin:
This issue is related to commit 101aaae

michelguerin:
Side effect issue by clicking the plugin in the plugin menu is related to this commit d3b2a0a.
I will split this issue in two.

michelguerin:
Fixed by #1626

michelguerin:
Fixed by #1626

PR #1626, opened 2025-02-18
Fix: inconsistent plugin activation behaviour caused by refactoring

trusz:
resolves:
- https://github.com/openscd/open-scd/issues/1622
- https://github.com/openscd/open-scd/issues/1623


packages/openscd/src/addons/plugin-manager/custom-plugin-dialog.ts
```diff
@@ -0,0 +1,207 @@
+
```
trusz, line 1: I've extracted this component from the `Layout.ts`

packages/openscd/src/addons/plugin-manager/plugin-manager.ts
```diff
@@ -0,0 +1,186 @@
+import {
```
trusz, line 1: I've extracted this component from the `Layout.ts`

packages/openscd/src/addons/plugin-manager/plugin-manager.ts
```diff
@@ -0,0 +1,186 @@
...
+          <mwc-list
+            id="pluginList"
+            multi
+            @selected=${(e: MultiSelectedEvent) => {
+              const selectedPlugins = this.pluginList.items
+              .filter((item, index) => e.detail.index.has(index))
+              // @ts-expect-error: we add plugin to the list item
+              .map(item => item.plugin) as Plugin[]
+
+              this.dispatchEvent(newSetPluginsEvent(selectedPlugins))
```
trusz, line 54: This is the main fix.
In the past we made the assumption that the order of plugins in the list is the same that we store in local storage and in memory and we have used the index to identify which plugin has been enabled/disabled.

I have changed the plugin so that instead of sending the indices of enabled plugins we send the plugins itself.

packages/openscd/src/open-scd.ts
michelguerin, line 1: Lots of commented parts that could be deleted ?

packages/openscd/src/plugins.ts
```diff
@@ -1,125 +1,127 @@
...
+  //   activeByDefault: false,
+  //   kind: 'editor',
+  //   requireDoc: true,
+  // },
+  // {
+  //   name: 'Subscriber Message Binding (GOOSE)',
+  //   src: generatePluginPath('plugins/src/editors/GooseSubscriberMessageBinding.js'),
+  //   icon: 'link',
+  //   activeByDefault: false,
+  //   kind: 'editor',
```
clepski, line 37: Are theses commented out by mistake?

trusz: yes, they need to stay. Thank you.

packages/openscd/src/open-scd.ts
trusz: maybe you looked an older version? I've removed them.

packages/openscd/src/plugins.ts
```diff
@@ -128,16 +130,25 @@ export const officialPlugins = [
...
+  {
+    name: 'Plugin Store',
+    src: 'https://sprinteins.github.io/oscd-plugin-store/index.js',
+    icon: 'create_new_folder',
+    activeByDefault: true,
+    kind: 'menu',
+    requireDoc: false,
+    position: 'bottom',
+  },
   {
```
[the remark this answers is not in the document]
trusz: I've turned it off by default and added a "Beta" tag to it

clepski (approved):

PR #1639, opened 2025-03-05
fix: Build entire project and attach distribution to build output

salvar3nga:


.github/workflows/release-please.yml
```diff
@@ -38,13 +38,13 @@ jobs:
       
       - name: Build Distribution
         if: ${{ steps.release.outputs.releases_created }}
-        run: cd packages/distribution && npm run build
```
clepski, line 41: I would like to have just one build step, so we should replace the step in lines 31-32 with this build and remove this step.

clepski (changes requested):

clepski (approved):

PR #1650, opened 2025-03-26
feat: remove Communication from the substation editor and use oscd-publisher plugin

### LINKED ISSUE #1644: Remove Communication from the substation editor
Remove 

From the substation editor including all sub menu’s,

This is replace by publisher plug-in to be more user friendly and feature complete.

Remove these as well:

Please don’t remove the wizards; It can be re-used later by others. [open-scd/packages/plugins/src/wizards at main · openscd/open-scd](https://github.com/openscd/open-scd/tree/main/packages/plugins/src/wizards)

noraeb:
- Removes the old publisher plug-in from OpenSCD
- Adds the new publisher plugin
- Removes communicagtion buttons

Closes #1644


packages/openscd/src/plugins.ts
```diff
@@ -103,7 +103,7 @@ export const officialPlugins: PluginConfig[] = [
   },
   {
     name: 'Publisher',
-    src: generatePluginPath('plugins/src/editors/Publisher.js'),
+    src:  'https://openenergytools.github.io/oscd-publisher/oscd-publisher.js',
     icon: 'publish',
```
clepski, line 107: Adding a plugin by url can be problematic: It doesnt work for all environments, for example this wont work on the Alliander deployment. Also there is no version control, it will always fetch the newest version.

Instead we should use git submodules which fixes a commit, but currently the extract plugins PR is still open and I recommend waiting for that on to be merged

https://github.com/openscd/open-scd/pull/1645

noraeb: Thanks for the feedback @clepski. I'll change it into a submodule and will wait for the plugins PR to be merged

clepski:
FYI, the publisher has been deployed as a openscd repository and we can replace the openenergytools version, see https://github.com/com-pas/compas-open-scd/pull/374

clepski:
> Removes the old publisher plug-in from OpenSCD

We can remove the "old" publisher code right? https://github.com/openscd/open-scd/tree/main/packages/plugins/src/editors/publisher

PR #1653, opened 2025-04-02
Feat: Vite migration

rdsov:
Migrates the build system from snowpack to vite.


Tamriel:
When running OpenSCD and clicking "New project", the console in the browsers dev tools logs `Uncaught TypeError: ae.target.items[ae.detail.index].nextElementSibling.run is not a function`

rdsov:
@Tamriel This PR is still a WIP. I can't really do much as of right now because of the way the plugins get built and deployed, hence why I am await a fix first.

Sander3003:
Nice initaitive!

1. What is the status? / Are you wainting for the Bearingpoint PR's to be accepted?
2. Do you need help?

PR #1654, opened 2025-04-03
fix: modify start script to clear snowpack cache

salvar3nga:
### 🐛 Problem
During the build process (`npx nx run-many -t build --all`), I've encountered the following error:

```
Error: Cannot find module 'lit-html/is-server.js' from '.../packages/distribution'
```

@clepski found that **Snowpack was caching an outdated or incompatible version** of `lit-html` which was solved on compas-openscd [here](https://github.com/com-pas/compas-open-scd/blob/main/packages/compas-open-scd/package.json#L74-L75). In OpenSCD this occurs specifically inside distribution:

```
packages/distribution/node_modules/.cache/snowpack/build/lit@2.8.0
```
This cache was preventing Snowpack from correctly resolving the module, leading to build failures.

### 💡 Solution
To prevent this issue, I've modified the `start` script to **automatically clear the problematic Snowpack cache before running the project**.

- I've added a cleanup step using `rimraf` to remove the cached Snowpack build files before running `lerna run start`.  
- This ensures that every time the project starts, it does so with a clean cache, avoiding errors caused by stale dependencies.  

**Updated `package.json` script:**
```json
"scripts": {
  "start": "npx rimraf packages/distribution/node_modules/.cache/snowpack/build/lit@2.8.0 && lerna run start"
}
```

### 🔍 Alternative Considered & Room for discussion
Instead of modifying the `start` script, we can consider creating a separate script (`start:clean`) that would allow developers to manually trigger the cache cleanup when needed.

#### **Option 1: Attach Cleanup to `start` (Chosen Solution)**
```json
"start": "npx rimraf packages/distribution/node_modules/.cache/snowpack/build/lit@2.8.0 && lerna run start"
```

✅ Advantages:

- Ensures that every time the project starts, it runs without cache-related issues.
- Reduces debugging time by automatically resolving the problem before it occurs.
- No extra steps required—developers just run npm start as usual.

❌ Trade off:

- Slows down the startup slightly since it clears the cache every time.
- Removes the cache even when it's not needed, potentially affecting Snowpack performance.

#### **Option 2: Create a Separate start:clean Script**
```json
 "start": "lerna run start",
  "start:clean": "npx rimraf packages/distribution/node_modules/.cache/snowpack/build/lit@2.8.0 && npm start"
```

✅ Advantages:

- Faster startup since the cache is only cleared when explicitly needed.
- Gives developers control over whether to clear the cache or not.

❌ Trade off:

- Requires developers to remember to run npm run start:clean when they encounter issues.
- Potentially leads to inconsistent development experiences if some devs forget to clean the cache.


clepski:
> ### 🔍 Alternative Considered & Room for discussion
> 
> Instead of modifying the `start` script, we can consider creating a separate script (`start:clean`) that would allow developers to manually trigger the cache cleanup when needed.

While the solution is kind of hacky, I dont think it is worth the time to investigate deeper into the issue, because we are planning to replace snowpack anyway. For now it works and in a couple of months we will probably be using another build tool.

clepski (approved):

Issue #1633, opened 2025-02-26
Unable to build OpenSCD distribution


ocou-aprico:
**Describe the bug**

Unable to build the OpenSCD project.
Snowpack fails to build the packages/distribution module.

Error message:  Cannot find module '@openscd/open-scd/src/open-scd.js' from '/Users/xxxx/open-scd-0.37.1/packages/distribution'

**To Reproduce**

1. Download the open-scd-0.37.1 ZIP file from Open SCD Git location.
2. Unzip
3. Launch command: npm run build or npm run serve
4.  Build failed with this error trace:

> @openscd/distribution@0.0.1 build
> snowpack build && workbox generateSW workbox-config.cjs && cp .nojekyll build/

[13:13:22] [snowpack] ! building files...
[13:13:22] [snowpack] Build Result Error: There was a problem with a file build result.
[13:13:22] [snowpack] Cannot find module '@openscd/open-scd/src/open-scd.js' from '/Users/xxx/open-scd-0.37.1/packages/distribution'
[13:13:22] [snowpack] Error: Cannot find module '@openscd/open-scd/src/open-scd.js' from '/Users/xxx/open-scd-0.37.1/packages/distribution'
    at Function.resolveSync [as sync] (/Users/xxx/open-scd-0.37.1/node_modules/resolve/lib/sync.js:111:15)
    at Object.resolveEntrypoint (/Users/xxx/open-scd-0.37.1/node_modules/esinstall/lib/entrypoints.js:149:59)
    at PackageSourceLocal.resolvePackageImport (/Users/xxx/open-scd-0.37.1/node_modules/snowpack/lib/cjs/sources/local.js:619:40)
    at resolveImport (/Users/xxx/open-scd-0.37.1/node_modules/snowpack/lib/cjs/build/file-builder.js:94:48)
   ...
npm error Lifecycle script `build` failed with error:
npm error code 1

**Expected behavior**

Build should work out of the box.

**Desktop (please complete the following information):**
 - OS: Mac OS 14.6.1 (23G93) (Sonoma)
 - node: v22.14.0
 - npm: 10.9.2

clepski:
> To Reproduce
> 
> Steps to reproduce the behavior:
> 
>     Download the open-scd-0.37.1 ZIP file from Open SCD Git location.
>     Unzip
>     Launch command: npm run build or npm run serve
>     Build failed with this error trace:
> 

Hey there, can you try again with these steps

- npm install
- npm run build
- npm run start

[...]

PR #1663, opened 2025-04-15
State that node version 18 is needed, not version 20

Tamriel:
Solves https://github.com/openscd/open-scd/issues/1633


ca-d (approved):
Did't check myself that it  doesn't work wih version 20, but it seems to work just fine with 18 either way.

ca-d (approved):

PR #1670, opened 2025-04-22
fix: change pattern definition of tIDNaming to non-whitespace characters

noraeb:


JakobVogelsang (approved):
Code looks good. Test are failing. So long those run I approve

PR #1680, opened 2025-05-08
feat: move gse smv between connectedAPs

### LINKED ISSUE #1613: Move `GSE` and `SMV` between `ConnectedAP`
As a user of OpenSCD I want to move a `GSE` or `SMV` element between different `ConnectedAP` element of the same IED.

The communication related settings of a GOOSE control block are saved within the `GSE` element in the `Communication` section and here in the parent `ConnectedAP`. As such, a GOOSE is linked to an access point of an IED. This is used by some vendors to scross check if a link - GOOSE data to input - does make sense.

At the moment when creating a GOOSE we add the `GSE` we add it the `ConnectedAP` that is linked with the server, which works for most of the users, but some need to move the `GSE` and `SMV` elements between different `ConnectedAP` of the same IED.

I want to navigate to the communication plugin and click on a `GSE` or `SMV` element and move it to another `ConnectedAP` element of the same IED.

**Requirements**
- add a button to the GSE/SVM element representation in the communication plugin
- on button click give the user the choice of all other ConnectedAP element of the same IED
- disable the new button in case of:
   - an IED has only one AccessPoint and therefore only one ConnectedAP
   - an IED has multiple AccessPoints but only on ConnectedAP (only one access point it linkt to a subnetwork)

**Additional information**
- Mockups

Image

Image

noraeb:
closes #1613


packages/plugins/src/editors/communication/foundation.ts
```diff
@@ -0,0 +1,35 @@
...
+  const currentConnectedAP = getCurrentConnectedAP(element);
+  if (!currentConnectedAP) {
+    return [];
+  }
+
+  const iedName = currentConnectedAP.getAttribute('iedName');
+
+  return Array.from(doc.querySelectorAll('SubNetwork > ConnectedAP')).filter(
+    connectedAP => connectedAP.getAttribute('iedName') === iedName
+  );
```
clepski, line 34: The filtering should be done in the querySelector for performance reasons, `'SubNetwork > ConnectedAP[iedName="${iedName}"]'`

packages/plugins/src/editors/communication/subnetwork-editor.ts
```diff
@@ -57,6 +68,33 @@ export class SubNetworkEditor extends LitElement {
...
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('request-gse-move', e =>
+      this.openMoveDialog(e as CustomEvent<{ element: Element }>)
+    );
+    this.removeEventListener('request-smv-move', e =>
+      this.openMoveDialog(e as CustomEvent<{ element: Element }>)
+    );
+  }
```
clepski, line 96: Template event listeners are a bit easier to use https://lit.dev/docs/components/events/#adding-event-listeners-in-the-element-template

Then we don't need to manually take care of removing the event listener

```
<gse-editor
  ...
  @request-gse-move="${event => doStuff(event)}" />

```

clepski (changes requested):

packages/plugins/src/editors/communication/foundation.ts
noraeb: thanks for the tip! Fixed it

packages/plugins/src/editors/communication/subnetwork-editor.ts
noraeb: Nice, fixed this as well

clepski (approved):

